PR #37, opened 2020-01-23
Fill in problem

ludopulles:
Basically a rewrite of the shutil.copytree, to allow to create a new problem when some of the files do already exist (in case of a CFP), so adding all the submissions directories, etc. goes automatically without overwriting existing files.


RagnarGrootKoerkamp:
Do you still want to change to using some library function? If now we can also just merge this.

ludopulles:
The distutils had a slightly different behaviour as I recall, so perhaps we could merge this.

PR #50, opened 2020-03-30
Add spec for generated testcases.

RagnarGrootKoerkamp:
Adds spec for #1: Generated testcases.

The latest yaml example is [here](https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/generated_testcases/doc/generated_testcases_v2.yaml)

Feel free to add discussion points on relevant lines in the PR.

Remarks:
- **manual cases** Specify manual cases outside of `data/` (and copy them) or require them to be listed in `gen.yaml`? -> both are OK
- **multiple file output**: How to allow generators that write to (multiple) files?
    * **prefix**: `test_group: multiple_output_generator.py {prefix}`, and only files starting with `{prefix}`=`test_group` are preserved. -> This is what we're doing now.
    * **directory**: `test_group: multiple_output_generator.py` which writes multiple files in new directory
    * **glob pattern**: `graph-*.in` or `*-graph.in` or `graph.*`: only files matching the glob are used. Keys with a `*` or `?` are patterns and must write files. Other keys are files and must write `stdout`.
         [this is the only option that allows re-ordering output from the generator]
    * **disallow this**

### DIFF
--- a/doc/generated_testcases_v2.yaml
+++ b/doc/generated_testcases_v2.yaml
@@ -148,8 +148,6 @@ data:
 # Introduce a testgroup by adding a dictionary with `type: directory` set.
 # The top-level is always assumed to be a directory.
       hard_cases_group:
-# TODO: We could drop this because testcase dictionaries need the input: key
-# while testgroups need the include: and/or data: keys.
         type: directory
 
 # Directories may contain a testdata.yaml that will be written as specified.
@@ -158,27 +156,6 @@ data:
           accept_score: 25
           range: 0 25
           grader_flags: min
-
-# A testgroup may include all cases in another testgroup. All files for each
-# testcase in the other testgroup will be symlinked into the current testgroup.
-# Nested included testcases are also included here.
-#
-# This must be a list of testgroups and/or testcases paths relative to data/.
-# Testcases must be referred to by their {name}.
-#
-# You may only include testgroups that come before the current group, either
-# earlier in a list or earlier alphabetically in dictionaries.
-#
-# TODO: How does this interact with numbering? How do we guarantee included
-# cases come first?
-# include numbered into numbered: OK
-# include dict into dict:     error on name collision. How to guarantee they
-# come first?
-# include numbered into dict: error on name collision. Do we allow this?
-# include dict into numbered: error on name collision. Do we allow this?
-        include:
-          - sample
-          - secret/03
 
 # To enable automatic numbering of testcases, data: may also contain a list of
 # dictionaries instead of a single dictionary. In this case, testcases and/or


doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,110 @@
...
+      01.in: stdout.py  3    
+      02.in: stdout.cpp 4   # c++ is compiled as for validators
+      03.in: dir 5          # directories are OK as for validators
+      04.in: generator.py 5 # python files may include local files. They are executed locally.
+      05.in: generator.py 5 # python files may include local files. They are executed locally.
+
+# The entire value is passed through python shlex.shlex.
+      06-string.in: stdout.py "a b" # gives a single string of length 3 as argument
+
+# {seed.*} anywhere in the argument string will be replaced by a 31-bit integer
```
eldering, line 52: Can `{seed.*}` contain spaces or other weird characters? I.e. how would the following be interpreted:
    01.in: stdout.py {seed  < foo > /dev/null #}

eldering: Is seed an 31-bit unsigned or signed integer?

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,110 @@
...
+# hash of the entire command. Computed as:
+# int(hashlib.sha512((random_salt+command).encode('utf-8')).hexdigest(), 16)%(2**31)
+      07-random-1.in:    stdout.py {seed}
+      08-random-1a.in:   stdout.py {seed}    # same seed. May warn for duplicate case.
```
eldering, line 56: Would this comment actually be valid, and would it be considered part of the command and therefore generated a different seed? (Although the comment says it doesn't.)

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,110 @@
...
+      09-random-2.in:    stdout.py {seed:2}  # different seed, because of extra `2`
+      10-random-3.in:    stdout.py seed={seed:2} # different seed, because command isn't the same.
+
+# Multiple commands are piped to each other. The last stdout is used.
+      10.in:
```
eldering, line 61: I think it would be clarifying to add a key `pipe:` (or similar) here to make that more explicit. Also, that gives more flexibility to later extend the specification, if you'd like to do something else that is not a pipeline.

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,110 @@
...
+        - stdout.py 5
+        - double.py     # reads 5 from stdin and writes 10 to stdout
+
+# Commands may write files, but only the final stdout is used.
+      11.in:
+        - tree.py 10 tree.txt           # write tree.txt
+        - path.py 20 path.txt           # write path.txt
+        - combine.py tree.txt path.txt  # read tree.txt and path.txt and write testcase to stdout
```
eldering, line 69: Would be good to guarantee that these commands are run in a temporary directory so that the intermediate files `tree.txt` and `path.txt` cannot overwrite or conflict with files that might already be present elsewhere.

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,110 @@
...
+# Directories may contains a testdata.yaml that will be written as specified.
+# Future versions may lift the scoring config directly into the directory scope.
+        testdata.yaml:
+          on_reject: break
+          accept_score: 25
+          range: 0 25
+          grader_flags: min
+
+# A testgroup may include all cases in another testgroup. These will be
+# symlinked. May be a single value or a list of groups. Paths are absolute,
```
eldering, line 91: Does Windows FSs support symlinks? If it doesn't, would the files be copied as a fallback, or Windows simply not supported?

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,110 @@
...
+# be numbered automatically. Each list item must contain a 1-item dictionary.
+          - a.in: stdout.py a
+          - b.in: stdout.py b
+          # Equivalent to the map:
+          # 1_a.in: stdout.py a
+          # 2_b.in: stdout.py b
+
+# Introduce a configuration scope by adding a dictionary with `type: scope` set.
+      scope:
+        type: scope     # doesn't create a subdirectory
```
eldering, line 107: It's not clear to me what the scope of this configuration is. Also, why the duplicate `scope:` and `type: scope` (or is the first just a name that could also be `test-scope:` for example)? Is this to apply a configuration only to a subset of files/subdirectories in a particular directory?

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,110 @@
+# The following extensions will be created for each .in that's found.
+# Any key starting with a dot will be interpreted as an extension generator.
+# .ans will be generated before all others.
+.ans: /solutions/accepted/sol.py
+.png: /visualizers/asy.py # Note that directly calling a shell executable is not allowed
```
eldering, line 5: I guess you don't mean you can't call a shell script, but that you can't use shell syntax on this line. I.e. you can write
```
.png: /visualizers/my-bash-script-that-calls-other-stuff.sh
```
but you can't write
```
.png: /visualizers/preprocess.py | grep "foobar" | tac
```

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,117 @@
...
+      01.in: stdout.py  3    
+      02.in: stdout.cpp 4   # c++ is compiled as for validators
+      03.in: dir 5          # directories are OK as for validators
+      04.in: generator.py 5 # python files may include local files. They are executed locally.
+      05.in: generator.py 5 # python files may include local files. They are executed locally.
+
+# The entire value is passed through python shlex.shlex.
+      06-string.in: stdout.py "a b" # gives a single string of length 3 as argument
+
+# The regex \{seed(:[0-9]+)\} (e.g. {seed} or {seed:1}) anywhere in the argument
```
eldering, line 52: The regex should be `\{seed(:[0-9]+)?\}` with added `?`.

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,117 @@
...
+          # 2_b.in: stdout.py b
+
+# Introduce a configuration scope by adding a dictionary with `type: scope` set.
+# This configuration (partially) overrides the parent configuration and only
+# affects generators inside this scope.
+      no_visualizer_scope:
+        type: scope     # doesn't create a subdirectory
+        .png: # Disable .png generation for ne
+        data:
+          14.in: stdout.py 1234
```
RagnarGrootKoerkamp, line 117: instead, do this:

```
1-testcase:
  input: generate.py 
  solution: /submissions/accepted/a.py
  visualizer: /visualizers/viz.sh
```

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,117 @@
+# The following extensions will be created for each .in that's found.
```
RagnarGrootKoerkamp, line 1: name it generators.yaml

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+## Layout of generators/ directory:
+# - stdout.py <val>:             takes a value and writes it to stdout
+# - stdout.cpp <val>:            idem, but is a c++ file.
+# - dir/lib.h
+# - dir/generator.cpp <val>:     idem, but now includes a header. Works like validators.
+# - lib.py
+# - generator.py <val>:          uses lib.py. Fine because python will be executed locally.
+# - lib.sh
+# - generator.sh <val>:          uses lib.sh. Fine because shell will be executed locally.
```
RagnarGrootKoerkamp, line 21: if a generator has dependencies, do like so:

```
generators:
  tree.py: tree.py lib.py
  graph.py: tree.py lib.py
  a.cpp: a.cpp a.h
  b.cpp: b.cpp # optional
  c.py: c.py # optional
  d: d.cpp
  e: e.py
  new_name.py: old_name.py

  #a.py: # maybe at some point in the future
    #dependencies: a.py
    #build: command
    #run: command
```

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,221 @@
...
+# Absolute paths are taken relative to the problem root, relative paths are
+# taken relative to the generators directory.
+#
+# TOOLING: may pick a default if not specified, but should raise a warning.
+solution: /submissions/accepted/sol.py
+
+# The visualizer is used when no suitable image was generated already.
+# This should read {name}.in and/or {name}.ans from the current working
+# directory, and write {name}.ext for an extension in:
+# .png, .jpg, jpeg, .svg
```
eldering, line 17: Nitpick: why keep both `.jpg` and `.jpeg`? I would stick to `.jpg` unless there is a clear reason to also support `.jpeg`.

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,221 @@
...
+# and usually isn't required.
+visualizer: /visualizers/asy.py {name}
+
+# Optionally, a salt for generating the {seed} variables. Will be prepended to
+# the command being run.
+random_salt: abcd
+
+# We support three types of generators:
+# - Standalone files, like generators/a.cpp, generators/b.py, ..., which will
+#   be compiled and run the same way as submissions.
```
eldering, line 29: Add: be compiled *if required* and run

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,221 @@
...
+# - Directories, like 
+#   - generators/gen/tree.cpp
+#   - generators/gen/lib.h
+#   This will be compiled and run the same was as directory validators. Build
```
eldering, line 33: The specification for validators **requires** a `run` script/executable, which *may* be generated by an optional `build` script/executable. I think for consistency it would be good to stick to this.
This is anyways necessary to prevent ambiguity of how to build/run this.

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,221 @@
...
+#   and run scripts may be used.
+# - 'implicit' generators whose dependencies are specified in the generators:
+#   key below. The dependencies may refer to any files relative to generators/.
+#   The generator will be build and run as if they formed a separate directory.
```
eldering, line 37: Typo: buil**t**

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,221 @@
...
+#   E.g. the first example would be equivalent to
+#   - generators/tree/lib.py
+#   - generators/tree/tree.py
```
eldering, line 40: Would be nice to say "first example **below**", I missed it. Secondly, how do you define how to run this? Using `lib.py` or `tree.py`?

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,221 @@
...
+  b:
+    - b.cpp
+# It is allowed, but not required, to explicitly list single-file generators
+# as well. It is allowed to reuse the same name, but introducing a new name is
+# also fine.
+  c.py:
+    - c.py
+    - lib.py
+
+data:
```
eldering, line 71: For later documentation purposes, might be nice to specify that `data:` here is a keyword to specify what testcases can be generated, not an explicit reference to the `/data` directory in a problem package.

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,221 @@
...
+    data:
+
+# Types of generator programs.
+      01: stdout.py  3   # stdout of the command is written to 01.in.
+      02: stdout.cpp 4   # c++ is compiled as for validators
+      03: dir 5          # directories are OK as for validators
+      04: tree 5         # keys from the global generators: dictionary may also be used.
+      05: a 5            # idem
+
+# The entire command is passed through python shlex.shlex.
```
eldering, line 100: Would be nice to clarify this in non-python-specific terms: does this mean that `a b` is passed as a single token argument to the command `stdout.py`? What kind of quoting and escaping is possible/necessary.
Does any of this conflict with YAML escaping/quoting?

doc/generated_testcases_v2.yaml
```diff
@@ -0,0 +1,221 @@
...
+       #  1-b: stdout.py b
+       #  2-testgroup:
+       #    type: directory
+       #    data:
+       #      3-c: stdout.py c
+       #  4-d: stdout.py d
+       
+
+# At the directory level, unknown keys are allowed for tooling-specific
+# extensions.
```
eldering, line 220: I guess we'd also want to allow it at the global level, just like already existing keys like `solution` and `visualizer`.

doc/generated_testcases_v2.yaml
RagnarGrootKoerkamp: The intent is simple: to allow shell-like quoting of arguments, so that `a b` will be passed as a single argument indeed.

So the processing will be:

1. YAML comments are ignored
2. YAML parses the string. it seems to me that escaping is rather minimal in YAML: ([stackoverflow source](https://stackoverflow.com/a/33000792/2716069))
    - put the entire value in single quotes. now use a double single quote `''` to get `'`, but nothing else
    - put the entire value in double quotes. Now you can use the normal `\` escapes: `\\`, `\n`, ...
    - Use fold mode: `>` followed by the (multi-line) string value on the next line(s).
  So with YAML you'd never quote only part of the string, and quoting the entire string in single quotes or fold mode doesn't interfere with what's inside.
3. Shell-parse the entire value. [shlex.split](https://docs.python.org/3.7/library/shlex.html?highlight=shlex%20split#shlex.split) calls [shlex.shlex](https://docs.python.org/3.7/library/shlex.html?highlight=shlex%20split#shlex.shlex) in POSIX mode and 'tries to be as close as possible to the POSIX shell parsing rules', with the actual rules [described here](https://docs.python.org/3.7/library/shlex.html#parsing-rules).

    I'm not sure whether or not we should set `punctuation_chars` to `True` or `False`. See [here](https://docs.python.org/3.7/library/shlex.html#improved-compatibility-with-shells) for the difference.


Anyway, you can use `'`, `"` and `\` within `"` as you're used from normal shells.

doc/generated_testcases_v2.yaml
RagnarGrootKoerkamp: Hmm I don't think that's true, see [Programs](https://clics.ecs.baylor.edu/index.php?title=Problem_format#Programs). 
The table there also lists the main entrypoint to prevent ambiguity.

> Language is determined by looking at the file endings

is not very specific, but problemtools picks the language matching the most files, and breaks ties on priority, as [specified](https://github.com/Kattis/problemtools/blob/develop/problemtools/config/languages.yaml#L27-L42) in `languages.yaml`. I implemented the same for BAPCtools.

The entrypoint will be the single file, or a file matching `main`, or the lexicographically smallest file, see [here](https://github.com/Kattis/problemtools/blob/develop/problemtools/config/languages.yaml#L79-L86).

Thinking about it, that still doesn't really work for manually specified generators:
```
generators:
   trees:
      - trees.py
      - lib.py
```
would get `lib.py` as the entrypoint because `'lib.py' < 'trees.py'`. 
**What about making the first item in this list the entry point? I think that's quite natural anyway.**

Side note: this part of the spec already mentions 'POSIX-complient scripts' so we can probably refer to POSIX-shell parsing as well.

eldering: Hmmm... in DOMjudge we use these stricter rules about build/run scripts, and I thought these were copied into this spec. I really think this leaves too much ambiguity: what if the code in a directory might compile to two binaries? And I don't think that lexicographically first is natural enough to use as an implicit rule. If it was easy to have a single binary/entry point, then you could have used a single-file generator/validator in the first place.

For declaring the first element in the YAML list the entry point: that only works if the entry point is a file, but not for example if it is a Java class, of which you could have multiple per file. But since this method of specifying multi-file generators is really a convenience-method, I'd be willing to accept some ambiguity here by leaving the details up to the tooling.

doc/generated_testcases_v2.yaml
eldering: Ok, good to keep it the same as there, although I'd say that maybe there also `.jpeg` should be dropped.

doc/generated_testcases_v2.yaml
eldering: Ok, but from a YAML what is the value of the string:
`stdout.py "a b"`
or
`stdout.py a b`
?

RagnarGrootKoerkamp: YAML only uses quotes at the very start and end, so in this case the value will be `stdout.py "a b"`.

Right, we could execute the entire command through a shell, but that opens another can of worms. Now suddenly everything is _actual_ shell and you can use `|`, `;`, `&&`, `>`, `<`, ... . I definitely don't want that.
Also, it's better to be precise in the escaping we allow so we can have compatibility with windows.
I'm not sure how things work on Windows, but [subprocess.Popen](https://docs.python.org/3.7/library/subprocess.html?highlight=subprocess%20run#subprocess.Popen) has some info (see `shell=True`).

doc/generated_testcases_v2.yaml
RagnarGrootKoerkamp: Well, we've used DOMjudge in the past without `run` files:
```
bapc/jazzitup % tree output_validators 
output_validators
└── output_validator
    ├── output_validator.cpp
    └── validation.h
```

If you have a `build` file, you should probably have/build a `run` file as well, but having neither works fine as long as it's a single binary and the language is clear.

I think the usual case is still to have a single binary/executable/entry point with some library dependencies, where a directory doesn't work but auto-language-detection is still fine.

Current language detection puts the `{mainclass}` as the [filename without extension](https://github.com/Kattis/problemtools/blob/develop/problemtools/config/languages.yaml#L88-L90), and that should probably work in most cases.

PR #54, opened 2020-06-23
Update generators doc

RagnarGrootKoerkamp:
Update the documentation:

- Clarify how generators are built.
- Clarify that numbered cases are zero padded.
- Submission/visualiser paths must be absolute.
- Do not allow quoting of arguments. Instead always split on white space (python string.split())
- All dictionary keys must be strings.

@eldering @niemela

Implementation is in https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/55


doc/generators.md
```diff
@@ -20,25 +20,34 @@ The two main object types are `directory` and `generator`. The root of `generato
...
-* `generators`: The root directory may contain a dictionary listing generators and their dependencies. This may be used when using non-directory programs as generators which depend on other files in the `generators/` directory. Each key of the dictionary is the name of a generator, and values must be lists of file paths relative to `generators/`.
+    * A list of dictionaries as above. In this case, testcases will be prefixed with zero padded 1-based integers in the order of the list. Items in the same dictionary will get the same number.
 
 **Generator objects** have one of the following forms:
-- `null`: An empty generator means that the testcase is a manual case and must not be modified or deleted by generator tooling.
+- `null`: An empty generator means that the testcase is a manual case and must not be modified or deleted by generator tooling. The corresponding `.in` file must be present in the `data/` directory. The corresponding `.ans` may be present, but may also be generated from a given solution. Note that this form is discouraged. Prefer specifying a path to a `.in` file as below.
```
eldering, line 32: I would add something like `but may also be generated *once* from a given solution` to clarify that it won't be regenerated.

doc/generators.md
```diff
@@ -20,25 +20,34 @@ The two main object types are `directory` and `generator`. The root of `generato
...
+    - A path to a `.in` file which must be relative to `generators/`. The `.in` file and corresponding files with known extensions will be copied to the specified location. If a `.ans` is not specified and a `solution` is provided, it will be used to generate the `.ans`.
+    - An invocation of a generator: `<generator_name> <arguments>`. `<generator_name>` must either be a program (file/directory) in `generators/` or else a name in the top level `generators` dictionary (see below). Arguments may contain `{name}` to refer to the name of the testcase and `{seed}` or `{seed:(0-9)+}` to add a random seed. Arguments are separated by white space (space, tab, newline). Quoting white space is not supported.
+
 - A dictionary containing `type: testcase`. In this case, `input` is a `command` as above, and the dictionary may furthermore contain the `solution`, `visualizer`, and `random_salt` keys to specialize them for this testcase only.
 
+**Root object**
+The root of the `generators.yaml` is a `directory` object with one additional key:
```
eldering, line 40: Add `one *optional* additional key`.

PR #74, opened 2020-09-05
Fix progressbar for Windows command prompt

jbransen:
As title.

The problem is that the `\r` does not work well when the line is completely full, because then the next character, even though it is a carriage returns, already wraps to the next line. The fix is simply to make the progressbar 1 character shorter.


RagnarGrootKoerkamp:
nice :+1:

PR #75, opened 2020-09-05
Use colorama for colored output

jbransen:
Instead of implementing terminal colors in a custom way, this uses the [colorama](https://pypi.org/project/colorama/) library, which makes the code a bit cleaner and (most importantly) works on all platforms, specifically supporting Windows too.

As a small change the orange text now changed to yellow because orange is not available, but IMO it looks fine.

### DIFF
--- a/bin/tools.py
+++ b/bin/tools.py
@@ -43,6 +43,8 @@ from util import *
 if not is_windows():
     import argcomplete  # For automatic shell completions
 
+# Initialize colorama for printing coloured output. On Windows, this captures
+# stdout and replaces ANSI colour codes by calls to change the terminal colour.
 colorama.init()
 
 # List of high level todos:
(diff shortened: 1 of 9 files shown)


bin/tools.py
```diff
@@ -42,6 +43,8 @@
 if not is_windows():
     import argcomplete  # For automatic shell completions
 
+colorama.init()
```
RagnarGrootKoerkamp, line 46: Can you add a comment this is needed for stdout wrapping on windows?

RagnarGrootKoerkamp (approved):
Hmmm, adding a new dependency isn't great, but if we're ever doing colors on Windows, something like this is probably the way to go, so lets just do it. Now might be a good time for me to make some linux packages though ;)

Also, the orange you see in the screencasts is actually the way my terminal renders yellow.

PR #91, opened 2020-12-27
Allow having no output format validators when validation=default

mpsijm:
- Note about lines 233-234: treating `output` and `output_format` the same is now done by setting `valitator_type = output` instead of the other way around, because then the condition on line 262 actually makes sense.
  - This change also has effect on line 269.
- Because the returned `validators` list is now allowed to be empty, I had to change the condition on line 476 from `not validators` to `validators not is False`, because the empty list is a false-y value in Python.

This works on some FPC problems that have `validation=default` (some of them having output format validators), and I ran `bapctools validate` on the two BAPC contests as well, seems to work on my machine :tm: :stuck_out_tongue:

### DIFF
--- a/bin/problem.py
+++ b/bin/problem.py
@@ -259,10 +259,13 @@ class Problem:
         paths = (glob(problem.path / (validator_dir + '_validators'), '*') +
                  glob(problem.path / (validator_dir + '_format_validators'), '*'))
 
-        if len(paths) == 0 and validator_type != 'output_format':
-            error(f'No {validator_type} validators found.')
-            problem._validators[key] = False
-            return False
+        if len(paths) == 0:
+            if validator_type == 'output_format':
+                log(f'No {validator_type} validators found (not required).')
+            else:
+                error(f'No {validator_type} validators found, at least one is required.')
+                problem._validators[key] = False
+                return False
         if validator_type == 'output' and problem.interactive and len(paths) > 1:
             error(
                 f'Found more than one output validator, but validation type {problem.settings.validation} needs exactly one.'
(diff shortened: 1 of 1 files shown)


bin/problem.py
```diff
@@ -260,11 +260,13 @@ def validators(problem, validator_type, check_constraints=False):
                  glob(problem.path / (validator_dir + '_format_validators'), '*'))
 
         if len(paths) == 0:
-            if validator_type != 'output_format':
-                warn(f'No {validator_type} validators found.')
-            problem._validators[key] = False
-            return False
-        if validator_type == 'output_format' and problem.interactive and len(paths) > 1:
+            if validator_type == 'output_format':
+                log(f'No {validator_type} validators found (not required).')
```
RagnarGrootKoerkamp, line 264: not needed to add this, since it will be clear to the user this is a LOG and not WARN or ERROR.
```suggestion
                log(f'No {validator_type} validators found.')
```

bin/problem.py
```diff
@@ -471,7 +473,7 @@ def validate_format(problem, validator_type, constraints=None):
             log('Not validating .ans for interactive problem.')
             return True
 
-        if not validators:
+        if validators is False:
```
RagnarGrootKoerkamp, line 476: Hrmm; I'm a bit conflicted on this: in this case, the `problem.validators()` function will already give an error, so it could also just return `[]` instead of `False`. That way, we wouldn't have to catch the `false` here and could just loop over the empty list. The `validate_format` function would always return true in that case, but that's not really an issue anyway I think, since calling `error` at least once makes the program return non-zero already.

But anyway, that's for another PR.

RagnarGrootKoerkamp (changes requested):
LGTM, thanks! Just one small suggestion.

PR #98, opened 2021-04-04
[latex] Switch to `latexmk` and add `--watch` argument

mpsijm:
Wanted my PDFs to autocompile on changes, so I added a `--watch` argument to `bt pdf` and `bt solutions` :smile:

To implement this, I switched to `latexmk`, which (according to its `man` page) "automates the process of compiling a LaTeX document", and should be included in recent `texlive` installations. It calls `pdflatex` as often as is required (instead of the fixed three times, as it was until now), and supports continuous previews, which saves the effort of having to implement file watching from Python. This has as downside that it only tracks changes in the TeX files and does not track changes in `*.yaml` config files, but I'm okay with that :slightly_smiling_face:

### DIFF
--- a/bin/latex.py
+++ b/bin/latex.py
@@ -143,30 +143,36 @@ def get_environment():
 
 def build_latex_pdf(builddir, tex_path, problem_path=None):
     env = get_environment()
-    run_once = getattr(config.args, '1', False)
-    watch = getattr(config.args, 'watch', False)
+
+    latexmk_command = ['latexmk', '-cd', '-g', '-pdf',
+                       '-pdflatex=pdflatex -interaction=nonstopmode -halt-on-error']
+    if getattr(config.args, 'watch', False):
+        latexmk_command.append("-pvc")
+    if getattr(config.args, '1', False):
+        latexmk_command.extend(['-e', '$max_repeat=1'])
+    latexmk_command.extend([f'-output-directory={builddir}', tex_path])
+
     ret = util.exec_command(
-        ['latexmk', '-cd', '-g', '-pdf',
-         '-pdflatex=pdflatex -interaction=nonstopmode -halt-on-error']
-        + (['-pvc'] if watch else [])
-        + (['-e', '$max_repeat=1'] if run_once else [])
-        + [f'-output-directory={builddir}', tex_path],
-        0,
-        False,
+        latexmk_command,
+        expect=0,
+        crop=False,
         cwd=builddir,
         stdout=subprocess.PIPE,
         env=env,
         timeout=None,
     )
+
     if ret.ok is not True:
         print(f'{Fore.RED}Failure compiling pdf:{Style.RESET_ALL}\n{ret.out}')
         error(f'return code {ret.ok}')
         error(f'duration {ret.duration}')
         return False
+
     # link the output pdf
     output_pdf = Path(tex_path.name).with_suffix('.pdf')
     dest_path = output_pdf if problem_path is None else problem_path / output_pdf
     ensure_symlink(dest_path, builddir / output_pdf, True)
+
     print(f'{Fore.GREEN}Pdf written to {dest_path}{Style.RESET_ALL}')
     return True
 
@@ -178,7 +184,7 @@ def build_latex_pdf(builddir, tex_path, problem_path=None):
 # 4. Create tmpdir/<problem>/samples.tex.
 # 5. Run latexmk and link the resulting problem.pdf into the problem directory.
 def build_problem_pdf(problem, solutions=False):
-    main_file = f'{"solution" if solutions else "problem"}.tex'
+    main_file = 'solution.tex' if solutions else 'problem.tex'
     prepare_problem(problem)
 
     builddir = problem.tmpdir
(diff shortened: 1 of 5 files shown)


bin/latex.py
```diff
@@ -144,20 +141,50 @@ def get_environment():
...
 # 1. Copy the latex/problem.tex file to tmpdir/<problem>/problem.tex,
 # substituting variables.
 # 2. Link tmpdir/<problem>/problem_statement to the problem problem_statement directory.
 # 3. Link bapc.cls
 # 4. Create tmpdir/<problem>/samples.tex.
-# 5. Run pdflatex and link the resulting problem.pdf into the problem directory.
+# 5. Run latexmk and link the resulting problem.pdf into the problem directory.
 def build_problem_pdf(problem, solutions=False):
-    t = 'solution' if solutions else 'problem'
+    main_file = f'{"solution" if solutions else "problem"}.tex'
```
RagnarGrootKoerkamp, line 181:
```suggestion
    main_file = 'solution.tex' if solutions else 'problem.tex'
```

bin/latex.py
```diff
@@ -144,20 +141,50 @@ def get_environment():
...
+    env = get_environment()
+    run_once = getattr(config.args, '1', False)
+    watch = getattr(config.args, 'watch', False)
+    ret = util.exec_command(
+        ['latexmk', '-cd', '-g', '-pdf',
+         '-pdflatex=pdflatex -interaction=nonstopmode -halt-on-error']
+        + (['-pvc'] if watch else [])
+        + (['-e', '$max_repeat=1'] if run_once else [])
+        + [f'-output-directory={builddir}', tex_path],
+        0,
```
RagnarGrootKoerkamp, line 154: Could you comment what these variables mean? (I know it's copy paste, but I forgot :P )

bin/latex.py
```diff
@@ -144,20 +141,50 @@ def get_environment():
     return env
 
 
+def build_latex_pdf(builddir, tex_path, problem_path=None):
+    env = get_environment()
+    run_once = getattr(config.args, '1', False)
+    watch = getattr(config.args, 'watch', False)
+    ret = util.exec_command(
+        ['latexmk', '-cd', '-g', '-pdf',
+         '-pdflatex=pdflatex -interaction=nonstopmode -halt-on-error']
```
RagnarGrootKoerkamp, line 150: This would be easier to parse (as a human) with double quotes around the 'argument' part (after the `=`). Can you add that, if it works at all.

bin/latex.py
```diff
@@ -144,20 +141,50 @@ def get_environment():
...
+        + (['-pvc'] if watch else [])
```
RagnarGrootKoerkamp, line 151: maybe refactor the command to a temp variable and do `if watch: command.append('-pvc')`.
(But if it doesn't look cleaner in the end; don't bother.)

doc/commands.md
```diff
@@ -193,9 +194,13 @@ See the [Implementation notes](implementation_notes.md#building-latex-files) for
 
 **Flags**
 
-- `--order`: The order of the problems, e.g. `BDCA`. Can be used to order problems from easy to difficult. When labels have multiple letters, `B1,A1,A2,B2` is also allowed.
 - `--cp`: Instead of symlinking the final pdf, copy it into the contest directory.
+- `--order`: The order of the problems, e.g. `BDCA`. Can be used to order problems from easy to difficult. When labels have multiple letters, `B1,A1,A2,B2` is also allowed.
+- `--order-from-ccs`: Order the problems by increasing difficulty, extracted from the api, e.g.: https://www.domjudge.org/demoweb. Defaults to value of ccs_url in contest.yaml.
```
RagnarGrootKoerkamp, line 199: :+1:

RagnarGrootKoerkamp (commented):
Good stuff! Including documentation updates even :D

The only thing that I'm not sure about is whether latexmk is also standard/easily available on Windows. Could you check?

bin/latex.py
mpsijm: And with "these variables" you mean the parameters `0` and `False` passed to `util.exec_command`? I "commented" them by passing them as keyword arguments.

bin/latex.py
mpsijm: I had tried this already, but it doesn't work :upside_down_face: 

In a shell, when you would write `-pdflatex="pdflatex -options..."`, the shell interprets this by appending the string to the rest of the argument, and it would basically be the same as writing `-pdflatex=pdflatex\ -options...`. Thus, when using a shell, the quotes are never passed to the `latexmk` command, since they are interpreted by the shell. Therefore, writing extra quotes on this line in Python will break stuff.

bin/latex.py
mpsijm: It looks slightly more Pythonic with your proposed refactoring, thanks for the suggestion! :slightly_smiling_face:

doc/commands.md
mpsijm: I noticed these were missing, so I just copied the descriptions from the `--help` command :joy:

mpsijm:
> The only thing that I'm not sure about is whether latexmk is also standard/easily available on Windows. Could you check?

Of course, this depends on how someone installed their LaTeX tools under Windows. TeX Live includes the `latexmk` binary, and TeX Live can also be installed in Windows, so I guess it's fine? I haven't actually tried on Windows, though.

PR #100, opened 2021-04-06
Fix #92 stderr output

### LINKED ISSUE #92: Print logs to stderr instead of stdout
I was running `bt gitlabci > .gitlab-ci.yml` with an empty `problems.yaml`, and the command produced no output but did exit with code 1. I couldn't figure out what happened until I found that the `.gitlab-ci.yml` file contained `FATAL ERROR: Did not find any problem in problems.yaml.`

My suggestion is to print errors/warnings/logs/etc. to stderr instead of stdout to avoid situations like this. Only results of subcommands (like `bt gitlabci`, `bt sort`, and `bt tmp`) should be printed to stdout.

What do you think?

RagnarGrootKoerkamp:
@mpsijm

### DIFF
--- a/bin/constraints.py
+++ b/bin/constraints.py
@@ -19,7 +19,7 @@ def check_constraints(problem, settings):
     ans_constraints = {}
     problem.validate_format('input_format', constraints=in_constraints)
     problem.validate_format('output_format', constraints=ans_constraints)
-    print(file=sys.stderr)
+    print()
 
     vinput = problem.path / 'input_validators/input_validator/input_validator.cpp'
     voutput = problem.path / 'output_validators/output_validator/output_validator.cpp'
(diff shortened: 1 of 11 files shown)


bin/constraints.py
```diff
@@ -105,39 +106,41 @@ def f(cs):
     left_width = 8 + name_len + 2 * value_len
 
     print('{:^{width}}|{:^30}'.format('VALIDATORS', '      PROBLEM STATEMENT', width=left_width),
-          sep='')
+          sep='',
+          file=sys.stderr)
```
mpsijm, line 110: I would say that the `constraints` command should show its output on the standard output stream, as it falls in the category of "actual 'program output' and not progress/diagnostics" (https://github.com/RagnarGrootKoerkamp/BAPCtools/issues/92#issuecomment-751534349).

bin/stats.py
```diff
@@ -63,7 +64,7 @@ def stats(problems):
             format_string += ' {:>' + str(width + len(Fore.WHITE) + len(Style.RESET_ALL)) + '}'
 
     header = header_string.format(*headers)
-    print(Style.BRIGHT + header + Style.RESET_ALL)
+    print(Style.BRIGHT + header + Style.RESET_ALL, file=sys.stderr)
```
mpsijm, line 67: The output of the `stats` command would also be printed to stdout, right? (https://github.com/RagnarGrootKoerkamp/BAPCtools/issues/92#issuecomment-752060134)

bin/latex.py
```diff
@@ -164,7 +165,7 @@ def build_latex_pdf(builddir, tex_path, problem_path=None):
     )
 
     if ret.ok is not True:
-        print(f'{Fore.RED}Failure compiling pdf:{Style.RESET_ALL}\n{ret.out}')
+        print(f'{Fore.RED}Failure compiling pdf:{Style.RESET_ALL}\n{ret.out}', file=sys.stderr)
```
mpsijm, line 168: Maybe some of the `print(..., file=sys.stderr)` can be replaced with calls to the logger functions of `util.py`? For example, in this case, a call to `error` could be used, and line 178 in this file could be a call to `log`.

mpsijm (changes requested):
Thanks for picking this up! Luckily, BAPCtools only has 84 calls to `print`, so I also double-checked the ones that you didn't change :slightly_smiling_face:

bin/stats.py
RagnarGrootKoerkamp: I figured there's no point currently. All stylized output (bold/colours) won't be easily parseable anyway, so there's no point in having it on stdout.

bin/constraints.py
RagnarGrootKoerkamp: works for me; i was in doubt for this one

bin/latex.py
RagnarGrootKoerkamp: done - the remaining uses for `Fore` and `Style` all do some kind of custom styling so can't be replaced by log/error.

PR #114, opened 2021-05-09
Fix display of leading/multiple spaces in samples

mpsijm:
When creating the problem "Fraud Checking" for the Freshmen Programming Contest 2021 (archived at https://commissies.ch.tudelft.nl/chipcie/?page=archief), we noticed that leading/multiple spaces are not displayed correctly in the Sample Inputs/Outputs. This PR fixes this in two parts:

- Leading spaces were only displayed in samples when they were at the start of the first line. To also display them at the start of following lines, I've added the flag `re.MULTILINE` (line 55).
- Multiple spaces were globbed together by the TeX compiler, so I replaced `'  +'` with a `\phantom{...}` command, with as many `.`s as the number of spaces (line 53).

I decided to do the multiple-spaces-check _before_ the leading-spaces-check, so that a line with multiple leading spaces will only get one `\phantom` command at the start.

### DIFF
--- a/bin/latex.py
+++ b/bin/latex.py
@@ -49,9 +49,10 @@ def tex_escape(text):
     if has_newline:
         text = text[:-1]
     text = regex.sub(lambda match: conv[match.group()], text)
+    # Escape multiple spaces separately
+    text = re.sub('  +', lambda m: '\\phantom{' + ('.' * len(m.group())) + '}', text)
     # Escape leading spaces separately
-    regex = re.compile('^ ')
-    text = regex.sub('\\\\phantom{.}', text)
+    text = re.compile('^ ', re.MULTILINE).sub('\\\\phantom{.}', text)
     if has_newline:
         text += '\n'
     return text


bin/latex.py
```diff
@@ -49,9 +49,10 @@ def tex_escape(text):
     if has_newline:
         text = text[:-1]
     text = regex.sub(lambda match: conv[match.group()], text)
+    # Escape multiple spaces separately
+    text = re.sub('  +', lambda m: '\\phantom{' + ('.' * len(m.group())) + '}', text)
     # Escape leading spaces separately
-    regex = re.compile('^ ')
-    text = regex.sub('\\\\phantom{.}', text)
+    text = re.compile('^ ', re.MULTILINE).sub('\\\\phantom{.}', text)
```
RagnarGrootKoerkamp, line 55: Maybe the re.compile could be extracted to a const global here, but since it's only used once (/few times) per invocation, and actually not invoked most of the time (eg `bt run`), it's probably better as is.

RagnarGrootKoerkamp (commented):
It's not pretty; this entire `tex_escape` function is a mess.

Maybe we should get some inspiration from problemtools ;)
https://github.com/Kattis/problemtools/blob/develop/problemtools/templates/latex/problemset.cls#L280-L281

IIRC `lstinputlisting` also has a form that takes text from the latex instead of a file, although that probably still has escape issues.
Even nicer would be to rework things so that we can actually use `lstinputlisting` itself and pass file path arguments. That may be a bit more work, but actually shouldn't be too bad:
The sample .in/.ans files already exist somewhere, so the `samples.tex` that's created can just replace each of the two columns by a `\lstinputlisting{/absolute/path/to/in_or_ans}`.

Anyway, this is good to merge now, but maybe you'd like to try to fix this in a separate PR ;)

bin/latex.py
mpsijm: I agree with you. It would be a trade-off between code readability and efficiency. Since this `tex_escape` function is "only" called once for every sample file (my guess would be around 42 times per invocation of `bt pdf` in an average contest), I also prefer to keep the regexes in-lined for code readability :slightly_smiling_face: 

(also, rewriting this right now wouldn't make much sense if we're planning to use Kattis' approach with `lstinputlisting`s later :stuck_out_tongue:)

PR #156, opened 2021-11-01
[download_scoreboard.sh] Add quotes around variables

mpsijm:
This broke when my password contained some special characters (like '{' or '(').


RagnarGrootKoerkamp:
Thanks

The moment where this should be rewritten in python is approaching ;)

PR #157, opened 2021-11-02
Fix #134: Testsession lettering from end of alphabet

### LINKED ISSUE #134: Fix testsession lettering from the end when problems.yaml is present
currently it starts at A but it shouldn't do this

mpsijm:
This fix consists of two parts:
- When running a command for the full contest (e.g., `bt pdf` in the contest root) while there is no `problems.yaml` file present.
  - In this case, start counting at chr(ord('Z') - len(problems) + 1).
- When calling `bt new_problem` with an empty `problems.yaml` file.
  - In this case, start counting at 'X'.

### DIFF
--- a/bin/contest.py
+++ b/bin/contest.py
@@ -33,12 +33,6 @@ def contest_yaml():
     return _contest_yaml
 
 
-def next_label(label):
-    if label is None:
-        return 'X' if (contest_yaml() or {}).get('testsession') else 'A'
-    return label[:-1] + chr(ord(label[-1]) + 1)
-
-
 _problems_yaml = None
 
 
--- a/bin/contest.py
+++ b/bin/contest.py
@@ -35,7 +35,7 @@ def contest_yaml():
 
 def next_label(label):
     if label is None:
-        return 'A'
+        return 'X' if (contest_yaml() or {}).get('testsession') else 'A'
     return label[:-1] + chr(ord(label[-1]) + 1)
 
 
--- a/bin/contest.py
+++ b/bin/contest.py
@@ -17,7 +17,7 @@ def contest_yaml():
     if _contest_yaml:
         return _contest_yaml
     if _contest_yaml is False:
-        return None
+        return {}
 
     path = None
     # TODO: Do we need both here?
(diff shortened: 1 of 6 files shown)


bin/skel.py
```diff
@@ -116,7 +116,7 @@ def new_problem():
         validation = _ask_variable('validation (default/custom/custom interactive)', 'default')
 
     # Read settings from the contest-level yaml file.
-    variables = read_yaml_settings(Path('contest.yaml'))
+    variables = contest.contest_yaml() or {}
```
RagnarGrootKoerkamp, line 119: Maybe `contest_yaml()` should just return `{}` instead of `None` when the file isn't found

bin/contest.py
```diff
@@ -35,7 +35,7 @@ def contest_yaml():
 
 def next_label(label):
```
RagnarGrootKoerkamp, line 36: This is only used in skel.py now maybe move it there.

I was thinking: It would be cool to start at Z, and every time you add a problem all labels are shifted down by 1 and the new problem is appended. But maybe that's too complicated so just starting at X also works fine; we always have 3 problems anyway.

bin/skel.py
mpsijm: Yeah, I was reluctant to do this, because some usages of `contest_yaml()` have different behaviour when there is no contest config. But I can take a look at this to see if they can be rewritten to simply check for `{}` instead (good thing that `bool({}) == False` in Python :joy:)

bin/contest.py
mpsijm: Good idea :slightly_smiling_face: 

Yeah I was thinking about that too, but that would then require prepending problems to `problems.yaml` (instead of appending), that could become messy. :joy: I also wonder what would now happen if you would add a fourth problem to a test session (or, for that matter, a 27th problem to a regular contest). The label will probably become `'['` or something :stuck_out_tongue: But I assume we don't have to add corner cases for that, if it won't happen during regular use :innocent:

mpsijm: Sounds good :slightly_smiling_face: 

I was also thinking about that, but I thought about prepending problems to `problems.yaml` (instead of appending them), so I didn't do it. Shifting down the labels would make sense, assuming that they follow the regular naming scheme. And any other cases probably fall outside of regular use, so we probably won't need to add corner cases for that. :innocent:

RagnarGrootKoerkamp: Thinking about it more; I'd prefer to keep it simple (i.e. what you did now). This is all quite rare anyway, and it would significantly complicate the case. Also, labels like `[X1, X2, Y]` are possible, and shifting those down is not good.

PR #158, opened 2021-11-02
Fix #149: Remove indent from `ruamel.yaml` written `problems.yaml`

### LINKED ISSUE #149: Remove indent from `ruamel.yaml` written `problems.yaml`

mpsijm:
The reason why ruamel.yaml adds two extra spaces at the start of each line, is the indent config (sequence=4, offset=2).
Either this indent config should be made inconsistent with the indent configs used on other places, or we post-process the yaml to remove the leading spaces (which I've done in this commit).

### DIFF
--- a/bin/skel.py
+++ b/bin/skel.py
@@ -160,9 +160,13 @@ def new_problem():
             ryaml.dump(
                 data,
                 problems_yaml,
-                # Remove spaces at the start of each line, caused by the indent configuration.
+                # Remove spaces at the start of each (non-commented) line, caused by the indent configuration.
                 # If there was a top-level key (like `problems:`), this wouldn't be needed...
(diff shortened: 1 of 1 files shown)


RagnarGrootKoerkamp:
Ugh; it's sooo ugly - but it seems that this is actually the recommended fix by the Ruamel.yaml author: https://stackoverflow.com/questions/58770666/how-to-remove-2-spaces-from-the-output-of-ruamel-yaml-dump

Maybe add the thing about checking that the line starts with two spaces though, to not remove non-indented comments.

mpsijm:
Yeah, I know, it's horrible :sweat_smile: 

Good point actually, I will add that check later today :slightly_smiling_face:

PR #160, opened 2021-11-03
Update generators.yaml skel

RagnarGrootKoerkamp:
@mpsijm, what do you think?

### DIFF
--- a/headers/validation.h
+++ b/headers/validation.h
@@ -61,10 +61,11 @@ inline std::string location_to_string(source_location loc) {
 	return std::string(loc.file_name()) + ":" + std::to_string(loc.line());
 }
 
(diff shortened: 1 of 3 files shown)


mpsijm:
The explanation of the `input_validator --generate` features is looking good to me :smile: Maybe you can still expand on how to add manual cases? :slightly_smiling_face: 

Also, I notice that you only advertise numbered test cases now, is this intended?

RagnarGrootKoerkamp:
Right; I'll add some text on manual cases.

I think my plan is to nudge more towards numbered cases - it's a bit
more complicated tool-wise, but more convenient in the end. Seeing all
this `random-1` to `random-9` typed out is ugly.

I'll make a separate issue that listing all cases in a directory should
be enforced, which should hopefully make things simpler.

PR #163, opened 2021-11-08
[config] Fix #132: Personal config file

### LINKED ISSUE #132: Personal config file
This could be used to make the `pdf --cp` flag on by default.

mpsijm:
Fixes #132.

This PR adds the ability to have a personal config file, located at `$XDG_CONFIG_HOME/bapctools/config.yaml`¹ or `$CONTEST_DIR/.bapctools.yaml`². The keys in this config file can be any option that can be passed on the command-line. Note that the keys should be written out in full (e.g., `username:` rather than `u:`) and any hyphens should be replaced with an underscore (e.g., `no_bar:` rather than `no-bar:`)³.

In preparation for this PR, I have made sure that all `config.args` are read consistently in the same way, being "the value is `None` if it is not set" (i.e., `hasattr` or `key in args` checks are no longer necessary). Three keys have a different default value than `None`, listed in `config.default_args`.

Additionally, I made sure that the `args` object returned by the argument parser _only_ contains the keys set on the command line, using [`argument_default=SUPPRESS`](https://stackoverflow.com/a/62789053). Among other things, this means that `store_true` flags no longer set to `False` in the `args` object by default. Because of this, the function that reads the personal config file(s) can confidently check which values are set on the command line, so that other keys can use values from the personal configuration file(s) or the default value, in that order. This results in the following priority order for arguments (from low to high):

```
default value
$XDG_CONFIG_HOME/bapctools/config.yaml
$CONTEST_DIR/.bapctools.yaml
command line flag value
```

TODO:
- [ ] @RagnarGrootKoerkamp I added a short description of this feature to `readme.md`, but perhaps it would also be good to add something to `doc/implementation_notes.md`, but I am not sure what would be the best place for that :stuck_out_tongue: 

¹) On Unix-ish machines, this defaults to `~/.config`. On Windows machines, we use `%AppData%` instead.
²) Note that, because the personal config file(s) are queried before `cd`-ing to the contest's root directory, placing a `.bapctools.yaml` file in a problem directory will also work, but only when running a command directly from this problem directory.
³) `argparse` automatically converts hyphens to underscores, which automatically means that keys in the personal config file should all use underscores as well. Note that, currently, some command-line arguments use hyphens, while others use underscores. We might want to make that consistent (but I'm not sure in which direction) :stuck_out_tongue:

### DIFF
--- a/bin/config.py
+++ b/bin/config.py
@@ -58,13 +58,42 @@ os.environ["PATH"] += os.pathsep + str(tools_root / 'third_party')
 
 # Below here is some global state that will be filled in main().
 
-# Some default values needed for tests.
 args = argparse.Namespace()
-args.testcases = None
-args.cpp_flags = None
-args.verbose = None
-args.force_build = None
-args.error = None
+
+
+def set_default_args():
+    # Set default argument values
+    for arg in [
+        '1',
+        'add_manual',
+        'all',
+        'check_deterministic',
+        'clean',
+        'clean_generated',
+        'cpp_flags',
+        'contest_id',
+        'default_solution',
+        'ignore_validators',
+        'interaction',
+        'memory',
+        'move_manual',
+        'move_to',
+        'no_bar',
+        'no_generate',
+        'no_timelimit',
+        'order',
+        'order_from_ccs',
+        'remove',
+        'samples',
+        'scoreboard_repo',
+        'submissions',
+        'table',
+        'testcases',
+        'timeout',
+        'watch',
+    ]:
+        if not hasattr(args, arg):
+            setattr(args, arg, None)
 
 
 level = None
--- a/bin/config.py
+++ b/bin/config.py
@@ -71,65 +71,16 @@ default_args = {
 """
 for cmd in $(bapctools --help | grep '^  {' | sed 's/  {//;s/}//;s/,/ /g') ; do bapctools $cmd --help ; done |& \
 grep '^  [^ ]' | sed 's/^  //' | cut -d ' ' -f 1 | sed -E 's/,//;s/^-?-?//;s/-/_/g' | sort -u | \
-grep -Ev '^(h|jobs|time|verbose)$' | sed "s/^/        '/;s/$/',/"
+grep -Ev '^(h|jobs|time|verbose)$' | sed "s/^/'/;s/$/',/" | tr '\n' ' ' | sed 's/^/args_list = [/;s/, $/]\n/'
 """
+# fmt: off
+args_list = ['1', 'add_manual', 'all', 'api', 'author', 'check_deterministic', 'clean', 'clean_generated', 'cleanup_generated', 'contest', 'contest_id', 'contestname', 'cp', 'cpp_flags', 'default_solution', 'directory', 'error', 'force', 'force_build', 'ignore_validators', 'input', 'interaction', 'interactive', 'kattis', 'memory', 'move_manual', 'move_to', 'no_bar', 'no_generate', 'no_solutions', 'no_timelimit', 'order', 'order_from_ccs', 'output', 'password', 'problem', 'problemname', 'remove', 'samples', 'scoreboard_repo', 'skel', 'skip', 'submissions', 'table', 'testcases', 'timelimit', 'timeout', 'username', 'validation', 'watch', 'web']
+# fmt: on
 
 
 def set_default_args():
     # Set default argument values.
-    for arg in [
-        '1',
-        'add_manual',
-        'all',
-        'api',
-        'author',
-        'check_deterministic',
-        'clean',
-        'clean_generated',
-        'cleanup_generated',
-        'contest',
-        'contest_id',
-        'contestname',
-        'cp',
-        'cpp_flags',
-        'default_solution',
-        'directory',
-        'error',
-        'force',
-        'force_build',
-        'ignore_validators',
-        'input',
-        'interaction',
-        'interactive',
-        'kattis',
-        'memory',
-        'move_manual',
-        'move_to',
-        'no_bar',
-        'no_generate',
-        'no_solutions',
-        'no_timelimit',
-        'order',
-        'order_from_ccs',
-        'output',
-        'password',
-        'problem',
-        'problemname',
-        'remove',
-        'samples',
-        'scoreboard_repo',
-        'skel',
-        'skip',
-        'submissions',
-        'table',
-        'testcases',
-        'timelimit',
-        'timeout',
-        'username',
-        'validation',
-        'watch',
-        'web',
-    ]:
+    for arg in args_list:
         if not hasattr(args, arg):
             setattr(args, arg, None)
     for arg, value in default_args.items():
--- a/bin/config.py
+++ b/bin/config.py
@@ -62,7 +62,7 @@ args = argparse.Namespace()
 
 default_args = {
     'jobs': os.cpu_count() // 2,
-    'time': 600,
+    'time': 600,  # Used for `bt fuzz`
     'verbose': 0,
 }
 
--- a/bin/config.py
+++ b/bin/config.py
@@ -75,13 +75,6 @@ n_error = 0
 n_warn = 0
 
 
-# Return the command line timeout or the default of 30 seconds.
-def timeout():
-    if hasattr(args, 'timeout'):
-        return args.timeout
-    return 30
-
-
 # Set to true when running as a test.
 RUNNING_TEST = False
 TEST_TLE_SUBMISSIONS = False
--- a/bin/config.py
+++ b/bin/config.py
@@ -67,21 +67,33 @@ default_args = {
 }
 
 
+# The list of arguments below is generated using the following command:
+"""
+for cmd in $(bapctools --help | grep '^  {' | sed 's/  {//;s/}//;s/,/ /g') ; do bapctools $cmd --help ; done |& \
+grep '^  [^ ]' | sed 's/^  //' | cut -d ' ' -f 1 | sed -E 's/,//;s/^-?-?//;s/-/_/g' | sort -u | \
+grep -Ev '^(h|jobs|time|verbose)$' | sed "s/^/        '/;s/$/',/"
+"""
+
+
 def set_default_args():
-    # Set default argument values
+    # Set default argument values.
     for arg in [
         '1',
         'add_manual',
         'all',
+        'api',
+        'author',
         'check_deterministic',
         'clean',
         'clean_generated',
         'cleanup_generated',
-        'cp',
-        'cpp_flags',
         'contest',
         'contest_id',
+        'contestname',
+        'cp',
+        'cpp_flags',
         'default_solution',
+        'directory',
         'error',
         'force',
         'force_build',
@@ -100,7 +112,9 @@ def set_default_args():
         'order',
         'order_from_ccs',
         'output',
+        'password',
         'problem',
+        'problemname',
         'remove',
         'samples',
         'scoreboard_repo',
@@ -111,6 +125,8 @@ def set_default_args():
         'testcases',
         'timelimit',
         'timeout',
+        'username',
+        'validation',
         'watch',
         'web',
     ]:
--- a/bin/config.py
+++ b/bin/config.py
@@ -60,6 +60,12 @@ os.environ["PATH"] += os.pathsep + str(tools_root / 'third_party')
 
 args = argparse.Namespace()
 
+default_args = {
+    'jobs': os.cpu_count() // 2,
+    'time': 600,
+    'verbose': 0,
+}
+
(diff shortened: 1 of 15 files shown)


bin/problem.py
```diff
@@ -107,17 +107,9 @@ def _read_settings(self):
...
-        timeout = 1.5 * self.settings.timelimit + 1
-        try:
-            if config.args.timeout:
-                timeout = config.args.timeout
-        except AttributeError:
-            pass
+        self.settings.timelimit = config.args.timelimit or self.settings.timelimit
+
+        timeout = config.args.timeout or 1.5 * self.settings.timelimit + 1
         self.settings.timeout = int(timeout)
```
RagnarGrootKoerkamp, line 113: you can inline `timeout` here

bin/latex.py
```diff
@@ -90,7 +90,7 @@ def get_tl(problem):
 
     if 'print_timelimit' in contest_yaml():
         print_tl = contest_yaml()['print_timelimit']
-    elif hasattr(config.args, 'no_timelimit'):
+    elif config.args.no_timelimit is not None:
```
RagnarGrootKoerkamp, line 93: `is not None` isn't really needed right?  `false` and `None` should be equivalent (if it was possible to set the flag to False at all at some point in the future.)

bin/config.py
```diff
@@ -58,13 +58,83 @@
...
 args = argparse.Namespace()
-args.testcases = None
-args.cpp_flags = None
-args.verbose = None
-args.force_build = None
-args.error = None
+
+default_args = {
+    'jobs': os.cpu_count() // 2,
+    'time': 600,
```
RagnarGrootKoerkamp, line 65: Add a comment that this is for `fuzz` only (option names aren't very descriptive out of context, but that's ok)

bin/problem.py
```diff
@@ -155,11 +147,8 @@ def maybe_copy(x):
             return x.copy() if copy and isinstance(x, (list, dict)) else x
 
         samplesonly = only_sample
-        try:
-            if config.args.samples:
-                samplesonly = True
-        except AttributeError:
-            pass
+        if config.args.samples:
+            samplesonly = True
```
RagnarGrootKoerkamp, line 151: merge with the line above: `samplesonly = config.args.samples or only_sample` (ugh these variable names aren't good)

bin/program.py
```diff
@@ -419,6 +419,10 @@ def add_callback(problem, path, c):
             problem._program_callbacks[path] = []
         problem._program_callbacks[path].append(c)
 
+    @staticmethod
```
RagnarGrootKoerkamp, line 422: I think making this non-static is a bit nicer; `self.timeout()` reads better than `Program.timeout()`, and `timeout` feels more like a per program thing than a global thing anyway.

Actually, it's a bit weird that all programs have a timeout, but submissions actually use the problem timeout instead of the program timeout, but 'fixing' that (by propagating the timeout from problems to submissions) probably adds more code than it removes so not worth it I think.

bin/config.py
```diff
@@ -58,13 +58,41 @@
...
-args.testcases = None
-args.cpp_flags = None
-args.verbose = None
-args.force_build = None
-args.error = None
+
+
+def set_default_args():
+    # Set default argument values
+    for arg in [
```
RagnarGrootKoerkamp, line 80: If you can get the autoformatter to pack this list instead of putting it on many lines I think that'd be nice.

bin/config.py
```diff
@@ -70,29 +76,49 @@ def set_default_args():
         'check_deterministic',
         'clean',
         'clean_generated',
+        'cleanup_generated',
```
RagnarGrootKoerkamp, line 89: The list can be a global const, next to default_args.

bin/tools.py
```diff
@@ -447,7 +456,9 @@ def build_parser():
         action='store_true',
         help='Rerun all generators to make sure generators are deterministic.',
     )
-    genparser.add_argument('--timeout', '-t', type=int, help='Override the default timeout.')
+    genparser.add_argument(
+        '--timeout', '-t', type=int, help='Override the default timeout (Default: 30).'
```
RagnarGrootKoerkamp, line 460: `Override the default timeout. Default: 30` for consistency, here and elsewhere.

Independent of that, should we specify the unit somewhere? `default timeout in seconds` maybe? Although that's probably obvious and more verbose

bin/tools.py
```diff
@@ -536,7 +543,11 @@ def build_parser():
     runparser.add_argument(
         '--table', action='store_true', help='Print a submissions x testcases table for analysis.'
     )
-    runparser.add_argument('--timeout', type=int, help='Override the default timeout.')
+    runparser.add_argument(
+        '--timeout',
```
RagnarGrootKoerkamp, line 547: Hmm; `generate --timeout` and `run --timeout` have the same name, but do different things: the first is a global program timeout, while the second only affects submissions (actually, it probably also affects everything else but that isn't the intention). Maybe we can detangle them. In that case, the `generate --timeout` should probably rename to `generate --global-timeout` or so, but that's quite ugly so not sure about this

bin/tools.py
```diff
@@ -861,6 +876,36 @@ def run_parsed_arguments(args):
...
+            else Path(os.getenv('HOME')) / '.config'
+        )
+
+    for config_file in [
+        # Highest prio: contest directory
+        Path() / '.bapctools.yaml',
+        Path() / '..' / '.bapctools.yaml',
+        # Lowest prio: user config directory
+        home_config / 'bapctools' / 'config.yaml',
+    ]:
```
RagnarGrootKoerkamp, line 897: We could even add `tools_root/'config'/'bapctools.yaml'` in here for global defaults, but for now the small dictionary in `config.py` is simpler and no need to read even more files during startup

bin/tools.py
```diff
@@ -861,6 +876,36 @@ def run_parsed_arguments(args):
...
+        if not config_file.is_file():
+            continue
+        config_data = read_yaml(config_file)
+        assert isinstance(config_data, dict)
```
RagnarGrootKoerkamp, line 901: maybe allow `None` as well for empty config
```suggestion
        assert isinstance(config_data, [dict, None])
```

bin/tools.py
RagnarGrootKoerkamp: Anyway I think I'd like to add `contest.yaml` to this as well (but maybe a separate PR, since that likely adds a whole lot more changes), and `../../` variants to allow repo-root config from problem directories.

Really it would be cleaner to parse these after `cd`ing to the contest dir, but that probably introduces other problems

readme.md
```diff
@@ -157,6 +157,21 @@ Use this command to compile the `problem.pdf` from the `problem_statement/proble
 
 This can also be used to create the contest pdf by running it from the contest directory.
 
+## Personal configuration file
+
+For some command-line flags, it is convenient if they are always set to the same value, which differs per user
+(e.g., `--username` or `--password` for commands that access a CCS like DOMjudge,
+or `--scoreboard_repo` for the `bt solvestats` command).
```
RagnarGrootKoerkamp, line 164:
```suggestion
or `--scoreboard-repo` for the `bt solvestats` command).
```

readme.md
```diff
@@ -157,6 +157,21 @@ Use this command to compile the `problem.pdf` from the `problem_statement/proble
...
+For this, you can create a configuration YAML file containing key-value pairs
+in one of the following locations:
```
RagnarGrootKoerkamp, line 166:
```suggestion
in one of the following locations, from low to high priority:
```

bin/tools.py
```diff
@@ -896,6 +876,36 @@ def run_parsed_arguments(args):
...
+def read_personal_config():
+    args = {}
+
+    if is_windows():
+        home_config = Path(os.getenv('AppData'))
+    else:
+        home_config = (
+            Path(os.getenv('XDG_CONFIG_HOME'))
+            if os.getenv('XDG_CONFIG_HOME')
+            else Path(os.getenv('HOME')) / '.config'
```
RagnarGrootKoerkamp, line 888: You can probably use `or` here as well.

Do we want to fall back to `~/.bapctools.yaml` as well?

RagnarGrootKoerkamp (changes requested):
Sweeeet! :heart: 

Mostly small comments and remarks. Not all of them are actionable and some are just random remarks; feel free to either comment or ignore ;)

bin/tools.py
RagnarGrootKoerkamp: or add `or {}` to the line above

bin/problem.py
mpsijm: Thanks, I also removed the `int(...)`, because this should all be `int`s already anyway :slightly_smiling_face: (57be5a8)

bin/latex.py
mpsijm: You're right, `None` defaults to `False` anyway, so I inlined it :slightly_smiling_face: (57be5a8)

bin/config.py
mpsijm: :+1: (57be5a8)

bin/problem.py
mpsijm: Hehe, I'll just leave the variable names as-is for now, I don't really have a really good suggestion anyway. :joy: But I merged the two lines :slightly_smiling_face: (57be5a8)

bin/program.py
mpsijm: Given that `timeout` is, as I agree, "more like a per program thing than a global thing", I've inlined the function into `Generator.run` and `Visualizer.run`. In the parameter descriptions of the argument parser, the default value is duplicated as well, anyway. :stuck_out_tongue: (57be5a8)

Related to your other comment about rethinking Program timeouts vs. Submission timeouts, we can probably better open a new issue for that for now :stuck_out_tongue:

bin/config.py
mpsijm: Done, see 87b84ca

bin/config.py
mpsijm: Done, see 87b84ca

bin/tools.py
mpsijm: Check, I've checked for consistency for "Default" clauses in the command descriptions :slightly_smiling_face: (57be5a8)

I don't think the unit is necessary here, "seconds" should™ be obvious :stuck_out_tongue:

bin/tools.py
mpsijm: Yeah, the naming is a bit awkward... we could also rename the `run --timeout` argument to something like `--abort-time`, but that also sounds awkward. I'm also trying to relate this to Kattis' `safety_margin`, but that's a factor, not a hard number (although we could rewrite the running of submissions to use this safety margin instead of a fixed timeout value).

Related to your other comment about rethinking Program timeouts vs. Submission timeouts, we can probably better open a new issue for that for now :stuck_out_tongue:

bin/tools.py
mpsijm: As we discussed on Slack, I think that `contest.yaml` should stay separate from the command-line args, to keep it clearer where certain values are coming from.

I've tried separating the `cd`ing to the contest dir before parsing the personal config file in TODO. This commit is on a separate branch, so it doesn't interfere with this PR, so we can move the discussion elsewhere. Splitting off a `change_directory()` function from the `get_problems()` function already makes things slightly cleaner, but:
- Moving `change_directory()` outside `run_parsed_args` means we'll have to duplicate this between `main()` and `test(args)`
- This will also require changes in the behaviour of `new_{contest,problem}`, as the `--contest` and `--problem` flags are not valid for these commands

bin/tools.py
mpsijm: I went for the second option :slightly_smiling_face: (57be5a8)

readme.md
mpsijm: I see you made the hyphens/underscores consistent, nice :smile: (57be5a8)

readme.md
mpsijm: :+1: (57be5a8)

bin/tools.py
mpsijm: Using `or` is difficult here, because if `os.getenv(...) = None`, then `Path(None)` will throw an exception.

I don't think we should use multiple config locations for now, I don't like polluting my home dir with random config files anyway :stuck_out_tongue:

bin/problem.py
mpsijm: Woops, this turned out to be a `float`, obviously. Time to stop programming for tonight :innocent: (force-pushed to cc6107c, as if nothing ever happened :innocent:)

RagnarGrootKoerkamp:
I think this is good to go now :+1: Made an issue for the `--timeout` thing, and looking at the `change_dir` commit which can be a separate PR after this one.

RagnarGrootKoerkamp (approved):

PR #166, opened 2021-11-12
[tools] Separate changing to contest directory and getting problems list

mpsijm:
Copying part of the discussion from https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/163#discussion_r748593896:

@RagnarGrootKoerkamp:
> Really it would be cleaner to parse these after cding to the contest dir, but that probably introduces other problems

@mpsijm:
> I've tried separating the `cd`ing to the contest dir before parsing the personal config file in f5ac03601745ae8935d89209356b8b5ba8862cd2. [...] Splitting off a `change_directory()` function from the `get_problems()` function already makes things slightly cleaner, but:
> - Moving `change_directory()` outside `run_parsed_args` means we'll have to duplicate this between `main()` and `test(args)`
> - This will also require changes in the behaviour of `new_{contest,problem}`, as the `--contest` and `--problem` flags are not valid for these commands

### DIFF
--- a/bin/tools.py
+++ b/bin/tools.py
@@ -67,31 +67,37 @@ if not os.getenv('GITLAB_CI', False):
 # Make sure f-strings are supported.
 f'f-strings are not supported by your python version. You need at least python 3.6.'
 
-# Get the list of relevant problems.
-# Either use the problems.yaml, or check the existence of problem.yaml and sort
-# by shortname.
-def get_problems():
-    def is_problem_directory(path):
-        return (path / 'problem.yaml').is_file()
 
-    contest = None
-    problem = None
-    level = None
+# A path is a problem directory if it contains a `problem.yaml` file.
+def is_problem_directory(path):
+    return (path / 'problem.yaml').is_file()
+
+
+# Changes the working directory to the root of the contest.
+# Returns the "level" of the current command (either 'problem' or 'problemset')
+# and, if `level == 'problem'`, the directory of the problem.
+def change_directory():
+    problem_dir = None
+    level = 'problemset'
     if config.args.contest:
-        contest = config.args.contest.resolve()
-        os.chdir(contest)
-        level = 'problemset'
+        os.chdir(config.args.contest.resolve())
     if config.args.problem:
-        problem = config.args.problem.resolve()
+        problem_dir = config.args.problem.resolve()
         level = 'problem'
-        os.chdir(problem.parent)
+        os.chdir(problem_dir.parent)
     elif is_problem_directory(Path('.')):
-        problem = Path().cwd()
+        problem_dir = Path().cwd()
         level = 'problem'
         os.chdir('..')
-    else:
-        level = 'problemset'
 
+    config.level = level
+    return problem_dir
+
+
+# Get the list of relevant problems.
+# Either use the problems.yaml,
+# or check the existence of problem.yaml and sort by shortname.
+def get_problems(problem_dir):
     # We create one tmpdir per contest.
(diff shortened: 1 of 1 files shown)


bin/tools.py
```diff
@@ -67,31 +67,37 @@
...
+    return problem_dir
+
+
+# Get the list of relevant problems.
+# Either use the problems.yaml,
+# or check the existence of problem.yaml and sort by shortname.
+def get_problems(problem_dir):
     # We create one tmpdir per contest.
     h = hashlib.sha256(bytes(Path().cwd())).hexdigest()[-6:]
     tmpdir = Path(tempfile.gettempdir()) / ('bapctools_' + h)
```
RagnarGrootKoerkamp, line 103: Maybe this should also change to `config.tmpdir`, like `config.level`. Not sure though.

mpsijm: That's possible, but then the same would hold for `contest_name`, `problems`, and maybe even `problem` :stuck_out_tongue: Where would you draw the line between moving parameters to `config` and just passing them as parameters to the functions? In the other extreme, the global `config.level` (and maybe even `config.args`) should not exist, but instead be passed as a local parameter :stuck_out_tongue:

RagnarGrootKoerkamp: Yeah fair enough. Also, I could see tmpdir change value within a single invocation, so let's just keep it as is.

RagnarGrootKoerkamp:
> * Moving `change_directory()` outside `run_parsed_args` means we'll have to duplicate this between `main()` and `test(args)`
That's find - duplicating some code for tests is OK with me.

> * This will also require changes in the behavior of `new_{contest,problem}`, as the `--contest` and `--problem` flags are not valid for these commands

Right. Do you want to spend some more time on this for this PR? (I assume that's why it's draft for now.)

I'm not exactly sure myself yet how we should handle this. One possibility:

1. parse args without reading any config files (only to find the subcommand)
2. if new_{contest,problem}, handle that and exit
3. if other subcommand, cd to contest directory
4. parse args again, now with reading the required context

mpsijm:
> duplicating some code for tests is OK with me.

All right :slightly_smiling_face: 

> (I assume that's why it's draft for now.)

Correct :yum:

> I'm not exactly sure myself yet how we should handle this. One possibility: [...]

Sounds good, but I don't like the fact that the arguments need to be parsed twice. I was thinking of something like the following (but I'll need to try this to see how it works out):

1. In `main()`:
    1. Read command line args (without passing the personal config as default values)
    2. Set `initial_cwd = os.getcwd()`
    3. `cd` to to contest directory (this reads `--contest` and `--problem` _only_ from the command line, but having them in a personal config file wouldn't make sense anyway :stuck_out_tongue:)
    4. Read personal config
    5. Call `run_parsed_arguments` with combined `args` object
2. In `run_parsed_arguments(args)`:
    1. `if action in ['new_contest', 'new_problem']:` check if `--contest` and/or `--problem` are set, and throw an error if so (only `new_problem --contest` would be valid). Then, `os.chdir(initial_pwd)` and continue with the action
    2. `else:` continue processing the `args` as usual

Maybe 2.i should be checked before 1.iv, but that requires a bit more rewriting. I'll experiment with it for a bit :slightly_smiling_face:

RagnarGrootKoerkamp:
@mpsijm what's the status here? I kinda forgot what this was about

mpsijm:
If I remember correctly, this is mostly to make the parsing of the personal config file more consistent. It's mostly a code cleanup that doesn't really have high priority, but I would still like to finish it if I do have the time :stuck_out_tongue:

PR #184, opened 2021-12-16
[build c++] Change default to -std=gnu++17

mzuenni:


RagnarGrootKoerkamp:
Let's make it gnu++17 instead. Validators often need to run on machines with old versions, and backporting from 20 to 17 on contest day is annoying

PR #185, opened 2021-12-23
made randomnes compiler/STL independent

mzuenni:


headers/validation.h
```diff
@@ -108,6 +109,52 @@ auto operator|(T1 /*unused*/, T2 /*unused*/) {
 
 enum Separator { Space, Newline };
 
+//this contains some specific code which would be available in c++20
```
RagnarGrootKoerkamp, line 112: I read this as 'this depends on c++20' at first. Maybe make it slightly clearer that instead this would be replaced by c++20.

headers/validation.h
```diff
@@ -108,6 +109,52 @@ auto operator|(T1 /*unused*/, T2 /*unused*/) {
...
+namespace C20 {
```
RagnarGrootKoerkamp, line 113: Probably nicer to call it something like `cpp20`, (or maybe even `future` as in python).


We could even consider putting this behind an `#ifdef [some cpp20 define]`, but macros are ugly so simpler to keep as is.

headers/validation.h
```diff
@@ -108,6 +109,52 @@ auto operator|(T1 /*unused*/, T2 /*unused*/) {
...
+
+constexpr int countl_zero(unsigned long long x) {
```
RagnarGrootKoerkamp, line 115: do you have a source for this (please add as comment) or handwritten?

headers/validation.h
```diff
@@ -108,6 +109,52 @@ auto operator|(T1 /*unused*/, T2 /*unused*/) {
...
+			res -= i;
+			x >>= i;
+		}
+	}
+	if (x > 0) res--;
+	return res;
+}
+
+int popcount(unsigned long long x) {
+	return static_cast<int>(std::bitset<64>(x).count());
```
RagnarGrootKoerkamp, line 128: Looks horrible, but I suppose `bitset<64>` is actually calling the platform dependant popcnt function so should be good :+1:

headers/validation.h
```diff
@@ -108,6 +109,52 @@ auto operator|(T1 /*unused*/, T2 /*unused*/) {
...
+	// a long double can represent more than 2^64 values in the range [0, 1)...
+	// another problem is that real64() < 1.0/3.0 is technically biased.
+	long double res = bits64(rng) / 0x1.0p64l;
+	res += bits64(rng) / 0x1.0p128l;
+	assert(res < 1.0l);
+	return res;
+}
+
+bool bit(std::mt19937_64& rng) {
+	return C20::popcount(bits64(rng));
```
RagnarGrootKoerkamp, line 153:
```suggestion
	return C20::popcount(bits64(rng)) & 1;
```

headers/validation.h
```diff
@@ -166,11 +213,25 @@ struct UniformGenerator {
 	template <typename T>
 	T operator()(T low, T high, std::mt19937_64& rng) const {
 		static_assert(is_number_v<T>);
-		if constexpr(std::is_same_v<T, long long>)
-			return std::uniform_int_distribution<T>(low, high)(rng);
```
RagnarGrootKoerkamp, line 170: It would be cool if we could somehow ban the distributions from this file. Since `<random>` is also needed for RNGs, that doesn't seem feasible though.

headers/validation.h
```diff
@@ -166,11 +213,25 @@ struct UniformGenerator {
...
+			// but definetly unbiased
+			unsigned long long ul = static_cast<unsigned long long>(low);
+			unsigned long long uh = static_cast<unsigned long long>(high);
+			int shitfs = C20::countl_zero(uh - ul + 1ull) % std::numeric_limits<unsigned long long>::digits;
+			unsigned long long res;
+			do {
+				res = Random::bits64(rng) >> shitfs;
+			} while (res > uh - ul);
+			return static_cast<long long>(res + ul);
+		}	else {
```
RagnarGrootKoerkamp, line 229: Can you make this `else if costexpr` for `long double`, and `[static_]assert(false)` in the remaining case

headers/validation.h
```diff
@@ -222,14 +283,25 @@ struct NormalDistributionGenerator {
...
 	T operator()(T low, T high, std::mt19937_64& rng) const {
+		assert(low < high);
 		T v;
-		do {
-			v = std::normal_distribution<T>(mean_, stddev_)(rng);
-		} while(v < low or v > high);
+		while (true) {
+			T u1 = Random::real64(rng);
+			T u2 = Random::real64(rng);
+			// Box-Muller-Methode
```
RagnarGrootKoerkamp, line 297: Can you add a link/reference?

headers/validation.h
```diff
@@ -222,14 +283,25 @@ struct NormalDistributionGenerator {
...
+			v = std::sqrt(-2.0l * std::log(u1)) * std::cos(2.0l * C20::PI * u2);
+			v = std::sqrt(stddev_)*v+mean_;
+			if (v >= low && v < high) return v;
+			v = std::sqrt(-2.0l * std::log(u1)) * std::sin(2.0l * C20::PI * u2);
```
RagnarGrootKoerkamp, line 301: Hmm, are `cos` and `sin` independent enough here? If the interval is small, the fact that `cos(u2)` is outside the interval may bias the cases where `sin(u2)` is inside

headers/validation.h
```diff
@@ -108,6 +109,52 @@ auto operator|(T1 /*unused*/, T2 /*unused*/) {
...
+
+namespace Random {
+
+unsigned long long bits64(std::mt19937_64& rng) {
+	static_assert(std::mt19937_64::max() == 0xFFFF'FFFF'FFFF'FFFFull);
+	static_assert(std::mt19937_64::min() == 0ull);
+	return rng();
+}
+
+long double real64(std::mt19937_64& rng) {
```
RagnarGrootKoerkamp, line 143: Add a comment that this returns in `[0, 1)` (as opposed to e.g. any positive float)

headers/validation.h
```diff
@@ -260,15 +335,19 @@ struct GeometricDistributionGenerator {
...
 	// NOTE: Currently this retries instead of clamping to the interval.
 	T operator()(T low, T high, std::mt19937_64& rng) const {
+		assert(low <= high);
 		T v;
-		do {
-			v = low + std::geometric_distribution<T>(p_)(rng);
-		} while(v < low or v > high);
-		return v;
+		while (true) {
+			v = low + std::floor(std::log(Random::real64(rng)) / std::log1p(-p_));
```
RagnarGrootKoerkamp, line 348: This could also use a source ;)
(I'm too lazy to think about formulas at the moment, and some indication for why this is correct would be nice.)

headers/validation.h
```diff
@@ -280,15 +359,23 @@ struct BinomialDistributionGenerator {
...
 
 	// NOTE: Currently this retries instead of clamping to the interval.
 	T operator()(T low, T high, std::mt19937_64& rng) const {
-		T v;
-		do {
-			v = std::binomial_distribution<T>(n_, p_)(rng);
-		} while(v < low or v > high);
-		return v;
+		assert(low <= high);
+		// this will be slow for large n
```
RagnarGrootKoerkamp, line 370: Hmm; that sounds like a regression. I don't think this is used much (at all) anyway, so it's fine. But maybe print a warning in the constructor when `n` is larger than 1000 or so

RagnarGrootKoerkamp (changes requested):
This looks very nice! I've some small comments.

This file is getting quite large now -- at some point some cleanup may be needed but that's for another day.

headers/validation.h
mzuenni: this is handwritten

headers/validation.h
mzuenni: that's what i hope for :D

headers/validation.h
mzuenni: maybe with some macro magic... but i don't know if that's possible... alternatively BAPCtools could check files for this as with `#include <bits/stdc++.h>`?

headers/validation.h
mzuenni: https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform

headers/validation.h
mzuenni: Wikipedia says this is fine

headers/validation.h
mzuenni: https://en.wikipedia.org/wiki/Geometric_distribution
look for  "The exponential distribution is the continuous analogue of the geometric distribution[...]"

headers/validation.h
RagnarGrootKoerkamp: > alternatively BAPCtools could check files for this as with #include <bits/stdc++.h>?

yeah, I already check for `list(set(..))` in python as well since sets don't have very stable ordering. Could add the distrubitions to that list. But lets do that in a separate PR.

headers/validation.h
```diff
@@ -362,6 +366,7 @@ struct BinomialDistributionGenerator {
 	explicit BinomialDistributionGenerator(long long n, double p) : n_(n), p_(p) {
 		assert(p_ >= 0);
 		assert(p_ <= 1);
+		std::cerr << "Warning: Large n (" << n_ << ") is slow for BinomialDistributionGenerator!" << std::endl
```
RagnarGrootKoerkamp, line 369: Add an `if(n > 1000)` around this ;)

RagnarGrootKoerkamp (approved):

headers/validation.h
RagnarGrootKoerkamp: sadly the solution [is ugly](https://stackoverflow.com/questions/38304847/constexpr-if-and-static-assert), so let's not

RagnarGrootKoerkamp:
Merging. The remaining error is unrelated.

PR #188, opened 2022-01-12
Keep symlinks inside copied directory

### LINKED ISSUE #186: Creating new problem doesn't copy generators/generator symlink
When I create a new problem, `generators/generator` is a hardcopy of `input_validators/input_validator` but should be a symlink to it, like it is a symlink in the /skel/problem/ directory.

mzuenni:
Should fix #186

### DIFF
--- a/bin/util.py
+++ b/bin/util.py
@@ -586,7 +586,7 @@ def copytree_and_substitute(src, dst, variables, exist_ok=True, *, preserve_syml
                 dstFile = dst / name
 
                 copytree_and_substitute(
-                    srcFile, dstFile, variables, exist_ok, preserve_symlinks=preserve_symlinks
+                    srcFile, dstFile, variables, exist_ok, preserve_symlinks=preserve_symlinks, base=base
                 )
             except OSError as why:
                 errors.append((srcFile, dstFile, str(why)))
--- a/bin/util.py
+++ b/bin/util.py
@@ -568,8 +568,13 @@ def substitute_dir_variables(dirname, variables):
 
 # copies a directory recursively and substitutes {%key%} by their value in text files
 # reference: https://docs.python.org/3/library/shutil.html#copytree-example
-def copytree_and_substitute(src, dst, variables, exist_ok=True, *, preserve_symlinks=True):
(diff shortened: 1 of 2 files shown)


bin/util.py
```diff
@@ -568,9 +568,14 @@ def substitute_dir_variables(dirname, variables):
 
 # copies a directory recursively and substitutes {%key%} by their value in text files
 # reference: https://docs.python.org/3/library/shutil.html#copytree-example
-def copytree_and_substitute(src, dst, variables, exist_ok=True, *, preserve_symlinks=True):
+def copytree_and_substitute(src, dst, variables, exist_ok=True, *, preserve_symlinks=True, base=None):
```
RagnarGrootKoerkamp, line 571: You don't modify any call site, so can you not just remove `base` and replace it by `src` everywhere?

bin/util.py
```diff
@@ -568,9 +568,14 @@ def substitute_dir_variables(dirname, variables):
...
+    if base is None:
+        base = src
+
     if preserve_symlinks and os.path.islink(src):
         shutil.copy(src, dst, follow_symlinks=False)
+    elif os.path.islink(src) and src.resolve().is_relative_to(base):
```
RagnarGrootKoerkamp, line 577: currently this is always true.
I don't actually understand why it was broken before anyway: if `preserve_symlinks` is true, then the if-statement above should catch it.

Or is it that for the case in the linked issue, preserve_symlinks should be ignored and symlinks should always be copied as symlinks, even when `preserve_symlinks` is `false`?

bin/util.py
mzuenni: fixed that. In the recursive call `base` should be the top level `src`

bin/util.py
mzuenni: Yes, it's for the related issue. Symlinks from one file in `skel/problem` to another file in `skel/problem` should always remain symlinks when creating a new problem?

RagnarGrootKoerkamp (approved):

bin/util.py
RagnarGrootKoerkamp: yes completely agree

RagnarGrootKoerkamp (approved):

Issue #152, opened 2021-10-30
stats sample count counts `.interaction` twice


RagnarGrootKoerkamp:

mzuenni:
This can be closed now?

PR #189, opened 2022-01-12
Fix counting .interaction and .in as separate test cases

mzuenni:
see issue #152


mzuenni:
Why does the stat command even consider any files besides `*.in`? Is it possible to have a test case without a corresponding `.in` file?

RagnarGrootKoerkamp:
Well...., it's complicated. I was messing around a lot with this last year. In particular, for interactive problems the samples sometimes have _only_ a `.interaction` file.

Anyway, LGTM. Will merge after #185.

RagnarGrootKoerkamp (approved):

PR #190, opened 2022-01-14
Added some new arguments to the generate command

mzuenni:


bin/generate.py
```diff
@@ -668,8 +667,7 @@ def up_to_date():
                         return
 
         # Generate visualization
-        # TODO: Disable this with a flag.
-        if t.config.visualizer:
+        if t.config.visualizer and not config.args.skip_visualizer:
```
RagnarGrootKoerkamp, line 670: I prefer this as
```suggestion
        if not config.args.skip_visualizer and t.config.visualizer:
```

bin/generate.py
```diff
@@ -636,8 +636,7 @@ def up_to_date():
                 return
 
         # Generate .ans and .interaction if needed.
-        # TODO: Disable this with a flag.
-        if not (testcase.bad_input or testcase.bad_output):
+        if not (testcase.bad_input or testcase.bad_output or config.args.skip_solution):
```
RagnarGrootKoerkamp, line 639: for symmetry with the other comment this should be
```suggestion
        if not config.args.skip_solution and not (testcase.bad_input or testcase.bad_output):
```
but it's quite ugly so maybe just `if not (config.args.skip_solution or ...)` instead

RagnarGrootKoerkamp (approved):
Thanks for fixing random TODOs :D

PR #191, opened 2022-01-14
Nicer prompts based on questionary

mzuenni:


RagnarGrootKoerkamp:
This is looking quite fancy :D Let me know when it's done for reviewing

mzuenni:
We could add additional inputs in `new_problem`  for `source_url`, `license` and `rights_owner` if they are not in the contest yaml. 
Right now they simple get initialised with an empty string and I am not sure if changing this would break anything?
Besides that it should be finished :)

RagnarGrootKoerkamp:
Oh yes, please go ahead and ask for those if they aren't set for the
contest. Or better yet, always ask for them, but give the value in the
contest as a default so pressing enter is sufficient.

mzuenni:
ok, now all values should get asked. The only remaining idea would be to introduce the `float` validator option.

And maybe we want to change for which fields it is allowed to input an empty string? I tried to emulate the original behaviour i.e. if the default value is the empty string it is ok and else it is not. Right now the fields `subtitle` `source` and `source_url` can be empty.

RagnarGrootKoerkamp:
> The only remaining idea would be to introduce the `float` validator option.

What exactly do you mean? https://icpc.io/problem-package-format/spec/problem_package_format#problem-metadata doesn't mention `validation: float`, if that's what you mean?
Ah, or do you mean asking for [these](https://icpc.io/problem-package-format/spec/problem_package_format#default-output-validator-specification)? We could do some simple question like `float output?` and if so set some default parameters. Asking for the tolerance right away doesn't make much sense IMO.

> And maybe we want to change for which fields it is allowed to input an empty string? I tried to emulate the original behaviour i.e. if the default value is the empty string it is ok and else it is not. Right now the fields `subtitle` `source` and `source_url` can be empty.

SGTM

mzuenni:
Yes, I mean when asking for validation we would give the option `float`, then ask for the eps and then write `default` and the `float_tolerance` in the yaml.

RagnarGrootKoerkamp:
> Yes, I mean when asking for validation we would give the option `float`, then ask for the eps and then write `default` and the `float_tolerance` in the yaml.

If you feel like it I'm happy to ask for `float?` indeed, but less exited about the tolerance. Just printing `using default tolerance of 10^-6` or so should be sufficient.

RagnarGrootKoerkamp:
Hrmm, looks like questionary needs `pip` to install. All other dependencies are available as ubuntu/arch packages directly. That's a bit unfortune for a non-critical dependency.

Maybe we could do
```
has_questionary = false
try:
   import questionary
   has_questionary = true
except ImportError:
   pass
```

and fallback to a simpler prompt similar to what we have now in case it's not available

mzuenni:
I think the changes are complete now :)

PR #192, opened 2022-01-18
Warn for more bad practice

mzuenni:
We could warn for even more stuff defined in `random` but let's at least warn for those methods implemented in `validation.h`.

### DIFF
--- a/bin/program.py
+++ b/bin/program.py
@@ -3,6 +3,7 @@ import shutil
 import stat
 import subprocess
 import threading
+import validate
 
 from util import *
 
@@ -283,7 +284,7 @@ class Program:
                     pass
 
         # Warn for known bad (non-deterministic) patterns in generators
-        if isinstance(self, Generator):
+        if isinstance(self, Generator) or isinstance(self, validate.Validator):
             if self.language == 'cpp':
                 for f in self.source_files:
                     try:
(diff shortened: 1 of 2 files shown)


RagnarGrootKoerkamp:
Nice!

Do you think we really need the `typeid()`? I don't think that's likely to be used for generators right?

Should we check this for all programs though? I think we only want to forbid these things for generators and (interactive) validators. Team submissions are fine

mzuenni:
well it already warned for `rand()` I just added a few more... I am not sure for which programs this is actually checked...

Probably no one uses typeid but it's compiler dependent and it won't harm I guess?

bin/program.py
```diff
@@ -287,19 +287,35 @@ def _checks(self):
             if self.language == 'cpp':
```
RagnarGrootKoerkamp, line 287: can you add `or (self.problem.interactive and isinstance(self, Validator)` and test that?  (I'm not exactly sure whether that `self.problem.interactive` is exactly right.)

mzuenni: why not just warn for all validators? I don't see why a normal output validator should be allowed to be non-deterministic but in an interactive problem, we don't want to allow it?

RagnarGrootKoerkamp: ok let's do all and revisit when people complain ;)

PR #193, opened 2022-01-21
Generic validation

### LINKED ISSUE #187: Input Validator/Testdata checks
Problemtools does some sanity checks on the input validator. A validator should (normally) reject the following:

- An empty file
- Binary file/non-printable characters
- Random text file
- A valid test case with duplicated spaces
- A valid test case with duplicated newlines
- A leading zero before an integer (could be ok)
- Garbage added behind a valid test case

Additionally, we could check the test cases:

- All files should have the same line endings
- All whitespace should be space/newline

This might be a good feature to add.

mzuenni:
This adds some generic checks to the testcases and warns if a testcase:

- is empty (contains no character besides whitespaces)
- does not end with a newline (I think files should always end with '\n')
- contains characters besides '\n' and printable ASCII characters (warns on '\r', Unicode or other control characters)
- starts with whitespace or contains two consecutive whitespaces (E.g a space at the start/end of a line or duplicated spaces. However, this could fail if the test case is supposed to contain '\n\n' but this rarely happens)
- is too big (>20Mb for input or output, but I am not sure what a `.ans` file contains for interactive problem...)

**This does not automatically generate bad testcases from valid ones! This might be an improvement for later...**

However, it also fixes some small bugs.

Fix #187


doc/commands.md
```diff
@@ -506,11 +509,11 @@ This is a convenience command (mostly for use in CI) that runs the following sub
 - Run all submissions
 - Clean up generated testcases when done, similar to `bt generate --clean-generated` when `--cleanup-generated` is passed.
 
-This supports the `--cp` and `--no-timelimit` flags which are described under the `pdf` subcommand.
+This supports the `--cp` and `--no-timelimit` flags which are described under the `pdf` subcommand and the `--skip-testcase-sanity-checks` flag from `validate`.
 
 ## `sort`
 
-Prints a list of all problems in the current contest (or single problem), together with their letter/ID:
+Prints a list of all proble You might want to set this in `.bapctools.yaml`.ms in the current contest (or single problem), together with their letter/ID:
```
RagnarGrootKoerkamp, line 516: whoops; revert this one?

bin/run.py
```diff
@@ -234,7 +275,7 @@ def run(self, *, interaction=None, submission_args=None):
             if (
                 not config.args.error
                 and self.out_path.is_file()
-                and self.out_path.stat().st_size > 1000000000
+                and self.out_path.stat().st_size > 1000_000_000
```
RagnarGrootKoerkamp, line 278: add one more `_`

bin/run.py
```diff
@@ -169,6 +169,47 @@ def validate_format(
...
+                            bar.warn('Testcase is empty but was accepted!')
+                        elif not re.search(r"\n$", input):
+                            bar.warn('Testcase does not end with a newline but was accepted!')
+                        elif len(bytes) > 20_000_000_000:
+                            bar.warn('Testcase is larger than 20Mb!')
+
+            if validator_type == 'output_format' and self.ans_path.exists():
+                with self.ans_path.open() as in_file:
+                    def invalid(byte):
+                        if byte == ord('\t'): return False;
```
RagnarGrootKoerkamp, line 199: Can you add a comment why this is different that the input invalid bytes?
I.e.: input allows only spaces and newlines, while output is a bit more lenient.

Also, `def ...` inside a loop is relatively slow in python. (Not that this is critical code, but anyway.) Better to define a separate function `invalida(byte, input/output)`

bin/run.py
```diff
@@ -109,7 +109,7 @@ def validate_format(
             flags = self.testdata_yaml_validator_flags(validator_type, validator)
             if flags is False:
                 continue
-
+
```
RagnarGrootKoerkamp, line 112: revert

RagnarGrootKoerkamp (changes requested):

RagnarGrootKoerkamp (approved):

PR #194, opened 2022-03-13
[util] Consistent writing of YAML files

mpsijm:
Now, all YAML files are read using `util.read_yaml` and written using `util.write_yaml`.
These function abstract over the indentation settings that used to be set all over the place.
`util.write_yaml` uses `ruamel.yaml`, so before using that function, you should check `util.has_ryaml`.

### DIFF
--- a/bin/generate.py
+++ b/bin/generate.py
@@ -21,12 +21,8 @@ from problem import Problem
 def check_type(name, obj, types, path=None):
     if not isinstance(types, list):
         types = [types]
-    if obj is None:
-        if obj in types:
-            return
-    else:
-        if any(obj is None if t is None else isinstance(obj, t) for t in types):
-            return
+    if any(isinstance(obj, t) for t in types):
+        return
     named_types = " or ".join(str(t) if t is None else t.__name__ for t in types)
     if path:
         fatal(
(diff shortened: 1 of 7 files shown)


bin/util.py
```diff
@@ -466,6 +474,24 @@ def read_yaml_settings(path):
...
 
+# Writing a yaml file only works when ruamel.yaml is loaded. Check if `has_ryaml` is True before using.
+def write_yaml(data, path):
+    ryaml.dump(
+        data,
+        path,
+        # Remove spaces at the start of each (non-commented) line, caused by the indent configuration.
+        # This is only needed when the YAML data is a list of items, like in the problems.yaml file.
+        # See also: https://stackoverflow.com/a/58773229
+        transform=(
```
RagnarGrootKoerkamp, line 485: I suspect this transform is broken when the data is not actually a list. Can you test, or just do `if instanceof(data, list): dump with transform else: dump without transform`

RagnarGrootKoerkamp (changes requested):

bin/util.py
mpsijm: I tested this with `bapctools generate --add-manual`, which rewrites the `generators/generators.yaml`. This yaml file is a `dict`, not a `list`, so `transform=None` in that case. :slightly_smiling_face: 

The code that you propose does exactly the same as the inline conditional that it currently is. Although I understand that the formatting of the code does not make that very clear, so I added some parentheses, hopefully that helps :smile:

RagnarGrootKoerkamp: Oh I completely missed the if statement inside there. Should be good now

bin/generate.py
```diff
@@ -273,19 +269,19 @@ class Config:
     # Used at each directory or testcase level.
 
     def parse_solution(p, x, path):
-        check_type('Solution', x, [None, str], path)
+        check_type('Solution', x, [type(None), str], path)
```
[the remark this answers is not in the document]
mpsijm: Unforunately not, no :upside_down_face: See the commit message of ce0bcd598bdf38dda8f069b21e8ad938959f40c8:

> Note that you cannot directly access `NoneType` in Python 3, so instead we need to write `type(None)`.
> This makes the `types` list more consistent as well, because all elements are now actually types (whereas `None` is a value, not a type).

I guess we could write `NoneType = type(None)` at the top of the file but that's also not too pretty :joy:

PR #196, opened 2022-03-15
[util] Add threading lock to write_yaml

mpsijm:
I started getting flaky errors when running `bt generate` (which by default runs in parallel), like this one:

<details>
<summary>Screenshot</summary>

image

</details>

Adding a threading lock seems to fix that :smile:

### DIFF
--- a/bin/util.py
+++ b/bin/util.py
@@ -474,8 +474,13 @@ def read_yaml_settings(path):
     return settings
 
 
+# Only allow one thread to write at the same time. Else, e.g., generating test cases in parallel goes wrong.
+write_yaml_lock = threading.Lock()
+
+
 # Writing a yaml file only works when ruamel.yaml is loaded. Check if `has_ryaml` is True before using.
 def write_yaml(data, path):
+    write_yaml_lock.acquire()
     ryaml.dump(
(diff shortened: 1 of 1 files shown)


RagnarGrootKoerkamp:
Hmm right yes. I suppose those cache metadata files don't really need ruamel.yaml, but they're not a bottleneck anyway so it's fine.

RagnarGrootKoerkamp:
Oh actually, could you use a `with ...` or so (if something like that is directly available), so that the lock is released even if the dump call throws?

mpsijm:
> I suppose those cache metadata files don't really need ruamel.yaml, ...

Yeah, not necessarily indeed, but I wanted to make things consistent :slightly_smiling_face: 

> ..., but they're not a bottleneck anyway so it's fine.

Agreed :smile:

> Oh actually, could you use a with ... or so (if something like that is directly available), so that the lock is released even if the dump call throws?

Ah, good point, I didn't know that that's a thing :smile: https://docs.python.org/3/library/threading.html#using-locks-conditions-and-semaphores-in-the-with-statement
I opened #197 for that :slightly_smiling_face:

PR #200, opened 2022-04-03
fix uniform random for closed interval...

mzuenni:
It seems like I made a mistake in the random integer generation... (the original code was for half-open intervals and I did not change it correctly for closed intervals when I introduced it in validation.h) This patch should fix the issue.

### DIFF
--- a/headers/validation.h
+++ b/headers/validation.h
@@ -224,7 +224,7 @@ struct UniformGenerator {
 			// but definetly unbiased
 			unsigned long long ul = static_cast<unsigned long long>(low);
 			unsigned long long uh = static_cast<unsigned long long>(high);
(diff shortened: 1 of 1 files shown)


RagnarGrootKoerkamp:
Can you explain this modulus by the number of bits in ull (in a code comment)? It's seems redundant to me. Maybe also comment why you're using the high bits instead of the low bits, since low bits would be the default IMO.

mzuenni:
Ah, I see, the modulo was also just needed for the half-open case. I will remove it.
I am using the high bits since it is easier to obtain... (and typically the high bits have better quality for many PRNGs but that shouldn't be too much of an issue for mersenne twister)

PR #203, opened 2022-04-10
Improved stdout.py

mzuenni:
This change allows escaped newlines (and other backslash escaped chars) in the argument for `stdout.py`


RagnarGrootKoerkamp:
Can you add that as a comment?

mzuenni:
added a comment

RagnarGrootKoerkamp:
Sidenote: yaml also supports multiline string arguments, but I'm not sure that works well in combination with shell invocations

PR #206, opened 2022-04-29
removed old try catch block

mzuenni:
As far as I can see `from_chars` should not throw any exception.

### DIFF
--- a/headers/validation.h
+++ b/headers/validation.h
@@ -1266,7 +1266,7 @@ class Validator {
 		long long v;
 		auto begin = s.c_str(), end = begin + s.size();
 		auto [ptr, ec] = std::from_chars(begin, end, v);
-		if(ec == std::errc::invalid_argument){
+		if(ec == std::errc::result_out_of_range){
 			WA(name, ": Number " + s + " does not fit in a long long!");
 		} else if(ptr != end) {
 			WA(name, ": Parsing " + s + " as long long failed! Did not process all characters");
--- a/headers/validation.h
+++ b/headers/validation.h
@@ -1264,16 +1264,10 @@ class Validator {
 			WA(name, ": Want integer, found nothing");
 		}
 		long long v;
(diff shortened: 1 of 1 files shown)


headers/validation.h
```diff
@@ -1264,16 +1264,10 @@ class Validator {
...
-			if(ptr != end or ec != std::errc{})
-				WA(name, ": Parsing " + s + " as long long failed! Did not process all characters");
-		} catch(const std::out_of_range& e) {
-			WA(name, ": Number " + s + " does not fit in a long long!");
-		} catch(const std::invalid_argument& e) {
-			WA("Parsing " + s + " as long long failed!");
-		}
+		auto begin = s.c_str(), end = begin + s.size();
+		auto [ptr, ec] = std::from_chars(begin, end, v);
+		if(ptr != end or ec != std::errc{})
```
RagnarGrootKoerkamp, line 1269: can you explicitly add an out-of-range check, like here:
https://en.cppreference.com/w/cpp/utility/from_chars

mzuenni:
like this?

headers/validation.h
```diff
@@ -1264,15 +1264,14 @@ class Validator {
...
 		long long v;
-		try {
-			auto begin = s.c_str(), end = begin + s.size();
-			auto [ptr, ec] = std::from_chars(begin, end, v);
-			if(ptr != end or ec != std::errc{})
-				WA(name, ": Parsing " + s + " as long long failed! Did not process all characters");
-		} catch(const std::out_of_range& e) {
+		auto begin = s.c_str(), end = begin + s.size();
+		auto [ptr, ec] = std::from_chars(begin, end, v);
+		if(ec == std::errc::invalid_argument){
```
RagnarGrootKoerkamp, line 1269:
```suggestion
		if(ec == std::errc::result_out_of_range){
```

Issue #165, opened 2021-11-12
Clean up timeouts


RagnarGrootKoerkamp:
- There is the global program timeout for generators/visualizers/building (i.e.anything not submissions), which defaults to `30s` and is overridden by `--timeout` for e.g. `bt generate`
- There is a per problem submission timeout, which is derived from the timelimit and overridden using `--timeout` on `bt run`.

This makes for some inconsistency because now timeout is both a program and problem property.

We _could_ 'fix' the code and always make timeout a program property, which is either derived from the global timeout or from the problem timeout, but I'm not sure that's actually worth fixing.

I think we should fix the duplicate `--timeout` flag name, since this makes `bapctools.yaml` configuration a bit annoying otherwise. I'm not exactly sure how though. One option is to give a different name to one of them, but have `-T` as shortcut for both anyway. (`-t` is currently the shortcut for timelimit already.)

mpsijm:
> For problems, I typically just change the timelimit, not the timeout, if I want to see how long things take.

I agree with that one, so in that sense I'd say we could rename/rework how the timeout works. As I have written in https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/163#discussion_r748593805, we could change the submission timeout to work more like Kattis' `safety_margin`, so that the distinction with a Program timeout is clearer?

RagnarGrootKoerkamp:
Implementing the `limits:` options as problemtools uses them is a long outstanding issue anyway, so `safety_margin` is definitely something we should do. Still, on the command line you would want to pass an absolute time, not a factor, I think.

Thinking about it, `timeout:` in a `bapctools.yaml` really should be for the generators, not the problem. So maybe the cleanest is to just drop the `run --timeout` flag (or make it do the same as `generate --timeout`), and force the user to do this using `safety_margin` in the problem config or by just doing `-t/--timelimit`. Alternatively, we could make `--timeout` override both generator and submission timeouts when it is specified on the command line, but that's probably confusing.

mzuenni:
is this still relevant after #415 ?

mpsijm:
Almost, I think? Because I think `--timeout` still overrides all types of timeouts, whereas this issue discusses how you want to control the timeout for specific types of programs? Although one could argue that that's less relevant now that the timeout for generators, validators, and visualizers can be set in `problem.yaml` 🤔

mzuenni:
[...]

PR #207, opened 2022-05-10
add timeout for validators

mzuenni:
I am not sure if this breaks something else...

### DIFF
--- a/bin/config.py
+++ b/bin/config.py
@@ -47,6 +47,9 @@ KNOWN_DATA_EXTENSIONS = [
 
 SEED_DEPENDENCY_RETRIES = 10
 
+# The default timeout used for generators, visualizer etc.
+DEFAULT_TIMEOUT = 30
+
 # The root directory of the BAPCtools repository.
 tools_root = Path(__file__).resolve().parent.parent
 
@@ -64,6 +67,7 @@ default_args = {
     'jobs': os.cpu_count() // 2,
     'time': 600,  # Used for `bt fuzz`
     'verbose': 0,
+    'timeout': DEFAULT_TIMEOUT,
 }
 
 
--- a/bin/config.py
+++ b/bin/config.py
@@ -49,6 +49,7 @@ SEED_DEPENDENCY_RETRIES = 10
 
 # The default timeout used for generators, visualizer etc.
 DEFAULT_TIMEOUT = 30
+DEFAULT_INTERACTION_TIMEOUT = 60
 
 # The root directory of the BAPCtools repository.
 tools_root = Path(__file__).resolve().parent.parent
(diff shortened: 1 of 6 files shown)


bin/program.py
```diff
@@ -475,7 +476,7 @@ def run(self, bar, cwd, name, args=[]):
 
         if result.ok == -9:
             # Timeout -> stop retrying and fail.
-            bar.error(f'TIMEOUT after {timeout}s')
+            bar.log(f'TIMEOUT after {timeout}s',color=Fore.RED)
```
RagnarGrootKoerkamp, line 479: why should this not be an error?

bin/validate.py
```diff
@@ -150,6 +151,7 @@ def run(self, testcase, run=None, constraints=None, args=None):
                     expect=config.RTV_WA if testcase.bad_output else config.RTV_AC,
                     stdin=ans_file,
                     cwd=cwd,
+                    timeout=self.problem.settings.timeout,
```
RagnarGrootKoerkamp, line 154: I think it's better to read `config.args.timeout` directly. If you do `bt run --timelimit 2`, this will automatically set a timeout of `4` for submissions in `problem.settings.timeout`, which should not apply to validators.

Similar to https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/master/bin/program.py#L467 for generators.

Coul you also extract the `30` here and elsewhere (also in `utils.py`) into a flag in `config.py`?

bin/program.py
mzuenni: I am actually not sure if this is the right way to handle this... However, I noticed that for example the visualizer does no error handling in this part of the code. Instead, the errors get handled in `VisualizerInvocation` in `generate.py`. The Generator handled it in both parts which lead to calling `bar.error` twice in case of a timeout which seemed to be wrong (i got a weird stack trace from the bar class)

bin/validate.py
mzuenni: I hope I found all :D

bin/program.py
```diff
@@ -504,4 +505,4 @@ class Visualizer(Program):
     # Stdin and stdout are not used.
     def run(self, cwd, args=[]):
         assert self.run_command is not None
-        return exec_command(self.run_command + args, timeout=config.args.timeout or 30, cwd=cwd)
+        return exec_command(self.run_command + args, timeout=config.args.timeout, cwd=cwd)
```
RagnarGrootKoerkamp, line 508: The `or DEFAULT` is needed here. Passing None removes the timeout

RagnarGrootKoerkamp (approved):

bin/program.py
mzuenni: but `config.args.timeout` should never be `None`? it should get a default value (at least with the new commit where I fixed my attempt to add it to `set_default_args`).

bin/config.py
```diff
@@ -74,7 +78,7 @@
 grep -Ev '^(h|jobs|time|verbose)$' | sed "s/^/'/;s/$/',/" | tr '\n' ' ' | sed 's/^/args_list = [/;s/, $/]\n/'
```
RagnarGrootKoerkamp, line 78: Add timeout to this list

bin/interactive.py
```diff
@@ -37,7 +37,7 @@ def run_interactive_testcase(
     output_validator = output_validators[0]
 
     # Set limits
-    validator_timeout = 60
+    validator_timeout = 2*config.DEFAULT_TIMEOUT
```
RagnarGrootKoerkamp, line 40: Better make this a separate constant

RagnarGrootKoerkamp (approved):

bin/interactive.py
mzuenni: Done

RagnarGrootKoerkamp:
Thanks n

mpsijm:
@mzuenni With `bt run`, the abort timeout for TLE submissions now seems to be always 30 seconds, instead of `1.5 * time_limit`.  Is that intentional?

bin/config.py
```diff
@@ -64,17 +68,18 @@
     'jobs': os.cpu_count() // 2,
     'time': 600,  # Used for `bt fuzz`
     'verbose': 0,
+    'timeout': DEFAULT_TIMEOUT,
```
RagnarGrootKoerkamp, line 71: The new issue is probably caused by this, since the computation for 1.5*timelimit now thinks there is a timelimit set

mzuenni: Hmm, in that case, it is probably a bad idea to set the timeout with a default value. I could instead add a function like this:
```
def get_timeout(default=DEFAULT_TIMEOUT):
    return args.timeout or default
```
or i could add the `or DEFAULT_TIMEOUT` everywhere. I am not sure which is better practice.

RagnarGrootKoerkamp: Either is fine

RagnarGrootKoerkamp: Alternatively, you could explicitly check if the timeout argument is set for solution timeout computation

mzuenni:
That was not intentional and is somewhat related to #165? I was not aware that timeout is used for the TLE submissions ):

PR #208, opened 2022-05-17
don't set default timeout but use it

mzuenni:


bin/config.py
```diff
@@ -104,3 +99,10 @@ def set_default_args():
 # Set to true when running as a test.
 RUNNING_TEST = False
 TEST_TLE_SUBMISSIONS = False
+
+# The default timeout used for generators, visualizer etc.
+DEFAULT_TIMEOUT = 30
+DEFAULT_INTERACTION_TIMEOUT = 60
+
+def get_timeout(default=DEFAULT_TIMEOUT):
```
RagnarGrootKoerkamp, line 107: you can remove the argument

RagnarGrootKoerkamp (approved):

PR #211, opened 2022-06-15
added --timeout to bt fuzz

mzuenni:


mzuenni:
done :)

RagnarGrootKoerkamp:
Too many `time` flags here :see_no_evil:. But should be fine

PR #213, opened 2022-06-30
Print sample in full width if either input or output is too wide

mpsijm:
The old code checked only if the sum of input+output width was too large, but that does not cover all cases.


RagnarGrootKoerkamp:
pff; good hacking there :sweat_smile: 

Alternatively we could make them not the same width, but just making multiple lines is simpler and more consistent.

PR #214, opened 2022-07-04
Implemented more efficient binomial sampling

mzuenni:


RagnarGrootKoerkamp:
Hmm, this is a lot of extra code. Is it really needed? Anyway it needs some more explanation ideally.

mzuenni:
I don't know if it's really useful. I recently implemented this and remembered that `validation.h` uses naive sampling. This seems to be the shortest algorithm which is efficient. Giving more explanations is a bit difficult for this algorithm... it has many magic constants. However, the referenced paper has pseudocode which is pretty detailed and this implementation exactly matches that.

mzuenni:
I will close this for now. If we ever want faster binomial sampling we can still copy the code from my implementation (which by now is more or less the implementation of python)

PR #223, opened 2022-10-25
Time to get all those minor fixes upstream!

mpsijm:
All right, time to get all those minor fixes that I've done during BAPC merged upstream! :smile: See the commit log for the changes, most of them should speak for themselves.

The most complicated changes are in the fixes for contest exporting. I've tested exporting to a local DOMjudge instance:
- With and without setting `contest_id` in `contest.yaml` (or via command line)
- With and without `ruamel.yaml`
The `-v` flag now shows the `problems.yaml` and `contest.yaml` that are being uploaded, to ease debugging in the case that any of the values are formatted weirdly by the YAML import (which is what happened with the date/time values in `contest.yaml`).

P.S: The CI works again after updating `validation.h` from the BAPC repo :partying_face: 

P.P.S: Merging the new solve stats script into BAPCtools is still work-in-progress, will finish that later :slightly_smiling_face:

### DIFF
--- a/headers/validation.h
+++ b/headers/validation.h
@@ -47,6 +47,11 @@ bool operator<(const source_location& l, const source_location& r) {
 	return l.line() < r.line();
 }
 } // namespace std
+namespace std::experimental::fundamentals_v2 {
+bool operator<(const source_location& l, const source_location& r) {
+	return l.line() < r.line();
+}
+} // namespace std::experimental::fundamentals_v2
 #else
 constexpr bool has_source_location = false;
 struct source_location {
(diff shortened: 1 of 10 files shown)


headers/validation.h
```diff
@@ -1264,14 +1264,15 @@ class Validator {
...
+			if(ptr != end or ec != std::errc{})
+				WA(name, ": Parsing " + s + " as long long failed! Did not process all characters");
+		} catch(const std::out_of_range& e) {
 			WA(name, ": Number " + s + " does not fit in a long long!");
-		} else if(ptr != end) {
-			WA(name, ": Parsing " + s + " as long long failed! Did not process all characters");
-		} else if (ec != std::errc{}) {
-			WA(name, ": Parsing " + s + " as long long failed!");
+		} catch(const std::invalid_argument& e) {
+			WA("Parsing " + s + " as long long failed!");
```
RagnarGrootKoerkamp, line 1275: std::from_chars doesn't throw: https://en.cppreference.com/w/cpp/utility/from_chars

I think this hunk needs reverting

mpsijm: Apparently, this `try`/`catch` block was there initially in the BAPC repo, but it was removed after that (in April: https://github.com/RagnarGrootKoerkamp/BAPCtools/commit/ad35772ef75974dd300aa8013b1efa5a8895b0f0). Thanks for checking! :smile:

PR #224, opened 2022-10-26
validation.h: stold => std::from_chars

RagnarGrootKoerkamp:
@mpsijm @mzuenni  what do you think?

https://en.cppreference.com/w/cpp/utility/from_chars
vs
https://en.cppreference.com/w/cpp/string/basic_string/stof

### DIFF
--- a/headers/validation.h
+++ b/headers/validation.h
@@ -596,10 +596,10 @@ class Validator {
 
   public:
 	// No copying, no moving.
-	Validator(const Validator&) = delete;
-	Validator(Validator&&)      = delete;
+	Validator(const Validator&)      = delete;
+	Validator(Validator&&)           = delete;
 	void operator=(const Validator&) = delete;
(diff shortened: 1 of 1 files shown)


mpsijm (approved):
This looks like the same fix as https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/223#discussion_r1005651123, but for `long double` instead of `long long`, so I guess that's fine :smile: 

The CI failure seems to be related to `checktestdata`, not the validator. I've tested it on the problem set of BAPC 2022, and it seems to still work! :rocket:

mzuenni:
This is somewhat dangerous. Even though `from_chars` for doubles is part of `c++17` compilers only started to support this recently ):
But `from_chars` is preferable as soon as people stop using gcc 10.

RagnarGrootKoerkamp:
Ah good find. In that case let's wait.

Issue #228, opened 2022-11-25
Visualizer crash should not block generator


RagnarGrootKoerkamp:
Currently a crashing (TLE/RTE/...) visualizer stops generating the testcase. This shouldn't be a blocker for generating the testcase.

PR #234, opened 2022-11-29
allow failing visualizer

mzuenni:
#228


mpsijm (approved):
Calling `bar.error` should increment `config.n_error`, meaning that the command will have a non-zero exit code after a failing visualiser, so looks good to me :slightly_smiling_face:

Issue #195, opened 2022-03-15
Compact submission/testcase overview


mzuenni:
I would like to add a compact overview of all submission/testcases to the `bt run` command (similar to what domjudge displays).
It could look something like this:
![example](https://user-images.githubusercontent.com/19410663/158419468-08c90e97-0fb2-4d95-bc4e-620329436f4e.png)
Right now it first prints all testcases with an id and then displays a table with the verdicts. (A for AC, T for TLE, R for RTE and W for WA)

some thoughts about this
- less terminal pollution than -v but also less information
- the table could need some more tweaking (split samples and secret testcases, make the mapping from column to testcase clearer)

what do other think about this?

RagnarGrootKoerkamp:
@mpsijm @ludopulles wdyt?

I don't like the printing of the list of testcases so much. Could also do that separately in `bt list-testcases` or so.

I'd like it a lot if the table was rendered incrementally.

What if the number of testcases is larger than terminal width?

Numbering should continue from sample to secret, but could have a column of | in between

mpsijm:
Since the total number of test cases can become quite large, I don't think this should be the default. I think having `--table` flag for the `run` command is fine, because the behaviour of the command doesn't change, only the presentation. Additionally, printing the full list of test cases every time sounds like a bit much, so adding an extra command or flag to do that sounds good.

Other than that, I agree with @RagnarGrootKoerkamp :slightly_smiling_face:

ludopulles:
For rendering it seems like the current `bt run --table` works most likely. I think adding the verdict {A,T,R,W} with same green/red coloring is a nice way to add more info to the table.
May I ask what the added benefit is for transposing the table? I think this `--table` option is already rarely used, so perhaps adding a `--transpose` flag to it might be a bit too much.

mzuenni:
For me the transposed table would fit better on screen (especially if the row wrappes like described above) and I am using the table for different purposes i.e. I don't want to find a specific testcase but want to quickly see if the submissions fail on specific/multiple testcases. Therefore the transposed table feels more natural. Also note that the current `--table` command discards some rows since it has a different purpose

 Adding the ATRW to the current `--table` would also be nice I guess.

RagnarGrootKoerkamp:
@mzuenni could you make a PR for this?

mzuenni:
Yes, but right now it is always printed, we probably only draw this if a flag is passed?

[...]

PR #238, opened 2022-12-10
Overview

mzuenni:
See #195


RagnarGrootKoerkamp:
I'm wondering if we can/should combine the two tables into one:
existing: `Running <submission> <time> <submission verdict> <slowest case>`
new: `<submission> <testcases verdicts>`
could be merged to
`<submission> <testcase verdicts> <time> <submission verdict> <slowest case>`

Only when there are >>100 cases that would become problematic.

Basically your version is pretty similar to the old one but with a prettier progress bar.

I'm also not sure yet whether or not to show all not-yet-started submissions already from the start. I suppose the main benefit is that the table has a constant size and it prevents some jumping around.

I do have some flickering still. Would have to look in to this more but I suppose it all comes down to buffering and then flushing at the right times.

RagnarGrootKoerkamp:
Screenshot for context
image

RagnarGrootKoerkamp:
Maybe stderr buffer size can be increased more using something like https://stackoverflow.com/a/65035026/2716069.
That should reduce flickering for large tables. On my screen the flickering typically happens in the same position each time, which seems to suggest that even without line buffering the stderr is flushed regularly

RagnarGrootKoerkamp:
Oh, we should either test or disable this for windows

RagnarGrootKoerkamp:
I'm leaning towards enabling this by default. Now that it (seems to) work well I don't really see a reason not to. On terminals without overwriting you already needed `--no-bar` anyway so that doesn't change. Will have to test `--no-bar`, `-v`, and `--table` a bit more later

mzuenni:
I can remember that the table behaves a bit weird if not all submission with testcases can be displayed at the same time, i.e. if the Terminal is too small. But i am not sure if that can be fixed at all. Maybe you want to try it on an NWERC problem with maaany submissions and many testcases

bin/util.py
```diff
@@ -391,18 +406,159 @@ def finalize(self, *, print_done=True, message=None):
...
+                v = self.results[s][testcase]
+                if v == 'ACCEPTED':
+                    return Fore.GREEN + 'A' + Style.RESET_ALL
+                else:
+                    return Fore.RED + v[0] + Style.RESET_ALL
+            elif s + 1 == len(self.results) and testcase in self.current_testcases:
+                return Style.DIM + Fore.BLUE + '?' + Style.RESET_ALL
+        return Style.DIM + Fore.WHITE + '-' + Style.RESET_ALL
+
+    # TODO only print this if a flag is given! (config.args.???)
```
mpsijm, line 518: I was going to propose `--table` as flag, but turns out we already have that :stuck_out_tongue: The only difference is that the existing `--table` flag has the test cases on the rows, instead of the submissions. Some options:
- Rename `--table` to `--verdict-analysis` (because that's what the existing table is used for), and use `--table` for this new Overview
- Keep `--table` what it is now, and think of a new name for this new Overview (like `--overview` or `--progress`)

mpsijm:
> Maybe you want to try it on an NWERC problem with maaany submissions and many testcases

This is when running it at NWERC 2023 Lateral Damage on a terminal with 146×36 characters. The problem has 134 secret test cases and 26 jury submissions in total. For the first few minutes, I don't see anything happening, and scrolling up is very difficult as well, because stuff is being rewritten outside the viewport. :sweat_smile: 

| For the first few minutes | When it's done | When I zoom out a bit (380×82 chars) |
|---|---|---|
| image | image | image |

Also notice how, after everything's finished, scrolling up still doesn't get me to the list of submissions with their maximum time.

mzuenni:
Yes this is broken as soon as the table is larger than the terminal area, but i fear that is unfixable... thats why i don't want it to be the default. One addition would be to only display if a flag is given and then either print "sorry termianl too small for table" or print the table if it fits

RagnarGrootKoerkamp:
You can query terminal width and wrap lines manually?

mpsijm:
> You can query terminal width 

Yes, this is possible, same for the height

> and wrap lines manually?

The lines are already wrapped manually, see my screenshot: they're cut off at multiples of 10. Besides, this doesn't really affect the usability, because the problem is mostly that there are too many lines, so that scrolling becomes impossible.

> One addition would be to only display if a flag is given and then either print "sorry terminal too small for table" or print the table if it fits

I think that would be my preference :slightly_smiling_face: We could even add an extra flag that always forces the display of this table, even when the terminal is too small.

The calculation for whether the terminal is too small will probably somewhat arbitrary, though. My suggestion:
- First, detect the number of lines needed per submission in the table. For problems with few test cases, this will simply be `1`, but for Lateral Damage, this will be `2`, as seen in my screenshot.
- Calculate the total number of lines needed to fit all submissions on screen.
- If the total number of lines is smaller than half* the terminal height minus 5, do not show the table.

*) "half" is somewhat arbitrary, I chose this so that the "regular" output of `bt run` will still be visible. I guess something like ⅔ or ¾ could also work, because then at least it's visible that something is happening above the table.

mpsijm:
I think the only thing still missing right now, is a mention in the documentation? Other than that, I think this is ready for merging :smile:

mpsijm (approved):

RagnarGrootKoerkamp:
I won't have time to review. Feel free to merge

PR #239, opened 2023-01-01
support bmp for visulizers

mzuenni:


RagnarGrootKoerkamp:
The spec doesn't actually support BMPs. Not sure if domjudge does either. This can still be useful for local development, but maybe a (verbose) log that it's out-of-spec would be good?

mzuenni:
Hmm, in that case, I will first look at how hard it is to write a raster image in one of the supported formats... but I guess png will be much more complicated ^^'

mpsijm:
In that case, if you want to write a BMP file manually, you could use a conversion tool to convert it to PNG? :slightly_smiling_face:

mzuenni:
I just used BMP because it's easy to write in c++ without any library... how could I add the conversion tool in the generator.yaml?

mpsijm:
Ahhh good point, using libraries is hard for compatibility... never mind then :sweat_smile: Unless you ask all your jury members to install ImageMagick, for the `convert` command :upside_down_face:

mzuenni:
suppose everyone had the tools installed, how would i do that?
`/generators/visualizer.cpp {name} && convert {name}.bmp {name.png}`
does not work? or would you want to do a `std::system` call in the visualizer?

RagnarGrootKoerkamp:
Yeah I think it's ok to simply accept .bmp for bapctools purposes, but for export it's useless. We could auto-convert them only on export though?

mzuenni:
There is a python lib that could do this but I think it's a bit weird to add special handling for this in bapctools, if the spec does not support it... I would prefer to get it in the spec though. For all purposes, I can think of, it sounds more reasonable to add a run script which does the conversion either with `convert` or with one line of python.

RagnarGrootKoerkamp:
I'm happy to merge this, but ideally we do have a solution for exporting them as well. Anything works (fixing the spec; fixing the export; showing a warning), but I don't have time to work on this. Allowing to generate .bmp without being able to export them sounds suboptimal anyway.

The simplest may be to update the export job with a warning when it encounters .bmp files?

mzuenni:
I don't know how to get this to the spec, therefore I wrote a few functions to write an uncompressed PNG ^^'. I think we can just delete this pull request.

RagnarGrootKoerkamp:
but closing this is also OK

PR #240, opened 2023-01-04
[export] Add `--colors` parameter to `update_problems_yaml` command

mpsijm:
This PR adds a `--colors` parameter to the `update_problems_yaml` command, which can be used to set a list of hex colours from the command line. Typically, when I get a list of balloon colours from an organisation, it's easier to transform this into a comma-separated list than to manually copy all the hex codes :stuck_out_tongue: 

I've also added the command (including this new parameter) to the [documentation](doc/commands.md) :smile:

P.S. I would've preferred to use the British "colours" instead of "colors", but it looks like most of BAPCtools uses American English (which is consistent with most programming tools, so I guess it's fine :upside_down_face: )

### DIFF
--- a/bin/export.py
+++ b/bin/export.py
@@ -285,7 +285,7 @@ def update_problems_yaml(problems, colors=None):
 
     log('Updating problems.yaml')
     path = Path('problems.yaml')
-    data = read_yaml(path) if path.is_file() else []
+    data = path.is_file() and read_yaml(path) or []
 
     change = False
     for problem in problems:
--- a/bin/export.py
+++ b/bin/export.py
@@ -275,7 +275,7 @@ def export_contest():
     return new_cid
 
 
-def update_problems_yaml(problems):
+def update_problems_yaml(problems, colors=None):
     # Update name and timelimit values.
     if not has_ryaml:
         log(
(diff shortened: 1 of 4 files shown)


bin/export.py
```diff
@@ -321,6 +321,13 @@ def update_problems_yaml(problems):
                 }
             )
 
+    if colors:
+        for d, c in zip(data, colors):
+            color = ('' if c.startswith('#') else '#') + c
```
RagnarGrootKoerkamp, line 326: Simply `c or '#'+c` is clearer imo

bin/export.py
```diff
@@ -321,6 +321,13 @@ def update_problems_yaml(problems):
...
+            if d['rgb'] != color:
+                change = True
```
RagnarGrootKoerkamp, line 328: Maybe check the field exists?

bin/export.py
```diff
@@ -321,6 +321,13 @@ def update_problems_yaml(problems):
                 }
             )
 
+    if colors:
+        for d, c in zip(data, colors):
```
RagnarGrootKoerkamp, line 325: Add comment/doc that order is as in problems.yaml, and check the lists have the same length to be sure.

bin/tools.py
```diff
@@ -966,7 +970,9 @@ def run_parsed_arguments(args):
         if action in ['export']:
             export.export_contest_and_problems(problems)
         if action in ['update_problems_yaml']:
-            export.update_problems_yaml(problems)
+            export.update_problems_yaml(
+                problems, config.args.colors.split(",") if config.args.colors else []
```
RagnarGrootKoerkamp, line 974: None instead of []

bin/export.py
mpsijm: That does not check whether `c` starts with a `#`. The expectation is that `c` is a string of the shape `#rrggbb` or `rrggbb`, and this normalises `c` to always start with `#`

bin/export.py
mpsijm: The field is guaranteed to exist, thanks to lines 302 and 319 :slightly_smiling_face:

bin/tools.py
mpsijm: Thanks, fixed :smile:

bin/export.py
mpsijm: Never mind, this is not true when the problem directory is not found, will fix!

PR #241, opened 2023-01-20
[solve_stats] Fix #236: Port script used for past contests to BAPCtools

### LINKED ISSUE #236: Merge updated solvestats in solution slides
@mpsijm Could you make a PR for this at some point?

mpsijm:
Behold, the shiny new solve stats script to fix #236! :star_struck: 

This new `bt solve_stats` command executes roughly the following steps:
- Fetch contest data from CCS
- Calculate statistics about the submissions
- Plot one activity chart per problem
- Output a LaTeX file that populates the `\solvestats` command
- Plot language statistics

Note that the build fails, because the Docker image doesn't have `python-matplotlib` as dependency yet. I've added it to the `Dockerfile`, but I can't push the new container of course :stuck_out_tongue: See also RagnarGrootKoerkamp/bapctools-git#1, for an update to the AUR `PKGBUILD`.


bin/solve_stats.py
```diff
@@ -0,0 +1,141 @@
...
+    judgement_types = req_assoc('judgement-types')
+    judgement_types[''] = {'id': '', 'name': 'pending'}
+    judgement_colors = {'AC': 'lime', 'WA': 'red', 'TLE': '#c0f', 'RTE': 'orange', '': 'skyblue'}
+
+    for j in req('judgements'):
+        # Note that the submissions list only contains submissions that were submitted on time,
+        # while the judgements list contains all judgements, therefore the submission might not exist.
+        if j['valid'] and j['submission_id'] in submissions:
+            # Merge judgement with submission. Keys of judgement are overwritten by keys of submission.
+            submissions[j['submission_id']] = {**j, **submissions[j['submission_id']]}
```
RagnarGrootKoerkamp, line 53: what is happening here? Can you not do something like `submissions[id]['judgement'] = j`?

How does this handle multiple judgements for a single submission? Is `'valid'` the one that is 'active'?

bin/solve_stats.py
```diff
@@ -0,0 +1,141 @@
...
+    ac_teams = {p: set() for p in problems}
+    stats = {p: [{j: 0 for j in judgement_types} for _ in range(bins)] for p in problems}
+    stats_sum = {p: {j: 0 for j in judgement_types} for p in problems}
+    language_stats = {l: {j: 0 for j in judgement_types} for l in languages}
+
+    for i, s in submissions.items():
+        if s['team_id'] not in teams:
+            continue
+        minute = time_string_to_minutes(s['contest_time'])
+        if 0 <= minute < contest_duration:
```
RagnarGrootKoerkamp, line 66: does this work for submissions in the last second? probably it does

bin/solve_stats.py
```diff
@@ -0,0 +1,141 @@
...
+            + label
+            + r'}{\printsolvestats{'
+            + '}{'.join(
+                str(x) for x in [sum(stats_sum[p].values()), len(ac_teams[p]), stats_sum[p]['']]
+            )
+            + '}}\n'
+        )
+        bar.done()
+
+    open(f'solve_stats/problem_stats.tex', 'w').write(problem_stats)
```
RagnarGrootKoerkamp, line 123: can you do this with `Path(<path>).write_text(..)` instead

RagnarGrootKoerkamp (commented):
some small changes, but looks pretty nice!

Cool how doing the work yourself is only 2x longer than calling the previous tool :)

bin/solve_stats.py
```diff
@@ -0,0 +1,141 @@
...
+
+from contest import call_api, get_contest_id
+from util import ProgressBar
+
+
+def generate_solve_stats(post_freeze):
+    contest_id = get_contest_id()
+
+    # The endpoint should not start with a slash
+    def req(endpoint):
```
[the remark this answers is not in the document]
mpsijm: I am using `util.call_api` inside `req`, which is probably the function you're referring to. This local `req` function is specialised a bit further for inside the `generate_solve_stats` function: it handles the `bar`, possible errors thrown by the request, and returns `r.json()`. In other places, `util.call_api` is used slightly differently.

bin/solve_stats.py
mpsijm: That's probably cleaner indeed, thanks for the suggestion! :slightly_smiling_face: 

The DOMjudge API does not state anything about the `'valid'` field, but from what I've seen, I think your idea is correct.

bin/solve_stats.py
mpsijm: It should, yeah: `time_string_to_minutes` returns a `float`, so `299.983` should™ be the last second of the contest. This is, of course, assuming that submissions at `5:00:00.000` receive the `TOO-LATE` verdict: I'm not 100% sure about that, but if this ever happens, nobody is gonna notice it :stuck_out_tongue:

bin/solve_stats.py
mpsijm: Costs an extra import, but sure :stuck_out_tongue:

bin/solve_stats.py
RagnarGrootKoerkamp: ok can you add the thing about 'valid' as a comment

mpsijm: Sure, done :slightly_smiling_face:

PR #242, opened 2023-01-29
[problem,run] Make behaviour of `@EXPECTED_RESULTS@: ` consistent with DOMjudge

mpsijm:
The `@EXPECTED_RESULTS@: ` tag is ignored when an imported submission is one of the standard submission directories (e.g., accepted/, wrong_answer/), as discussed in https://github.com/DOMjudge/domjudge/issues/1861. I've made BAPCtools consistent with this behaviour, to give less confusion when importing contest problems.


RagnarGrootKoerkamp:
hmm, this probably breaks many cases where we used it inside a normal directory. But oh well, those were already broken in domjudge anyway.

test/problems/hello/submissions/accepted/test-hello-lf-cr.c
```diff
@@ -1,8 +1,6 @@
 /*
  * This should give CORRECT or WRONG-ANSWER on the default problem
```
RagnarGrootKoerkamp, line 2: move this file to rejected/?

bin/run.py
```diff
@@ -401,15 +401,19 @@ def _get_expected_verdicts(self):
...
+            # See https://github.com/DOMjudge/domjudge/issues/1861
+            subdir = self.short_path.parts[0].upper()
+            if subdir in config.VERDICTS:
+                if len(verdicts) != 0:
+                    warn(f'@EXPECTED_RESULTS@ in submission {self.short_path} is ignored.')
                 verdicts = [subdir]
+            else:
+                if len(verdicts) == 0:
+                    error(
+                        f'Submission {self.short_path} must have @EXPECTED_RESULTS@. Defaulting to AC.'
```
RagnarGrootKoerkamp, line 415: Maybe ACCEPTED for consistency with the tag you're supposed to add

bin/run.py
```diff
@@ -401,15 +401,19 @@ def _get_expected_verdicts(self):
...
 
-        subdir = self.short_path.parts[0].upper()
-        if subdir in config.VERDICTS:
-            if len(verdicts) > 0 and subdir not in verdicts:
-                error(
-                    f'Submission {self.short_path} must have implicit verdict {subdir} listed in @EXPECTED_RESULTS@.'
-                )
-                verdicts = [subdir] + verdicts
-            elif len(verdicts) == 0:
+        if self.path.parts[-3] == 'submissions':
```
RagnarGrootKoerkamp, line 404: `[-3]` is ugly, but it works and is fine :shrug:

mpsijm:
> this probably breaks many cases where we used it inside a normal directory. But oh well, those were already broken in domjudge anyway.

That's exactly the point of this PR :stuck_out_tongue: Without the warnings, we wouldn't know that DOMjudge would have different behaviour than you might expect :slightly_smiling_face:

bin/run.py
mpsijm: Yeap, I didn't know a better way to make this consistently go right :sweat_smile: Ugly but works, indeed :upside_down_face:

test/problems/hello/submissions/accepted/test-hello-lf-cr.c
mpsijm: It is always `ACCEPTED` in BAPCtools because it does not have that kind of white-space config (it always normalises whitespace, I think). But moving it to `mixed/` doesn't hurt indeed :slightly_smiling_face:

RagnarGrootKoerkamp:
thank!

PR #244, opened 2023-02-09
override problem.tex

### LINKED ISSUE #243: Problem PDF
For the nwerc we added code to change the context pdf (see 0ce33b42757616811685e10d212f58017d00bcea). However, this does not apply to the problem pdf build for each problem individually. Unfortunately, these pdfs are displayed in the DOMjudge and in the zip which is pretty inconsistent... (see [nwerc problems zip](https://commissies.ch.tudelft.nl/chipcie/archive/2022/nwerc/solutions.zip))

mzuenni:
fixes #243

### DIFF
--- a/bin/latex.py
+++ b/bin/latex.py
@@ -171,8 +171,11 @@ def build_problem_pdf(problem, solutions=False):
 
     builddir = problem.tmpdir
 
+    local_data = Path(main_file)
     util.copy_and_substitute(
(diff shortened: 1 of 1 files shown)


RagnarGrootKoerkamp:
Did you test this locally? If so, i'll merge :)

mzuenni:
Yes, tested it locally with a modified problem.tex.

PR #246, opened 2023-02-18
[export] Export single problem, check if user has a team, and documentation updates

mpsijm:
### Support (re-)exporting a single problem
This is useful when a contest is already imported, but you change the time limit or some test cases for a single problem afterwards.<br>
Note that this probably doesn't work on DOMjudge 8.1.3 and older.
These older DOMjudge versions require `delete_old_date: True` to work properly
(see DOMjudge/domjudge@cb68f9b).

### Add documentation to doc/commands.md
Speaks for itself :slightly_smiling_face: 

### Check if exporting user is associated to a team
I made the check in such a way that it works for the different "Data source" configurations in DOMjudge (i.e., this code should work both when using internal (numeric) IDs and when using external IDs).

### Remove TODO about reusing problem IDs because it's not possible
Same deal about the "Data source" configuration holds here:
we could probably get away with reusing the result of `export_problems` if DOMjudge would always use external IDs.
However, with numeric IDs, they cannot be matched directly with the problems that we are exporting.

### DIFF
--- a/bin/export.py
+++ b/bin/export.py
@@ -376,7 +376,7 @@ def export_problems(problems, cid):
         fatal(f'{msg}\n{e}')
     log(f'Uploaded problems.yaml for contest_id {cid}.')
     data = yaml.load(r.text, Loader=yaml.SafeLoader)
-    return data
+    return data  # Returns the API IDs of the added problems.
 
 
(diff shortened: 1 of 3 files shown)


RagnarGrootKoerkamp:
thanks :)

PR #249, opened 2023-03-11
Improve startup time

mpsijm:
I noticed that tab-completion was slower than it used to be. After some investigation, it turns out that I caused this myself :innocent: 

Turns out that importing Matplotlib takes more than a second to resolve. So, instead of doing a conditional import at the top of the file, I've moved the import inside the function, so that it's only imported when it's actually used:

| Before | After |
|---|---|
| image | image |

And while I was at it, I found that the Requests library also takes significantly longer to resolve than most other packages, so I did the same there:

| Before | After |
|---|---|
| image | image |

P.S. @RagnarGrootKoerkamp Can you also add the `python-requests` dependency to https://github.com/RagnarGrootKoerkamp/bapctools-git/? :slightly_smiling_face:

### DIFF
--- a/bin/contest.py
+++ b/bin/contest.py
@@ -1,12 +1,7 @@
 from pathlib import Path
-from util import *
 import json
 
-# Optional, only needed for API stuff
-try:
-    import requests
-except:
-    pass
+from util import *
 
 # Read the contest.yaml, if available
 _contest_yaml = None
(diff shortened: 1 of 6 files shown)


RagnarGrootKoerkamp:
update AUR package and docker image

mpsijm:
@RagnarGrootKoerkamp thanks! :smile: I don't see the update in https://github.com/RagnarGrootKoerkamp/bapctools-git though, and at https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=bapctools-git the description of the optional dependency `python-requests` is copy/pasted from questionary, but it should be something like "for commands that call the DOMjudge API or Slack API" :slightly_smiling_face:

RagnarGrootKoerkamp:
updated comment and pushed to both github and AUR now :)

Issue #109, opened 2021-04-21
Support multiple languages


RagnarGrootKoerkamp:
Maybe something like this:
Have a `language: <en|nl|...>` in `contest.yaml` and/or `problem.yaml`, so that `bt` will look for `problem.<language>.tex` instead of the default `problem.en.tex`.

A clean way to change the latex constants would be to have language specific `en.tex`, `nl.tex`, ... includes that contain all the hardcoded texts as commands.

Most of the [diff to the NIO fork](https://github.com/RagnarGrootKoerkamp/BAPCtools/compare/master...ludopulles:nio) should be resolved by this.

CC @ludopulles

mzuenni:
maybe use [translator](https://ctan.org/pkg/translator) for latex with `<language>.dict` files.

thorehusfeldt:
Regarding specification, `language` in `problem.yaml` would be very close to the already-existing `languages` (which specifies programming languages.)

Suggestion: according to the specification, the `name` key in `problem.yaml` already takes map; it can look like this:
```yaml
name:
  en: Danish Pastry
  nl: Stroopwafel
  de: Kopenhagener Gebäck
  da: Wienerbrød
```
From the keys in that map, we can infer the four languages and expect `problem.nl.tex` to exist (and complain otherwise).

RagnarGrootKoerkamp:
I'd be happy to review a PR for this. I don't imagine this would be too hard to support once you know where to look. I could probably give some pointers if you'd like.

(We don't use/need this for BAPC, and sadly I don't really have time and motivation to write new code.)

thorehusfeldt:
OK, I like this idea. My current ambition is to use BAPCtools for the next Will Code for Drinks, scheduled for November 2023; but I need to find my way around it a bit more before I commit to anything. 

I forked the repo for now and implemented the default grader, which doesn’t really have much to do with statement languages to be honest.

thorehusfeldt:
First attempt and discussion at https://github.com/RagnarGrootKoerkamp/BAPCtools/discussions/254

thorehusfeldt:
@RagnarGrootKoerkamp : my https://github.com/thorehusfeldt/BAPCtools/tree/i18n branch now supports some basic functionality, such as 

```
myproblem> bt pdf
myproblem> bt pdf --language nl
cd ..
mycontest> bt pdf
mycontest> bt pdf --language nl
```
[...]

PR #256, opened 2023-05-18
Add support for multiple languages, addresses #109 

thorehusfeldt:
* Adds `--language` to `bf pdf` and `bt new_problem`
* `bt rename_problem` is language-aware
* Introduces the configuration `config.args.languages`
* Problem metadata `problem.yaml` now always uses a map for `language` (instead of a string)
* When no other language is specified anywhere, defaults to `en`
* When only one language is specified (or none, which implicitly specifies `en`), `bt pdf` produces `problem.pdf`; otherwise it produces `problem.xy.pdf` for all languages `xy` for which it makes sense
* Documentation in `docs/multiple_languages.md`


thorehusfeldt:
It would be cute if the logging of PDF jobs used the ProgressBar class, if only to save a few lines of output.

I get this far, at least:
image
but I’m honestly stumped by the logic about `log`, `finalise`, and `done` and need help. All we want is to write “Building PDF for de” at the start of compilation and then confirm success and write the name of the output file (or report `error`) on the same line. (Getting the parallelisation logic to work would be even more ambitious but completely beyond my ken.)

RagnarGrootKoerkamp:
I'll take a look at your commits and try to fix the PDF progress reporting. (Yes it's complicated :see_no_evil:. Somehow supporting all the slightly different ways in which progress bars are used with/out overwriting became quite tricky.)

RagnarGrootKoerkamp:
I'm back from holidays :) Could you reach out to me on chat somewhere to discuss where you need my input/help/reviewing?

RagnarGrootKoerkamp:
@ludopulles since you translated to Dutch for NIO, you may be interested. Would be cool if we could make it so that you don't need to maintain a fork anymore.

@obiwac same for you for French.

I wasn't aware of the French fork before. Given that there are now at least to forks with the main intent to simply translate, I'm more invested to make this actually work smoothly

mzuenni:
Regarding the `Input`, `Output` environment vs `\section*{}`: `bt constraints` won't recognize any `\section` command that is not `Input` or `Output`. So we should find a way to translate the the environment.

thorehusfeldt:
(rebased my i18n branch on top of master at https://github.com/RagnarGrootKoerkamp/BAPCtools)

chrisgdt:
Incredible work ! I started a few weeks ago a similar feature on the [French fork](https://github.com/karwa-org/BAPCtools/tree/tongue) but I left it behind me because of my master's thesis !

A suggestion if you mind : Create a config file to translate some hard-coded words of the contest pdf (e.g. "Time limit" on the top right corner, "This page is intentionally left blank", "Sample" for in/output examples, ...), the full list in on [my last commit](https://github.com/karwa-org/BAPCtools/commit/6b4edd0fee0b5fb0b4310e210d2d3eb1b217c165#diff-18665cbb936232bc6bdb32ad4b97f0da64cf1111e7fcb647b5419b9a45b45012). As these words are directly in the LaTeX code, this config file should somehow be a tex file manually integrated into the full tex file before Python calls LaTeX to compile.

Maybe I will have some time to get into it and help you but I'm not sure yet. Anyway, the French fork was not expected to stay long, only to translate those hard-coded words English so far.

RagnarGrootKoerkamp:
Oh this looks great! Thanks for your interest and the example code; having something to start is nice.

I'll have to spend some time thinking about how this will work _exactly_, but the config file you have seems like a promising approach.

This way we can make most files language-independent, and only inject some `lang-en.tex` files or so.
Probably I'd go for something like:
- add some default `latex/lang-xy.tex` (or maybe `latex/lang/xy.tex`)
- allow users to also put a custom `lang-wz.tex` in the `contest/` directory, so you don't need to fork bapctools for a one-off language (and then you can upstream it separately/later/never).
- Include this file at the top of the current class files, so the commands are always available.

Then you can also modify `contest.tex` without having to duplicate it for each language.

RagnarGrootKoerkamp:
Some progress:
- Merged @chrisgdt's branch.
- Moved lang-files to `latex/lang/xy.tex`. Putting them in a contest directly as `contest/xy.tex` also works.
- Added `input{lang/{%stmlang%}.tex}` to `problem.tex` (main problem-pdf entrypoint), and `contest-data.tex` (included for each contest-pdf and solutions-pdf).
- Renamed `solution.tex` to `solution.xy.tex`. A warning to rename is printed if only `solution.tex` is found.
- `contest/solution-{header,footer}.tex` can be used under that name, or, if present, replaced by `contest/solution-{header,footer}.xy.tex`

RagnarGrootKoerkamp:
woop woop tests are green :heavy_check_mark: :D 
I think this should be ready to merge

RagnarGrootKoerkamp:
Made it so that we always write to `problem.xy.tex` for simplicity and consistency, and updated/added some docs.

RagnarGrootKoerkamp:
Fixed some tests, comments, and other things that still referred to `{problem,contest,solution,solutions}.pdf` instead of the new versions with language suffix. Also updated `bt zip` with an optional `--language xy` so that the zip made for domjudge export can choose the right `problem.xy.pdf` as source for the `problem.pdf` in the final exported `.zip`. (But this is only required if the union of languages over all problems is larger than 1.)

chrisgdt:
Hey, it seems that I don't have the permission to add a commit here. Can you add this translation file `fr.tex` for me please ? :)

```tex
\newcommand{\lang}{fr}% FRENCH

% bapc.cls
\newcommand{\langblank}{Cette page est intentionnellement laissée vierge.}
\newcommand{\langtimelimitmissing}{LIMITE DE TEMPS MANQUANTE}
\newcommand{\langproblemname}{Nom du problème}
\newcommand{\langproblemauthor}{Auteur du problème}
\newcommand{\langproblems}{Problèmes}
\newcommand{\langtimelimit}{Limite de temps}
\newcommand{\langproblem}{Problème}
\newcommand{\langinputsample}{d'entrée}% Used in e.g. "Input Sample 1"
\newcommand{\langoutputsample}{de sortie}% Used in e.g. "Output Sample 2"
\newcommand{\langinteraction}{Interaction}
\newcommand{\langexplanation}{Explication des exemples}
\newcommand{\langsample}{Exemple}
\newcommand{\langinput}{Entrée} % Input section heading
\newcommand{\langoutput}{Sortie}
\newcommand{\langread}{Lire}
\newcommand{\langsampleinteraction}{Interaction d'exemple}
\newcommand{\langwrite}{Écrire}

% solutions-base.tex
\newcommand{\langstatistics}{Statistiques}
\newcommand{\langsubmissions}{soumissions}
\newcommand{\langaccepted}{acceptées}
\newcommand{\langunknown}{inconnues}
```

thorehusfeldt:
There’s a rich set of translations here, by the way: https://github.com/nordicolympiad/baltic-olympiad-2018/tree/master/templates 

(The German translation there has _Aufgabe_ but _Zeitlimit_, by the way. :-) )

mzuenni:
Hm, in that case, I would stay with Aufgabe but change
```
\newcommand{\langtimelimit}{Zeitlimit}
\newcommand{\langsubmissions}{Abgabe}
```
and maybe even simplify this:
```
\newcommand{\langread}{\langinput}
\newcommand{\langwrite}{\langoutput}
```

mpsijm:
@chrisgdt Added it for you, thanks for providing it! :smile: 

@ludopulles Any reason why the Dutch version of `\langblank` is missing a translation of "intentionally"? I think I would write something like "Deze pagina is opzettelijk leeg (gelaten)." (with "gelaten" as translation of "left", but I don't think that this word is strictly necessary). 

Another fun suggestion thanks to [Linguee](https://www.linguee.nl/nederlands-engels/search?source=auto&query=this+page+is+intentionally+left+blank): "Deze pagina is opzettelijk blanco (gelaten)."

thorehusfeldt:
We’re in geek mode, but this has a name in German: https://de.wikipedia.org/wiki/Vakatseite, see also https://en.wikipedia.org/wiki/Intentionally_blank_page .

RagnarGrootKoerkamp:
@mpsijm My `nl.tex` was just a quick draft without much thinking, that I'm not very happy with anyway. Feel free to modify (both you and/or @ludopulles with NIO data).

@mzuenni / @thorehusfeldt  feel free to change the de.tex as you like.

Re. the `read/write` for interactive problems, I think I copied the worldfinals template (made by Per Austrin I think). I suppose the reason is to set it apart more from the usual input/output, but yes I agree when naming pipes it's always unclear whether you're naming them on the submission or validator end.

mpsijm:
Comparing with https://github.com/RagnarGrootKoerkamp/BAPCtools/compare/master...ludopulles:BAPCtools:nio, I changed "probleem" to "opgave" (also my personal preference :innocent:), and all other translations appear to be consistent with what you drafted :slightly_smiling_face: 

During this comparison, I found two more things that need to be internationalized:
- We currently use `\usepackage[english]{babel}`, but we should probably switch `[english]` per language: https://github.com/RagnarGrootKoerkamp/BAPCtools/commit/6c0e591ea9e1ac84f41824a7f29cd352c86a0153. I think we can simply add a key to `latex/lang/*.tex` for this, as long as this file is loaded before `babel` is loaded.
- The copyright notice on the second page of the contest PDF is not translated (and also is not translated for NIO: https://github.com/RagnarGrootKoerkamp/BAPCtools/commit/a0ebdf4a2da85304b780b91a9ecd39636e1ef4b0#diff-fbec4113ad51f182f1c2b16fd792d22c67a93f57470137e9ea2596d643b919ab). Do we even want to translate this? The use of the variables `\@year` and `\@author` in the strings make it a bit of a hassle, unless we make a parameterized language string, e.g.:
      
  ```latex
  \newcommand{\langcopyright}[2]{
      Copyright \copyright\ #1\ by #2.
      This work is licensed under the
      \\
      Creative Commons Attribution-ShareAlike 4.0 International License.
  }
  ```

mpsijm:
I've experimented with providing `\lang` as `-usepretex` to `latexmk`, so that we only need a single `\input{lang/\lang.tex}` in `bapc.cls` and `solutions-base.tex`, instead of having to import it everywhere. This also allows setting the Babel option to the correct language from `bapc.cls` :smile: I'm not 100% sure if this is the best way to go, though, so I've pushed this to my own fork: https://github.com/mpsijm/BAPCtools/commit/713cc124e875a7a8979e3bb112ed57cb1ca02cb0

Also, a bug report: I tried running `bt solutions` in `test/problems/hellowholeworld`, but that tries to compile Danish slides even though they don't exist. It looks like `Problem._determine_statement_languages` only checks `problem.*.tex`, not `solution.*.tex`. If we want to do that, how would that work? Do we want both `Problem.statement_languages` and `Problem.solution_languages`, or just throw an error if no matching `solution.X.tex` is found while there exists a `problem.X.tex`?

RagnarGrootKoerkamp:
* Yes, defining `\lang` in the latex invocation sounds good. I was wondering if something like that exists but didn't bother looking into it. This does somewhat increase the coupling between the latex files and needing BAPCtools to build them, but that's already very much the case anyway so it's OK I think.
* Oh right; for contest-wide solutions I simply skip over problems where the `solution.xy.tex` does not exist for the given language, but for single-problem solutions that doesn't make much sense. Just iterate over all `solution.xy.tex` files that are present makes more sense here.

mpsijm:
- Thanks, I'll push it! :smile: About the coupling: partly yes, but not more than how it was with the `{% stmlang %}` variable. And in the version with `\lang`, there's always a fallback to `en` :slightly_smiling_face: 
  - I've also fixed a bug related to setting the working directory when compiling a contest. Apparently, this required passing `%O` to `pdflatex` as well :stuck_out_tongue: (follow-up to the discussion at https://github.com/mpsijm/BAPCtools/commit/713cc124e875a7a8979e3bb112ed57cb1ca02cb0#r120586496)
- Sounds good! Would it make sense to add a warning if no `solution.X.tex` exists, while a `problem.X.tex` does exist?

RagnarGrootKoerkamp:
Merged :tada: Thanks all for your contributions! This probably the biggest PR (in #commits) so far :smile: 
If there are further issues we can fix them directly on master or in a new branch.

PR #260, opened 2023-06-17
added notes after samples

mzuenni:
This is a bit hacky but works...

tldr:
use `\begin{Notes} ... \end{Notes}` to add notes after the samples. This is done by storing the environment content in a variable that's rendered at the end


RagnarGrootKoerkamp:
LGTM! Thanks :) While we may decide not use this in BAPC/NWERC for consistency reasons, supporting this makes sense, and this approach of storing the data in a variable seems like the most obvious approach :+1:

PR #261, opened 2023-06-19
Hash tests

mzuenni:
Cache generator invocations by hash to allow renumbering testcases. Behaviour changes:

- testcases are stored in `tmp/data/<hash>` instead of `tmp/data/name`
- warning for unlisted testcases in `data/`
- removed `--move-manual` and `--add-manual`
- added new `--add-manual` which addes all testcases in a directory

TODO:

- [ ] move unlisted cases to `/data/unlisted` instead of warning (decrease warnings when renumbering cases)
- [ ] move cache to `generators/.cache/<hash>` instead? (tmp gets deleted on restart...)


bin/program.py
```diff
@@ -403,20 +407,21 @@ def build(self, bar):
         if self.tmpdir.is_file():
             self.tmpdir.unlink()
         self.tmpdir.mkdir(parents=True, exist_ok=True)
-        self.timestamp = 0
         self.input_files = []
+        hashes = []
         for f in self.source_files:
```
RagnarGrootKoerkamp, line 415: maybe ensure this is sorted so that `combine_hashes` is deterministic

bin/generate.py
```diff
@@ -545,16 +552,13 @@ def up_to_date():
                 last_change = max(last_change, (problem.path / t.source).stat().st_mtime)
                 t.cache_data['source'] = str(t.source)
             else:
-                if t.generator.program is not None:
-                    t.cache_data['generator_hash'] = t.generator.program.hash
+                t.cache_data['generator_hash'] = t.generator.hash(seed=t.seed)
                 t.cache_data['generator'] = t.generator.cache_command(seed=t.seed)
```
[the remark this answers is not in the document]
mzuenni: i kept them since it is the only way for a human to find out to which generator the cached file belongs, might be useful for some debugging

bin/program.py
mzuenni: combine_hashes sorts the given list

bin/program.py
```diff
@@ -95,6 +96,9 @@ def __init__(
         # Make sure we never try to build the same program twice. That'd be stupid.
         if not skip_double_build_warning:
             if path in problem._programs:
+                print('aaaa', file=sys.stderr)
```
RagnarGrootKoerkamp, line 99: revert?

mzuenni:
Is that a new bug? Anyway, I am unsure if you should be allowed to include anything outside the `generators/` directory... or at least you should not include anything in `data/`. A generator is not allowed to depend on another testcase.

RagnarGrootKoerkamp:
@mzuenni `include` can be used for test-data-groups to include some testcases in a group, so that a group `easy` can become a subset of group `hard`. 

I think this is also failing on master, so I think it's cleaner to fix this in a separate PR -- this one is large enough already

mzuenni:
oh, but that forces some kind of order on the generators ): Right now the cases are not generated in any specific order? i.e. whatever the scheduler decides happens ^^'

RagnarGrootKoerkamp:
I'm going to merge this since it seems to work well on my machine. Possibly notes/warnings could be slightly more optimized but that will be easy afterwards once we use it more.

Renumbering testcases is much nicer now for sure, and it's also more stricts about eg samples not being mentioned in the yaml, which is good IMO.
The rewritten `add-manual` command also works well.

Thanks again :) This solves quite some outstanding issues.

PR #263, opened 2023-06-21
Small fixes

mpsijm:
A few small fixes that I had not yet upstreamed. Commits should be reviewable one-by-one. Enjoy :smile:

Note that some of these fixes are cherry-picked from #251, but I won't have time to revisit that PR yet :innocent:


RagnarGrootKoerkamp:
thanks :) small fixes indeed :P

PR #262, opened 2023-06-21
generator with `include` can only run once

RagnarGrootKoerkamp:
Here is a simple generator, say for an `addtwonumbers`-like problem:
```yaml
data:
  sample:
    type: directory
    data:
      '1': stdout.py 1 2

  secret:
      type: directory
      data:
        'foo': stdout.py 4 6
      include:
        - 'sample/1'
```
It uses the `include` key to include the `1 2` instance called `1` also in the secret data.

This works once:
```sh
> bt generate
PROBLEM generatorincludes
[…]
Generate: sample/1   NEW 1.in
Generate: sample/1   NEW 1.ans
Generate: secret/foo NEW foo.in
Generate: secret/foo NEW foo.ans
> ls data/secret
1.ans   1.in    foo.ans foo.in
```

but not twice
```sh
> bt generate
PROBLEM generatorincludes
Traceback (most recent call last):
[...]
  File "/Users/thore/Developer/BAPCtools/bin/generate.py", line 955, in parse
    assert t.path not in self.known_cases
```


RagnarGrootKoerkamp:
@thorehusfeldt Ok I think this is in pretty good shape now. I've added caching of validators flags so they should only rerun for includes if arguments changed.

Could you test it out? After generating once rerunning it should be pretty much instant provided there are no errors. Then if you modify the flags in a `testgroup.yaml` that validator should rerun.

One open question is whether I should by default skip running on included cases. Since I (and kattis) don't support having distinct output validator flags, running once should be OK right.

RagnarGrootKoerkamp:
It seems there are some parallellism issues now: we must make sure testcases have finished processing before we can include them elsewhere. I'll probably rewrite things into two passes after all:
- first pass in parallel to generate testcases
- second pass sequential to generate includes. (Not in parellel, because includes can also depend on each other. Waiting for sequential validation should be rare and fast anyway.)

mzuenni:
Generation is already done in two steps (first all listen then all unlisted) a third step won't harm :D

thorehusfeldt:
Alas, I have no success `bt generate`-ing the example problem `generatorincludes` (which is zipped above.)

It just gets stuck.

```
generatorincludes git:(pull_262) > bt generate
PROBLEM generatorincludes
^CFATAL ERROR: Running interrupted         [############------------------------------------------------------------------- 4/20]
```

I’m on 9a0d821, and on a 2017 Mac, Python 3.10.

RagnarGrootKoerkamp:
Can you run with `-j0`? Runtime errors usually make it hang when parallelization is on.

thorehusfeldt:
(to success using `-j0` moved to slack)

In better news, I have rewritten https://dpop23.kattis.com/contests/dpop-23-1/problems/dpop23.storebededag to use `generators.yaml`, and it works.

thorehusfeldt:
An interesting issue is the following:

If, in the running example, we use _numbered_ testcasenames in `secret/small/positive`
```yaml
            data:
              - "sm-all-pos": stdout.py 4 6
              - "sm-zero": stdout.py 4 0
```
what is then the proper value in the later include in `secret/small/generate`:
```yaml
            include:
              - "sample/2"
              - "sm-zero"
```
or
```yaml
            include:
              - "sample/2"
              - "2-sm-zero"
```
Currently, the latter work, and the former predictably throws
```sh
  File "/Users/thore/Developer/BAPCtools/bin/util.py", line 467, in read_yaml
    assert path.is_file()
AssertionError
```

I don’t pretend to have thought this through, but I think I prefer `sm-zero` (otherwise the yaml-file author needs to count and possibly even rename `sm-zero` when the other testgroup changes.) Allowing _both_ would also be option.

RagnarGrootKoerkamp:
Re. numbered testcases:
- `include:` references should never include numbers, since they are meant to be unstable.
- Numbered cases can (and at least for me often do) have identical names, and so they can have non-unique names when referring to them.
- Anything included into a numbered directory should be renumbered. Probably by first numbering included cases from 1, and then continuing with generated cases.
- But renumbering cases introduces non-unique names for them, which sounds problematic given that kattis only runs once for each unique name. (Or do they dedup by symlink target instead of name?) Either way having non-unique names (including number) doesn't sound ideal to me either.

So for now the simplest solution to me sounds like forbidding using `include:` inside numbered directories, and also forbidding `include:` to point to numbered directories.
Since the convention is already that all testcases should have unique names, this doesn't sound like a big drawback. (One benefit of numbered directories is that they allow being lazy and reusing names, but you already shouldn't do that it seems.)

A completely different approach would be to have a *global* numbering instead of per-directory numbering. Then, we could include cases and reuse their existing number. Still the issue remains that testcases may have non-unique names.

TL;DR: Unless you would really like to use it with numbering, let's only allow `include:` in named (non-numbered) directories and make explicit errors when used in combination with numbering.

thorehusfeldt:
> Since the convention is already that all testcases should have unique names

I agree with that observation and also like the convention very much, and would be very happy to move the specification in that direction more explicitly. 

Before going there I just want to make sure we aren’t talking past each other about what “unique names” means: “name” here means that the name of testcase whose indata is at `secret/hard/foo.in` is `foo`. And unique means “appears exactly once as a testcase name in the _problem_” (as opposed to “the _testgroup_”). Right?

thorehusfeldt:
From my own (limited) experience:

1. the ability to impose a total order on the testcases (so that judging happens in a specified way) is crucial; the tradition of numbering them by order of appearance in the `generator.yaml` is a sufficiently pleasant way of supporting this. It is used in `testdata_tools`. It should have _high_ priority to support this also for generators specification, _in particular_ for problems with testgroups (which would want to use `include` a lot.)
2. I am happy with the restriction that testcase names be unique; it is satisfied by every problem I have looked at (and almost-satisfied by definition for problems that use only the testgroups `secret` and `sample`). It’s consistent with the Kattis user interface (which identifies testcases by name, not by path.)
3. `generators.yaml` should be easy to read, easy to write, and pleasant to edit.

Thus, I prefer the following:
1. the syntax should be 

  ```yaml
  include: testcase_key | testgroup_path
  ```
  Here, `testcase_key` is something like `lg_connected_small_deg_12` and `testgroup_path` is something like `secret/edgecases`. Note the distinction between a `key` (which is what the `testcase_rule` is identified with in `generators.yaml`) and the _name_ of the testcase (which may end up as `042-lg_conneced_small_deg_12`).

2. a testcase gets an automatic number at most once (namely when its rule is specified in a list); it is included elsewhere _with that number_.

I believe that these are the rule, _mutatis mutandis_, that are implicit in `testdata_tools`, they have been tested for hundreds of problems over the years and work well. 

In particular, the above decisions would lead to the much more pleasant
```yaml
include:
  - 'ragnars_case_that_kills_thores_greedy'
  #  WRONG: 'secret/greed/ragnars_case_that_kills_thores_greedy', use key (not path)
  - 'foo'
  # WRONG: '017-foo', do refer to testcases by key (not by name)
```

**Equivalence broken**

Note that the suggestion breaks the current `generators.yaml` specification in the sense that `The above data: list is equivalent to the map:` in line 201 is now no longer true. (Because the _key_  `01-a` can be `include` -ed.)

**Allowing nonunique names**

In principle, nothing prevents us from implementing the above and still allow non-unique testcase names. The syntax could be 

```yaml:
include: testcase_key | testcase_path | testgroup_path
```

and the tool should complain when a `testcase_key` is not unique. I don’t think this is worth doing, but there may be problem-setting traditions I don’t know about.

**Per-problem-numbering?**

I am somewhat indifferent to whether automatic numbering should be per problem or per testgroup. (`testdata_tools` uses “per problem”, and always uses 3 digits.) I mildly think that “per-problem” avoids some weird hypothetical situations (where the judgement order in `secret/foo` changes because an included testcase from `secret/bar` received a different number) , is easier to specify, and less surprising.

RagnarGrootKoerkamp:
> I am happy with the restriction that testcase names be unique

I explicitly do not want to enforce this globally, since (in BAPC/NWERC) we often have repeated testdata. E.g. like this for one of last years NWERC problems:

```
      - min: stdout.py 1 1 1 2
      - max: stdout.py 1000 1000 999 1000
      - max_diff: stdout.py 1000 1000 1 1000
      - max_diff: stdout.py 1 1000 1 1000
      - close: stdout.py 1000 1000 482 483
      - close: stdout.py 1 1000 482 483
      - close: stdout.py 1 1000 1 2
      - close: stdout.py 1000 1000 1 2
      - close: stdout.py 961 1000 979 983
      - close: stdout.py 997 1000 977 983

      - manual: stdout.py 1 928 1 66
      - manual: stdout.py 1 928 2 88
      - manual: stdout.py 1 928 3 74
      - manual: stdout.py 1 928 3 41
      - manual: stdout.py 1 928 1 12
      - manual: stdout.py 2 928 89 93
      - manual: stdout.py 7 928 51 72

      - random: rand.py {seed:1}
      - random: rand.py {seed:2}
      - random: rand.py {seed:3}
      - random: rand.py {seed:4}
      - random: rand.py {seed:5}
      - random: rand.py {seed:6}
      - random: rand.py {seed:7}
      - random: rand.py {seed:8}
      - random: rand.py {seed:9}
```

- From the above: we cannot enforce globally unique keys. Then what does `include: non-unique-key` mean? It could either crash, or include all cases with that key. In fact, the latter sounds quite reasonable to me. 
- We only need `include: testcase_path` if a key is generated is distinct testgroups. In fact, while I think `include: random` would be OK to include multiple cases for the example above, I think this may work out confusing if `random:` was used as a key in multiple testgroups. But restricting this may be artificial and cause confusion as well.
- Having a single (re-used) number per testcase sound good to me.
- I think that should imply global numbering. With local numbering you could en up with `1-x`, `1-y`, and `1-z` in a given directory, where each is generated first in their own testgroup and then included.
- Global (indeed, per *problem*) numbering naturally ensures that included cases before generated cases.
- Broken equivalence: yes I think it's OK to make `01-a: ..` and `- a: ..` to be slightly distinct in the way they need to be referred to.

mzuenni:
Just some notes:

- referring to keys with numbers is also weird since the number of leading zeros depends on the number of other test cases and may even change
- We may want to support some glob pattern-like style? If yes we could do something like this:
  - `path/ends/with/` to include specific paths
  - `keys/end/without` to include keys with a unique name (else error)
  - `if/name/is/not/unique*` to include all files with that name

thorehusfeldt:
Thank you for the examples! My misunderstanding is based on the following line 126 in `generators.yaml`:

```yaml
 # No key (testcase or testgroup) may be a prefix of another key.
```

So “key” is not what we mean here.

RagnarGrootKoerkamp:
I'm not entirely sure we currently enforce that, but the idea of that rule is to guarantee that files belonging to distinct testcases don't collide. This way you can generate any file starting with the given prefix without worrying it would be used for another testcase. (I.e. you could generate `a-2.in` for testcase `a` and it will still be OK since testcase `a-2` cannot exist.)

So anyway `key` in that sentence includes optional numbering, but as used in this issue we've used it for `name as specified in generators.yaml`.

thorehusfeldt:
(In any case, the `random` `random` example belongs to `generators.yaml`; it’s a useful feature, easy to misunderstand (I did!), different from what happens in lines 120+,  and interestingly different from a design choice behind `testdata_tools` does.)

thorehusfeldt:
Huh. I only now realise that the specification even allows for auto-numbered _testgroups_. So I can do

```yaml
data:
  sample: ...
  secret:
    type: directory
    data:
    - paths:
        type: directory
    - trees: ...
    - connected: ...
    - general: ...
```
and rely on the automatic numbering to ensure they become group 1, 2, etc. Neat. (And I indeed want to `include: secret/paths` (not `secret/1-paths`) in `secret/trees`.

RagnarGrootKoerkamp:
Right yes; once you number a directory you'd also want to number the groups right? to ensure they are run in the right order.

Other random remarks while we're at it:
- This `type: directory` is kinda ugly. It would be nice to avoid it. So far we decided to make parsing simple and not do too much analysis on the type of `data` and other keys available in the testgroup, but auto-inferring it in some deterministic way would be nicer.
- I mentioned this somewhere recently: possibly we could do `testcase.in: 1 2` to directly encode file contents, so that `stdout.py` isn't needed anymore. Then you can also do `testcase.ans: 3` alongside it. I'd say that when some `testcase.*` is present as key, `testcase` itself can not be used anymore.

thorehusfeldt:
Re the random remarks:

* I’m happy to get rid of `type: directory`.
* I would love the `stdout`-syntax. (Elsewhere you propose `< 1 2` which also makes sense to me.)

thorehusfeldt:
Random observations from transforming a few scoring problems (that use `include`) from `testdata_tools` to `generators.yaml`. (It works quite well so far.)

**1. type: directory at the root** 
 
Apparently, in the current implementation the root directory is the only directly that doesn’t require `type: directory`. Incidentally, the root is the _only_ level where I could see this key play a useful role. For instance, the value could be `scoring`, which could make a hypothetical YAML schema validator (such as cue) perform a lot of sanity checks.

**2. Testdata.yaml checks for test groups**

When writing a scoring problem, one needs to specify, at the `secret/group_i` level:

```yaml
testdata.yaml:
  accept_score: 21
  range: 0 21
  grader_flags: min
```
and at `secret`:
```yaml
on_reject: continue
range: 0 100
grader_flags: first_error accept_if_any_accepted
```
and at the root:
```yaml
on_reject: continue
range: 0 100
grader_flags: ignore_sample
```
That’s quite error-prone if done by hand. (And the only actually interesting value is `21`, everything else is always the same for a scoring problem with range `0 100`.) `testdata_settings` writes these `testdata.yaml` values automatically. 

In principle, the generators spec could actually enforce this for scoring problems, or `bt generate` could sanity-check, or (ideally) a schema-aware editor would infer those values automatically. But the `yaml`-file itself would need some way of saying “I specify a scoring problem with test groups that sum to 100”. I’m not sure how to do this.

**3. Glob musings**

The fact that cases and groups get automatic numbers in a systematic fashion is useful, but leads to difficulties for including them. For instance, if I use 5 test groups specified with the automatic numbering logic (so their keys are all `group` and their names become `secret/1-group`, `secret/2-group`, `secret/3-group`, `secret/4-group`, `secret/5-group`) then it’s difficult to `include` the second group in the third (which is the _main_ use-case for include.)

I’d need to do refer to the group by a name that is not the yaml:
```yaml
include: secret/2-group
```
Alternatively, if I give the groups meaningful names like `paths`, `trees`, etc. they get autonumbered to `1-paths`, `2-trees`, etc. and I can write
```yaml
include: secret/trees
```
which is weird because no file is actually named `secret/trees`. Is this the place for

```yaml
include: secret/*trees
```

Dunno.

RagnarGrootKoerkamp:
* I'll update the implementation to not require `type:directory` anywhere. For more syntax validation, I'd use a different key.
* Right, if you use the automatic numbering so that you don't have to explicitly number yourself anymore, that's problematic. In this case either use a non-numbered top-level, or else my preference would be `secret/group`. In this case I would error when `group` refers to multiple cases/groups, so I could see some value in adding `secret/group*` or so to allow multiple matches, but I would only add that if you think this is actually useful, and otherwise simply always error on multiple matches.

thorehusfeldt:
If others here are playing around with `generator.yaml`-files, here’s an imperfect JSON schema that does a bit of validation. It interestingly fails on empty yaml values in commands, such as this: 
```yaml
sample:
  '1': 
```
mainly because I don’t know how to handle `null` in JSON.

I can give this more attention when the specification converges; I’ve found it quite nice to get at least some help from the editor. (Editor support for CUE schemas doesn’t exist, otherwise that would be way more useful and pleasant.)
 
[schema.json.gz](https://github.com/RagnarGrootKoerkamp/BAPCtools/files/11881626/schema.json.gz)

RagnarGrootKoerkamp:
Oh that's super nice, we should definitely bundle that in BAPCtools, possibly together with the big example `generators.yaml` that's currently in the spec PR.

For emacs: I can start the Yaml LSP mode, and then there is a function `lsp-yaml-set-buffer-schema` where I can point to the file, and it indeed gives some basic completion.

We should also decide some prefix (ie `_`) for 'out-of-spec' keys, so that we can warn on other unrecognised keys

RagnarGrootKoerkamp:
Update (I'm just putting all my generators work in this PR now):
- Copied your json config into `support/generators_yaml_scheme.json`
- Updated `doc/generators.yaml` with latest version from spec PR.
- Removed the need for `type:` from code and examples and spec.
- Added restriction that each dictionary inside a numbered list has exactly one entry.

TODO:
- Add support for `testcase.in: 1 2 3` (will probably do that in a separate PR afterwards though)
- Implement including numbered cases. As it stands, I would go with `group/name` without numbers, asserting that `group` is unique and that `name` is unique in `group`.

thorehusfeldt:
After transmogrifying a dozen or so group-based scoring problems to use `generators.yaml` I would much prefer

```yaml
include: 
- group1
- funky_testcase
```

to 

```yaml
include:
- secret/group1
- sample/funky_testcase
```

To spell it out: `include`d files and groups are referred to by (yaml) _key_, not by _path_ or (God forbid) possibly-autonumbered- _name_. In particular, if you want to `include` such a file-or-group, its key has to be unique in the testdata for this problem, each `include` leads to exactly _one_ testcase or testgroup being `include`d.

`include`ing a file-or-group that several times as a key to `data` in the same `generators.yaml` file is an error.

This convention is 
1. easier to type, read, maintain, and search for in the same file
2. easier to specify and avoids the weird logic of referring to `secret/1-path/015-funky_case` as `secret/path/funky_case`.
3. no restriction in any problem I’ve seen (because all those names are unique anyway; anything else would have been way too confusing, in particular for names that should be `include`d.)
4. no restriction in general (because any path you want to include somewhere else you can just call `random-to-include` instead of `random`, which minimises surpises)
5. presumably avoids a ton of implementation issues, possible misunderstandings, and nasty-to-find-errors during problem development

`testdata_tools` does the  same (but there it just follows from the convention there that all testcases have script-unique names).

RagnarGrootKoerkamp:
Ok, then that's what we'll go with. I completely agree that this avoids a bunch of edge cases that we otherwise need to handle in some way that is always going to be at least slightly unintuitive.

So internally I will build a dictionary `key => list of testcase/testgroup`. For each include I do a lookup and assert that the list has exactly size 1.

Once edge case is if a key is defined exactly once before an include, but then defined for the second time later on. That should also be an error, but would pass the simplest implementation of the above idea.

thorehusfeldt:
One issue I thought of for automatic numbering:

Yaml specifies absence of ordering among dictionary keys; formally
```yaml
data:
  secret:
    group1:
      - tc-a
      - tc-b
      - tc-c
    group2:
      - tc-e
      - tc-f
      - tc-g
```
specifies the same structure as if the keys `group1` and `group2` had been reversed. This makes the auto-number of `tc-e` sometimes `4-tc-e` and sometimes `1-tc-e`, depending on how the yaml parser decides to relay the dictionary keys to the tool.

Maybe I’m missing something, but I think this can lead to confusion and in particular nondeterministic generators. (This phenomenon does not occur with your original rule to number per-testgroup; I apologise for not thinking this through.)

Fixes I can think of:

1. go back to numbering-per-testgroup
2. define numbering in terms of lexicographic order of testgroup paths. This makes the result deterministic and re-establishes that “group1 precedes group2”. In 99% of the cases, the author would never even realise there is a problem.
3. insist that auto-numbered testcase (i.e., testcases defined as list entries rather than dictionary keys) are successors of auto-numbered testgroups (i.e., testgroups defined as list entries rather than dictionary keys.)

RagnarGrootKoerkamp:
this is getting a bit long here; let's move the testcase numbering to #274.

RagnarGrootKoerkamp:
* I've implemented including by `key` and cleaned up the commit history a bit.
* Non-numbered cases are processed lexicographically again.
* This should be working now for non-numbered cases.
* Numbered cases can be referred to by their `key` (without number) as well. They'll be included in the target directory with the original number.
* Mixing the inclusion of numbered and non-numbered cases is OK.

Numbering of testcases itself I'll fix in a separate PR after merging this.

let me know whether you think this can be merged now.

thorehusfeldt:
generators.yaml:
```yaml
accept_score: 25
```
should be
```yaml
accept_score: "25"
```

(The joys of actually validating against a schema. `25` looks like a number to yaml. But the spec requires string here.)

RagnarGrootKoerkamp:
Oh ok, sure. (I assume the testdata.yaml spec is somewhat well established, and changing it to a number is too late now? And this was not simply an oversight?)

RagnarGrootKoerkamp:
ok I think it's time to merge this; some other changes depend on it, and it seems to work so far

PR #267, opened 2023-06-22
Support FreeBSD & aquaBSD

obiwac:
Replaces use of `is_mac()` with `is_bsd()` (which is a superset of `is_mac()`), as we don't want to limit stack/virtual memory usage on both systems.


RagnarGrootKoerkamp:
thanks :+1:

PR #268, opened 2023-06-26
reimplemented parallel

mzuenni:
The current implementation has `q.get()` operations with unmatched `q.task_done()` operations which leads to infinite waits if `q.join()` is called.

- in line 36, we break without 'q.task_done()'
- in line 39, we don't call  'q.task_done()' if an exception occurred 
- in line 91, we clear the queue without 'q.task_done()'

but even if these are added we have the variables `self.stopping` and `self.error` on which race conditions can occur. The `self.error` variable is no problem but a race condition on `self.stopping` in line 82 could lead to `2*num_threads` `None`s getting pushed in the queue. Which would break `q.join()` again.

Therefore, I propose a new implementation with a lock to prevent such race conditions (This lock also allows to do all other synchronization required).

### DIFF
--- a/bin/parallel.py
+++ b/bin/parallel.py
@@ -40,7 +40,7 @@ class Parallel:
         while True:
             with self.mutex:
                 # if self.abort we need no item in the queue and can stop
-                # if self.finifh we may need to wake up if all tasks were completed earlier
+                # if self.finish we may need to wake up if all tasks were completed earlier
                 # else we need an item to handle
                 self.todo.wait_for(lambda: len(self.tasks) > 0 or self.abort or self.finish)
                 
@@ -54,8 +54,8 @@ class Parallel:
                     # get item from queue (update self.missing after the task is done)
                     task = self.tasks.pop(0)
             
-            # call f and catch all exeption occuring in f
-            # store the first expetion for later
+            # call f and catch all exceptions occurring in f
+            # store the first exception for later
             try:
                 current_error = None
                 self.f(task)
@@ -82,7 +82,7 @@ class Parallel:
 
         with self.mutex:
             # no task should be added after .done() was called
-            assert(not self.finish)
+            assert not self.finish
             # no task will be handled after self.abort anyway so 
             # we can skip adding
             if not self.abort:
--- a/bin/parallel.py
+++ b/bin/parallel.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-import queue
 import threading
 import signal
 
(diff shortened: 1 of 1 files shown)


bin/parallel.py
```diff
@@ -29,67 +34,82 @@ def __init__(self, f, num_threads=True):
...
-        try:
-            while not self.stopping:
-                task = self.q.get()
-                if task is None:
+        while True:
+            with self.mutex:
+                if len(self.tasks) == 0:
+                    self.todo.wait_for(lambda: len(self.tasks) > 0 or self.abort or self.finish)
+                
+                if self.abort or len(self.tasks) == 0:
```
RagnarGrootKoerkamp, line 42: replace `len == 0` by `self.finish` for consistency with the `or`s above

bin/parallel.py
```diff
@@ -29,67 +34,82 @@ def __init__(self, f, num_threads=True):
...
+        if not self.num_threads:
+            self.f(task)
             return
 
-        if self.num_threads:
-            self.q.put(task)
-        else:
-            self.f(task)
+        with self.mutex:
+            assert(not self.finish)
```
RagnarGrootKoerkamp, line 71: remove ()

bin/parallel.py
```diff
@@ -29,67 +34,82 @@ def __init__(self, f, num_threads=True):
...
-        if self.error:
-            raise self.error
-        self.q.join()
-        if self.error:
-            raise self.error
+        if not self.num_threads:
+            return
+
+        with self.all_done:
+            if self.missing > 0:
```
RagnarGrootKoerkamp, line 82: could remove the `if` and do the next line unconditionally

bin/parallel.py
```diff
@@ -29,67 +34,82 @@ def __init__(self, f, num_threads=True):
...
+                raise self.first_error
 
     # Wait for all tasks to be done and stop all threads
     def done(self):
         if not self.num_threads:
             return
 
-        for _ in range(self.num_threads):
-            self.q.put(None)
+        with self.todo:
```
RagnarGrootKoerkamp, line 92: is this different from `with self.mutex`?

RagnarGrootKoerkamp (commented):
Looks pretty understandable. Just some extra comments for clarify would be nice.

bin/parallel.py
```diff
@@ -36,14 +39,23 @@ def __init__(self, f, num_threads=True):
...
+                    break
+                elif self.finish and len(self.tasks) == 0:
+                    # on finish we can only stop after the queue runs empty
                     break
                 else:
+                    # get item from queue (update self.missing after the task is done)
                     task = self.tasks.pop(0)
             
+            # call f and catch all exeption occuring in f
+            # store the first expetion for later
```
RagnarGrootKoerkamp, line 58: typo

bin/parallel.py
mzuenni: we don't want to directly leave on `self.finish`, we still want to handle all items in the queue first but i made that more obvious by changing the code :)

PR #269, opened 2023-06-27
Interaction

mzuenni:
Allow judging interactive programs in parallel by using `wait4` instead of `wait3`

### DIFF
--- a/bin/interactive.py
+++ b/bin/interactive.py
@@ -173,6 +173,37 @@ def run_interactive_testcase(
         val_in = team_log_in
         team_in = val_log_in
 
+    # Use manual pipes with a large buffer instead of subprocess.PIPE for validator and team output.
+    if validator_error is False:
+        validator_error_in, validator_error_out = mkpipe()
+    else:
+        validator_error_in, validator_error_out = None, validator_error
+    if team_error is False:
+        team_error_in, team_error_out = mkpipe()
+    else:
+        team_error_in, team_error_out = None, team_error
+
+    validator = subprocess.Popen(
+        validator_command,
+        stdin=val_in,
+        stdout=val_out,
+        # TODO: Make a flag to pass validator error directly to terminal.
+        stderr=validator_error_out,
+        cwd=validator_dir,
+        preexec_fn=limit_setter(validator_command, validator_timeout, None, 0),
+    )
+    validator_pid = validator.pid
+
+    submission = subprocess.Popen(
+        submission_command,
+        stdin=team_in,
+        stdout=team_out,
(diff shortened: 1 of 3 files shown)


bin/interactive.py
```diff
@@ -173,6 +173,38 @@ def mkpipe():
...
+    validator = subprocess.Popen(
+        validator_command,
+        stdin=val_in,
+        stdout=val_out,
+        # TODO: Make a flag to pass validator error directly to terminal.
+        stderr=validator_error_out,
+        cwd=validator_dir,
+        preexec_fn=limit_setter(validator_command, validator_timeout, None, 0),
+    )
+    validator_pid = validator.pid
```
RagnarGrootKoerkamp, line 195: you don't actually use/need this one it seems

bin/interactive.py
```diff
@@ -194,82 +226,29 @@ def mkpipe():
...
-
     # Wait for first to finish
     left = 4 if interaction else 2
     first_done = True
     while left > 0:
-        if kill_submission:
-            # Kill the submission asynchronously.
-            signal.setitimer(signal.ITIMER_REAL, 0.001, 0)
-        pid, status, rusage = os.wait3(0)
+        pid, status, rusage = os.wait4(-gid, 0)
```
RagnarGrootKoerkamp, line 251: add some comment here or on `gid` about how this grouping and waiting works

bin/interactive.py
```diff
@@ -305,15 +282,16 @@ def kill_submission_handler(signal, frame):
...
-                first_done = False
-                continue
+                pass
+            else:
+                assert False
+        else:
+            assert False
+
+        left -= 1
+        first_done = False
```
RagnarGrootKoerkamp, line 294: hrmmm, I think `first_done` should only become false for submission and validator finishing. If the validator closes stdout, that should kill one of the `tee`s, but by that time the validator can still finish before the submission after which we want to kill the submission. Maybe we should just introduce separate `done` indicators for each of the non-validator processes and kill the ones that are not done yet?

RagnarGrootKoerkamp (commented):
this looks very nice! Just some small comments

bin/interactive.py
mzuenni: a tee should never end first but i changed it anyway. We don't need extra state, we can kill groups :D

RagnarGrootKoerkamp:
I also tested some problems and it seems to work fine.
There is some slowdown of timings, but this is expected also for non-interactive problems anyway.
Let's merge and rebase #251 on top of it.

PR #275, opened 2023-06-28
[parallel] Fix #270: Set minimum number of jobs to 1

### LINKED ISSUE #270: New parallelism breaks lazy judging
We want to stop judging as soon as we hit a `TLE (aborted)` testcase.
This broke in #268 (27daebadea8a0a32ed032f72c08d0dcbe9dd6b53).

cc @mzuenni @mpsijm

mpsijm:
Funnily enough, setting `--jobs=0` _does_ correctly start a single job, but it _doesn't_ correctly abort the submission when lazily judging. 😂

Note that I didn't spend much time looking into how this code works, but I did note that `self.num_threads` is now always "truthy" when `num_threads is True`. There still exists cases where `num_threads` is "falsy" (i.e., `None`/`False`/`0`), causing `self.num_threads` to be "falsy" as well. These cases are `bin/fuzz.py` and `bin/generate.py`, when parallelization is disabled when the problem is interactive. Stuff may still keep running in this case when there is an early abort, but I didn't test this. (perhaps we can even enable parallelization now, since #269?)


mzuenni:
I prefer to fix the queue :D. I will also check if we can enable more parallelism ^^

RagnarGrootKoerkamp:
Superseeded by #276

PR #276, opened 2023-06-28
fix parallel if num_threads is false

### LINKED ISSUE #270: New parallelism breaks lazy judging

mzuenni:
Fix #270 correctly

### DIFF
--- a/bin/parallel.py
+++ b/bin/parallel.py
@@ -85,7 +85,8 @@ class Parallel:
             # no task will be handled after self.abort
             if not self.abort:
                 self.f(task)
-                return
+            
+            return
 
         with self.mutex:
             # no task should be added after .done() was called
--- a/bin/parallel.py
+++ b/bin/parallel.py
@@ -21,9 +21,12 @@ class Parallel:
         # condition used to notify join that the queue is empty
         self.all_done = threading.Condition(self.mutex)
 
+        # only used in parallel mode
         self.first_error = None
         self.tasks = []
         self.missing = 0
+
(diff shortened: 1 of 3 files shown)


bin/parallel.py
```diff
@@ -77,14 +80,17 @@ def _interrupt_handler(self, sig, frame):
...
     def put(self, task):
         if not self.num_threads:
-            self.f(task)
-            return
+            # no task should be added after .done() was called
+            assert not self.finish
+            # no task will be handled after self.abort
+            if not self.abort:
+                self.f(task)
+                return
```
RagnarGrootKoerkamp, line 88: Move this return one level up? With `self.abort == True` we also need to return here.

bin/parallel.py
mpsijm, line 1: Note that the `num_threads` parameter is no longer used now (at least, according to `grep -R 'Parallel(' bin`), so it can be removed :slightly_smiling_face:

mpsijm:
Works on my machine as well, thanks for fixing this @mzuenni! :smile:

bin/parallel.py
RagnarGrootKoerkamp: oh well it's OK to keep I'd say

mpsijm: All right, then it's fine in that case :smile:

PR #277, opened 2023-06-28
add warning

### LINKED ISSUE #210: `bt pdf` does not warn when there are no interaction samples
`bt pdf` warns when there are no `.in` or `.ans` files for "regular" problems, but for interactive problems, it does not warn if there are no `.interaction` samples.

mzuenni:
Add a warning if `.interaction` is missing (fix #210 )

### DIFF
--- a/bin/problem.py
+++ b/bin/problem.py
@@ -177,18 +177,17 @@ class Problem:
             t = run.Testcase(p, f)
             # Require both in and ans files
             if needinteraction and not t.in_path.with_suffix('.interaction').is_file():
-                if not t.bad_input:
-                    warn(f'Found input file {str(f)} without a .interaction file. Skipping.')
+                assert only_sample
+                warn(f'Found input file {f} without a .interaction file. Skipping.')
                 continue
             if needans and not t.ans_path.is_file():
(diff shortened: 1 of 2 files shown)


bin/problem.py
```diff
@@ -176,6 +176,10 @@ def maybe_copy(x):
         for f in in_paths:
             t = run.Testcase(p, f)
             # Require both in and ans files
+            if needinteraction and not t.in_path.with_suffix('.interaction').is_file():
+                if not t.bad_input:
+                    warn(f'Found input file {str(f)} without a .interaction file. Skipping.')
```
RagnarGrootKoerkamp, line 181: - I don't think `str(f)` is needed? just `f` should work?
- maybe assert that `needinteraction` is only set if `only_sample` is `True`, and then drop the `if not t.bad_input`?

bin/problem.py
```diff
@@ -185,9 +189,11 @@ def maybe_copy(x):
 
         if len(testcases) == 0:
             # For interactive problems, allow 0 samples.
```
RagnarGrootKoerkamp, line 191: remove the comment

PR #255, opened 2023-06-28
Explain output validation

RagnarGrootKoerkamp:
(Original issue by @thorehusfeldt converted into PR.)

BAPCtools performs two different modes of output validation, as explained in the source code comment of `class OutputValidator(Validator)` in `bin/validate.py`.

This is both useful and important to understand for problem authors (in particular because `verifyproblems` does not do this, and the specification does not mention this either) but it is not documented, as far as I can tell.

(And I am honestly unsure about whether the default output validator (with flags) is run for `bt validate -o`, or in which order the validators are run – for instance, I think `output_validator/correctness.py` cannot assume that `ouput_validator/check_grammar.py` is run first, so it might `JE` if it expects to be fed well-formed team output.)

### DIFF
--- a/bin/problem.py
+++ b/bin/problem.py
@@ -1,5 +1,4 @@
 import re
-import glob
 import argparse
 import hashlib
 import random
--- a/bin/validate.py
+++ b/bin/validate.py
@@ -4,7 +4,6 @@ from util import *
 
 
 class Validator(program.Program):
-
     # NOTE: This only works for checktestdata and Viva validators.
     FORMAT_VALIDATOR_LANGUAGES = ['checktestdata', 'viva']
 
@@ -83,17 +82,17 @@ class InputValidator(Validator):
         if self.language in Validator.FORMAT_VALIDATOR_LANGUAGES:
             return Validator._run_format_validator(self, testcase, cwd)
 
-        run_command = self.run_command + ['case_sensitive', 'space_change_sensitive']
+        run_command = self.run_command.copy()
+
+        if args is not None:
+            assert isinstance(args, list)
+            run_command += args
 
         if constraints is not None:
             constraints_path = cwd / 'input_constraints_'
             if constraints_path.is_file():
                 constraints_path.unlink()
             run_command += ['--constraints_file', constraints_path]
-
-        if args is not None:
-            assert isinstance(args, list)
-            run_command += args
 
         with testcase.in_path.open() as in_file:
             ret = exec_command(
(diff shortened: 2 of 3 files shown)


RagnarGrootKoerkamp:
Hrmm, the [spec](https://www.kattis.com/problem-package-format/spec/problem_package_format#output-validators) is not explicit whether validators are run in lexicographic order, but I'm pretty sure that's always the case.

thorehusfeldt:
Thank you for describing BAPC validation. Another surprise for me as a newcomer was the fact that input validation _also_ supplies `case_sensitive` and `space_change_sensitive`. My BAPC-aligned old problems need to handle these explicitly (by ignoring them) to get them out of `argparse`’s expected syntax:
 
 Here’s an example input validator:
 
```python3
from argparse import ArgumentParser

sys.argv.remove('case_sensitive')
sys.argv.remove('space_change_sensitive')
parser = ArgumentParser("Validate input")
parser.add_argument("--min_n", type=int, default=3)
parser.add_argument("--max_n", type=int, default=5)
parser.add_argument("--max_l", type=int, default=10**12, help="max solution size")
args = parser.parse_args()
```

RagnarGrootKoerkamp:
Yes I saw that while writing this down, and already remove `case_sensitive` and `space_change_sensitive` for `input format validation`. I kept them for `output format validation` though, since I do think it's nice to be strict on answer file syntax, and otherwise there would be no other way to distinguish `output format validation` from custom `output validation`.

I realize now I should also mention that the `validator-flags: ` in `problem.yaml` are passed for `output validation`.

thorehusfeldt:
After another read-through, I feel that `validate.md` requires another round of editing. 

And issue that remains honestly unclear to me is **Output Format Validation**. Assume there exists `output_validators/ans-format.py` and the setting is `validation:default`. I understand that `ans-format.py` is run against all `*.ans` during generation, and in particular run with the arguments `case_sensitive` and `space_change_sensitive`. (By the way, this is a great feature.)

What I honestly don’t understand is: 
1. is `ans-format.py` _also_ run against team output during `bt run`. I assume the answer is yes (after all, it resides in `output_validators`, and further I assume it is run _before_ the default output validator because `a<d`. (However, the filename of the default output validator isn’t really specified, so this is a guess.)
2. the problem author is supposed to write `ans-format.py` so that when it is run _without_ the flags `case_sensitive` and `space_change_sensitive` then it basically just `exit(42)`s so as not to conflict with the lenient attitude of the default validator.

The other interpretation would be that then BAPC-tools sees `validation:default` then it runs _only_ the default output validator during `bt run` and also never exports `output_validators` to Kattis. I’m sure I can find out by inspecting the code, but my confusion here can help us write the documentation better.

PR #280, opened 2023-06-29
more generic validation

mzuenni:


bin/validate.py
```diff
@@ -180,3 +180,79 @@ def run(self, testcase, run=None, constraints=None, args=None):
...
+    for byte in bytes:
+        cur_whitespace = byte == ord(' ') or byte == ord('\n')
+        if last == ord(' ') and cur_whitespace:
+            return True
+        if last == ord('\n') and byte == ord(' '):
+            return True
+        last = byte
+    return False
+
+def generic_validation(validator_type, file, *, bar, sample_ans=False):
```
RagnarGrootKoerkamp, line 219: 1. Add some comment on the purpose of this function. 'generic' wasn't super obvious to me on first reading.
2. Comment that `--skip-testcase-sanity-checks` disables this.
3. Instead of using `input` and `output` for `validator_type`, and `sample_ans`, it may be more consistent to use `input_format`, `output_format`, and `output`.

bin/validate.py
```diff
@@ -180,3 +180,79 @@ def run(self, testcase, run=None, constraints=None, args=None):
                 cwd=run.feedbackdir,
                 timeout=config.get_timeout(),
             )
+
+# Testcase input is only allowed to contain newlines and printable characters
+def _is_invalid_input_byte(byte):
```
RagnarGrootKoerkamp, line 185: maybe change this and the next function from `_is*` to `_has*` and make it take the string itself.

mzuenni: i would prefer to not try to decode the bytes :)

PR #282, opened 2023-06-29
Disable some hello subs, add others

### LINKED ISSUE #281: hello RTE woes on Mac: memsize, sigfpe, sigsegv
Not sure this should be fixed, but neither of `hello/submissions/run_time_error/{test-memsize.cc, test-sigfpe.c, test-sigsegv.c}` behave as expected on my Mac, leading to `pytest` failing already on the first problem.

For `memsize`, I haven’t found a way to limit `ulimit -v` (the size of virtual memory), so the machine blissfully keeps allocating memory for 5 seconds and then `TLE`s. This would be fixable if I understood how to limit virtual memory on my machine.

For `sigfpe`, it turns out that the behaviour of division by 0 even on ints is _undefined_, and some googling shows me that `gcc` on a mac indeed fails to crash also for other people. Maybe this particular submission might be better in a language where `7/0` is _guaranteed_ to crash. (`test-sigfpe.java` crashes just fine.)

For `sigsegv`, my `gcc`  seems to do something clever also at runtime (the compiler already warns about the `-Warray-bounds` violation). No run time error is produced.

thorehusfeldt:
Here are some suggestions for a more versatile `hello`. The most controversial suggestion is to disable `test-memsize.cc`; some other problem in `tests` should include that submission. Fixes #281


test/problems/hello/submissions/run_time_error/test-memsize.cc.disabled
```diff
@@ -0,0 +1,11 @@
+#include <iostream>
```
RagnarGrootKoerkamp, line 1: Can you revert to the original contents? And add a comment as to why this was disabled?

Maybe also add a comment to the others with the reason for disabling.

RagnarGrootKoerkamp:
You can ignore the failure here -- `.ctd` files apparently don't work on some CI instances but whatever.

thorehusfeldt:
I’ve cleaned up and commented the disabling changes to `hello`. I’ve also checked in `hellounix`; the goal is to retain important submissions that fail predictably on contest systems so they can be added to `tests` (but can by disabled on other operating systems.)

I cannot `bt test --problem hellounix`; please check that it passes on your machine.

test/test_problems.py
```diff
@@ -7,9 +7,12 @@
 import tools
 import problem
 import config
+import util
 
 # Run `bt run` on these problems.
-PROBLEMS = ['hello', 'helloproblemtools', 'different', 'fltcmp', 'boolfind', 'guess']
+PROBLEMS = ['hello', 'helloproblemtools', 'different', 'fltcmp', 'boolfind', 'guess'] + [
+    'hellounix' if not util.is_mac() and not util.is_windows() else []
```
thorehusfeldt, line 14: Consider
```
…'fltcmp'] + (['hellounix', 'boolfind', 'guess'] if  not util.is_mac()  else [])
```

RagnarGrootKoerkamp: `guess` i'll fix differently using the less strict checking for distinct error types. I'm still not quite sure why `boolfind` crashes though

RagnarGrootKoerkamp: both boolfind (remove `-static`) and guess (more lenient check) should be fixed now. can you check?

RagnarGrootKoerkamp:
Can you confirm that `guess` works on this branch on mac?

thorehusfeldt:
I hereby report that `bt run` works as expected (including randomly selecting an AC submission for `solution`) for `guess` since `3b1cd414a`.

thorehusfeldt:
Can we replace `#!/usr/bin/sh` with `#!/bin/sh` in various generators in `identity`? Or does that break anything on your side?

RagnarGrootKoerkamp:
Hmm, can you do a quick search on this. For python I now always use /usr/bin/env python3, so maybe /use/bin/env sh also works. Never saw that before though. But /bin/sh is probably also fine

thorehusfeldt:
Yes, I did some googling for `#!/usr/bin/env sh`, which would make sense to me, but apparently is seldom observed in the wild.  `#!/bin/sh` seems to be the way to go. Best stackoverflows I  found, largely inconclusive:
*  https://unix.stackexchange.com/questions/569964/posix-shell-scripts-shebang-bin-sh-vs-usr-bin-env-sh-any-difference 
* https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang

RagnarGrootKoerkamp:
Ok, then /bin/sh it is

thorehusfeldt:
For `bt constraints` on `identity`, I think where is no hope for Mac OS because of this:
```
#ifdef use_source_location
#include <experimental/source_location>
```
which apparently is not (yet?) available, see discussion here:
* https://discourse.llvm.org/t/unable-to-access-source-location-by-std-or-experimental/67565/7

My verdict is that this is not worth fixing; I suggest to take it out of `test`-ing depending on OS.

There is no easy fix

RagnarGrootKoerkamp:
C++ really grew a bit too fast... sad that 3y later this stuff isn't mainstream yet.

In fact I modified the `validation.h` in the BAPCtools repo recently to simply use the provided variable name as the 'unique key' instead over which to take min/max instead of the source location, and this seems to work well so far. (Really the only reason I used it was to make a unique counter for each independent invocation of `v.read_integer()`, but we already do `v.read_integer("a", ..)` and `v.read_integer("b", ..)` anyway, so I think I can just remove the `source_location` dependency entirely. Will try tomorrow.

mzuenni:
Side note, that is also what I do in my headers... but for compatibility, we should keep the line number entry in the constraints file (my headers write `LocationNotSupported:<variable name>`).

RagnarGrootKoerkamp:
Yes, sgtm

RagnarGrootKoerkamp:
@thorehusfeldt `bt constraints` should also work now

mpsijm:
Sorry I'm a bit late to the party, but:

> Can we replace `#!/usr/bin/sh` with `#!/bin/sh` in various generators in `identity`? Or does that break anything on your side?

@thorehusfeldt can you check if `#!/usr/bin/env sh` works on Mac? Because I believe that's the preferred way to do shebangs, since `env` will look for the correct location of `sh` on your `$PATH`. That is, assuming that `/usr/bin/env` exists on Mac :stuck_out_tongue:

thorehusfeldt:
Sure, it absolutely works.

Issue #271, opened 2023-06-27
Generators: Add `testcase: count: X` for multiple testcase copies


RagnarGrootKoerkamp:
Currently we do
```yaml
- random: gen.py {seed:1}
- random: gen.py {seed:2}
- random: gen.py {seed:3}
```
which would be nice to replace with
```yaml
- random:
    input: gen.py {seed}
    count: 3
```

(And maybe `input:` should be renamed to `gen:`?)

mpsijm:
Duplicate of #135 and #162? (the fact that you've opened the same issue three times, probably means we should implement it :joy:)

Also relates to #272 (/ #220), for the naming of the `input:`/`gen:` key.

RagnarGrootKoerkamp:
Hmm; the problem with 'inline' numbering is that that wouldn't work if testcases aren't already numbered in the first place. For named cases you'd have to do `random-1`, `random-2`, or so.

thorehusfeldt:
This is of course super useful for other generators than random ones. 

Assume `trees.py --i` is a generator that produces all nonisomorphic trees on 7 vertices (indexed by `k`), I would love to do

```yaml
- small:
    input: trees.py --k {seed}
    count: {2,4,5-12, 14}
```

instead of
```yaml
- small trees.py --k 1
- small trees.py --k 2
- small trees.py --k 3
...
```

RagnarGrootKoerkamp:
Ohh, interesting idea.

**Some context**
Currently we allow `{seed}` and `{seed:<any kind of salt>}`, and the entire thing is replace by an integer hash of *the entire* generate command. (This way, `gen.py {seed} 1` and `gen.py {seed} 2` have distinct seeds, and `gen.py {seed:1}` and `gen.py {seed:2}` also have distinct seeds.
This also means that `{seed}` is *not* a small incremental int typically. It's supposed to be used has seed a random number generator, but I still think it's good if `{seed}` is always pseudo-random, never `1, 2, 3, ...`.

**Variables**
The above means that your example isn't the intended way of using `{seed}`. But I think we can do something cleaner:
```yaml
- testcase:
    input: gen.py -n {n}
    n: [1,2,3, 10, 100, 1000]
```
Note that the value of `n` here is simply a list, and we generate one testcase for each element of the list.

Or maybe we should separate the variable `n` into a separate scope? (I don't think it's really needed though.)
```yaml
- testcase:
    input: gen.py -n {n}
    vars:
[...]

PR #272, opened 2023-06-30
Generators: Allow hardcoding testcases input/answer

RagnarGrootKoerkamp:
Currently we do
```
testcase: stdout.py 1 2 3
```
Would be nice to instead do
```
testcase:
  in: 1 2 3
```
Then, we can directly include samples and other small manual cases. This should support keys for some more extensions (corresponding to text files):
```
testcase:
  in: 1 2 3
  ans: 6
  hint: what is the sum of 1, 2, and 3?
  desc: sum of 3
```
(images probably shouldn't be made this way.)

Note that this would prevent future overriding of these keys by commands.
`input` (rename to `gen`), `solution`, and `visualizer` are used for that currently.

---

# Result

See doc/generators.yaml for the result. In brief:

```
testcase:
```
indicates `testcase.{in,ans,...}` are given directly in the target directory (discouraged).
```
testcase: generator.py {seed}
```
indicates the generator to run
```
testcase:
  copy: manual_cases/sample/3
```
copies `manual_cases/sample/3.*` to `testcase.*`
```
testcase:
  in: "1"
  ans: "1"
  desc: Identity of 1.
  hint: What is the identity function?
```

Priorities:
```
testcase:
  # First generate things
  generate: generator.py {seed}
  # Then copy in given files (overwriting generated files)
  copy: path/to/files
  # Then overwrite with hardcoded content
  in: 1 2
  ans: "3"
```

## Numbering:
- Testgroups are numbered per directory as `g1-<name>`, `g2-<name>`.
- Testcases are numbered globally within the problem.


RagnarGrootKoerkamp:
This turned out to be quite easy to implement. @mpsijm @mzuenni @thorehusfeldt who wants to give it a try?

todo:
- a bit more testing
- docs
- update generators.yaml
- update corresponding json and cue specs 

What I implemented now:
In each testcase, at least one of `input` (the current command) and `in` (the new hardcoding) is required.
Then:
- run the `input` command if present
- copy over `in:` to `testcase.in` (overwriting the command output), and also copy over other present keys in `ans`, `hint`, `desc`.
- validate as usual.

Most of the code in the PR is actually to get the caching and info messages and such correct.

thorehusfeldt:
I like this a lot. If I understand correctly, we are at something like this:
```json
generator :: command | {
    input: command
    in: string
    ans: string
    desc: string
    hint: string
    file_config
    directory_reserved
    ...
}
```
where the new fields are `in`, `ans`, `desc`, and `hint`.

(`input` should _really_ be `command` or `invocation` or `generator-invocation` (but not `generator`; sooner is better than later.)

I can’t see a use-case for having both (`solution` and `ans`), nor for having (`input` and `in` and `ans`), we could forbid these. Most other combinations do make sense for me but I haven’t thought this through.

thorehusfeldt:
Should 
```
my_testcase: 
```
still exist?

Or do we now have 
```
my_testcase:
  in:
  ans:
```
(I prefer the latter, it’s more explicit.)

RagnarGrootKoerkamp:
* Let me rename `input` => `command` now. It's a bit verbose but it's very unlikely you'll need to expand `testcase: command` into the long form anyway.
* Good idea to keep things simple and require **exactly one** of `in:` and `command:`.
* So far the semantics of `solution` are that it's run whenever `.ans` is not generated by `command`. That is compatible with having both `ans:` and `solution:` in a testcase, but agreed it's kinda silly so probably this should at least be a warning anyway.
* Hmm; I do like `sample-1:` as an empty key. Another problem I ran into with the tests is that we have some problem there where we provide `data/sample/1.{in,ans,png}`. After changing the yaml to
  ```yaml
  1:
    in: 1 2
    ans: 3
  ```
  the changes in this PR triggered a warning for `1.png` existing while it isn't generated. I reverted that so that for samples having extra files is OK again. Basically we don't have a good way currently of saying explicitly that `1.png` exists and was manually created, unless we add image extensions to the list and only allow them to be empty.
* In the current implementation and empty `in:` value simply means that the input should be empty, but this is somewhat conflicting with an empty `testcase:` key which means it's manually provided.

A bigger change could be to not allow any manually provided files but I think that's too much.

thorehusfeldt:
> A bigger change could be to not allow any manually provided files but I think that's too much.

My only uneasiness are manually provided files _that are not explictly listed_.

I’m fine with
```
"tc": /manual_inputs/bespoke_testcase.in
```

But I admit that I no longer have a consistent model for what “absence of value” means in `generators.yaml`.

thorehusfeldt:
I expect the following to be used a lot:
```yaml
data:
  sample:
    data:
      - :
          in: |
            123 32
            foo
            bar baz
      - :
          in: |
            2 45
            eki eki
            ptang
```

(And this is excellent! Finally a good way to specify samples.)

The indentation seems excessive, but this is proper yaml. (It looks weird because `- :` adds to layers of indentation.) 

Nag: If the user (understandably) forgets one level of indentation, as in 
```yaml
data:
  sample:
    data:
      - :
        in: |
          123 32
          foo
          bar baz
      - :
        in: |
          2 45
          eki eki
          ptang
```
the `bt generate` emits the unhelpful
```
FATAL ERROR: Dictionary must contain exactly one named testcase/group: sample
```

thorehusfeldt:
Consider being more explicit by adding the key `path:` go CUE object `generator`, next to `in`, `ans`, `command`, `desc`, etc:

```yaml
foo: 
  path: manual/foo.in
bar:
  in: 34 64
baz:
  command: random {seed}
bum:
  path: null # expect bum.in, deprecated
```
Setting exactly one of `path`, `in`, and `command` makes sense.

When the value is yaml-string (instead of yaml-dictionary), it is interpreted as a shorthand for either `path` (*) or `command` using some pattern matching:

```yaml
a : manual/foo.in # end in .in and contains no whitespace: interpreted as a path
b : random {seed} # a command
c : null # same as "path/to/c.in"
d :  # same as null
```
The introduction of `path` is mainly to avoid
```
command: manual/foo.in
command: 
```
which look wrong to me.

* I have no strong opinion on what `path:` should called. Other suggestions are `copy-from:` , `cp:` or `manual:`.
* Maybe overkill, but `path: [manual/foo.in, manual/foo.ans]` (allowing a list) makes sense to me, as does `path: manual/foo` (specifying both `manual/foo.in` and `manual/foo.ans` if they exist. But it’s implicit and therefore I don’t really like it.) I’m not sure this is a good idea.

thorehusfeldt:
After some thinking (and actually writing a specification):

path-to-infile should have the same syntax as path-to-ansfile, and the above proposal breaks that. Alternatives:

```yaml
tc-a:
  in: 34 643     # string specifies contents
  ans: 54        # string specifies contents
tc-b:
  in-path: manual/sample/foo.in # string specifies path
  ans-path: manual/sample/foo.ans # string specifies path
```
Arguably, `.hint` and `.desc` should also be first-level citizens (so we should have `hint-path` etc.), but then we could even elevate `.png`. But I don’t think any of this will ever be used.

Alternative is to be lenient in the value specification: 
```yaml
tc-c:
  in: 34 54 # string specifies contents of `tc-c.in`
tc-d:
  in: manual/secret/group4/foo.in # string looks like a path, so it's a path
```
That gives me gray hairs but is close in spirit to what `generators.py` does right now.

The third, possibly cleanest suggestion would be the field `copy`, which takes a suffix-less path:
```yaml
tc-e:
  in: 54 12 # string specifies contents
  copy: manual/sample/foo # copy whatever `.in`, `.ans`. `.whatever` is found at that path
```
The order might be:
1. if key `command` is present, run it, possibly creating `in`, `ans`, etc.
2. if key `copy: <path>` is present, copy `<path>.in`, `<path>.ans` etc., to `testcase.in` etc.,  possibly overwriting the existing files
3. if keys `in`, `ans`, `desc`, `hint` are present, copy those strings to `testcase.in` etc.,
4. if still no `testcase.ans` has manifested, run `solution`

thorehusfeldt:
I think this is what I would prefer the specification to be:
```json
package problemformat
import "struct"

#command: !="" & (=~"^[^{}]*(\\{(name|seed(:[0-9]+)?)\\}[^{}]*)*$")
#file_config: {
	solution?:    #command        // null disallowed (specify #testcase.ans instead)
	visualizer?:  #command | null // null means: skip visualisation
	random_salt?: string
}

#testcase: null | #command  | { // null means: expect file in <problemname>/data/...
	command?: #command   // invocation of a generator
	path?: #path         // copy <problemname>/generators/path.in etc.
	in?: string          // explicitly given
	ans?: string
	desc?: string
	hint?: string
	#file_config
}

#data_dict: [string]: #testgroup | #testcase

#testgroup: {
	#file_config
	testdata_settings?: #testdata_settings
	data: #data_dict | [...{#data_dict & struct.MaxFields(1)} ] // list of singleton dicts
}

#Generators: {
	generators?: [string]: [...string]
	#testgroup
	...                  // do allow unknown_key at top level for tooling
} & { data: close({          // top level data dict has keys {sample, secret}
	sample: #testgroup
	secret: #testgroup
	})
}
```
(I have now a CUE package `problemformat` that specifies `testdata.yaml` files, but which doesn’t belong in the generators spec. It defines `#testdata_settings`.)

RagnarGrootKoerkamp:
ok, many new ideas here to reply to. tl;dr: I really like your 'order of steps' proposal.

## 'inline' testcases
I call testcases for which `in`/`ans` files are provided directly in `data/*` 'inline' in the code.
These are currently specified with `testcase:` (with an empty/null value).

- We never used these for secret data
- We typically use this sample data:
  ```yaml
  sample:
    data:
      '1':
      '2':
  ```
  or I believe even this works to ignore that directory completely:
   ```yaml
  data:
    sample:
  ```
- Now that we can encode sample data directly, I don't see actually see much need for this at all anymore. Either you don't use generators and just manually put all your data in `data/` as you like, or you opt in to use this and then you need to be explicit about everything and basically give up all control over `data/`.

  This 'middle ground' of allowing a subset of `data/` to be 'inline' cases while the rest is generated also gives some mental overhead; making it fully generated is simpler and more consistent.

- One issue is that the `sample: '1':` syntax is used a lot in existing problems, so either way the implementation should keep supporting that, probably with a deprecation warning.

## 'manual' testcases
There are testcases currently given as `testcase: manual/testcase.in`.

I like your `path:` key that takes a path without extension:
```yaml
testcase:
  path: manual/testcase
```
where the semantics are that all files `generators/manual/testcase.EXT` for known extensions `EXT` are copied to the testcase in `data/`. This also removes the current inconsistency that `testcase: manual/testcase.in` actually copies all extensions, not just the `.in`.

Regarding the name:
- `copy-from:` could also work but sligtly long
- `cp:` is slightly too linuxy/unclear
- `copy:` would work well for me
- `path:` isn't super informative about what's actually being done with this path, but is OK.

Then again, `command:` is also just saying _what_ follows, not what is _done_ with it. It could also be e.g. `execute:` or `run:`.

thorehusfeldt:
I like option 2 and 3 (explicit) more than 1 and 4 (weird semantics of `null`.)

mpsijm:
Specific feedback for option 4: I think that the key should not be named `png:`, because the visualization is not always a PNG. Perhaps we can "reuse" the `visualizer:` key for this, although we then get overloaded behaviour where the value of this key is either a command to run for the visualizer, or a path to a pre-generated image.

To fix that, here's a proposal that is in some way a mix between options 3 and 4:

We could allow keys like `in`, `ans`, `hint`, `desc`, etc. to be either a string or a `path:` object. For example:
```yaml
- '':
    in: 2 4  # Content of a testcase, copied to `1.in`
    ans:
      path: manual/sample/1.ans  # Path to the answer, copied to `1.ans`

- '':
    in: 2 4  # Content of a testcase, copied to `2.in`
    ans: 6  # Content copied to `2.ans`
    visualizer:
      path: manual/sample/2.png  # Path to the PNG, instead of a command to run
```

thorehusfeldt:
`image` (or even `visualization`) may be better than `png`. Alternatively, support `png` `jpeg` `jpg`, and `pdf`. To me, keys are “conceptually cheap” (adding more keys adds little cognitive overhead and very little extra code), whereas overloading interpretations or funky semantics are ”conceptually expensive”. So I prefer the former.

RagnarGrootKoerkamp:
Oh yes, anywhere I wrote `png` I implicied meant that this can be any of the valid image extensions, and you have to choose the right one. That way it's consistent with `in` and `ans` directly corresponding to files.

mpsijm:
@thorehusfeldt I agree with you that "conceptually cheap" specifications are to be preferred.

In practice, I do not think we need the granular control that option 5 provides (why would we ever want not all files matching a prefix to be copied?). The way how option 3 copies all manual files matching the prefix given in the global-for-the-testcase `path:` key except those that have their contents explicitly provided, should be enough.

I also like options 2 and 3 best, in that case, but I also think we should keep maintaining options 1 and 4 (but with an empty value instead of an explicit `null`) for backwards compatibility (and I wouldn't mind adding a warning there :slightly_smiling_face:).

thorehusfeldt:
> but with an empty value instead of an explicit `null`

My understanding is that `foo: `, `foo: null`, and  `foo: ~` are the same in yaml.

RagnarGrootKoerkamp:
Ok so:
* option 1 (`testcase: null`) is already supported by backwards compatibility for samples (`data: sample: '1': null`) and should keep working (with a warning).
* options 2 and 3 come by default with the current proposal (the list in [this comment](https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/272#issuecomment-1617920571))
* Option 4 would require additional logic for an empty `testcase: png: null` key. But since we don't like this so much anyway, I'd prefer to just not implement this and disallow empty `testcase: <ext>: null` keys completely. We only keep the empty `testcase: null` for completely inline testcases.

Thus, you can either keep your testcase completely inline, or move all of it into generators.yaml.

RagnarGrootKoerkamp:
Ok I've implemented the 6-step approach. I think some code became quite a bit cleaner after some rewriting. Having a structure to start with makes the code much more structured as well :)

Didn't test anything yet so probably I have to make some more iterations before it actually works -- we'll see.

thorehusfeldt:
Brilliant! My only quibble is the terminology “inline”, which doesn’t attach to the right hooks in my brain. If anything, for me “inline” refers to the _hardcoded_ testcases, much like inline assembler code in a C source.

(Since inline it appears in an error message, and possibly will make it to the documentation, this is not just a source-code issue.)

RagnarGrootKoerkamp:
Yes I was also thinking the only remaining todo is to fix terminology, since I agree `inline` is confusing now that we have `hardcoded`. Feel free to suggest an alternative; I can't really come up with anything atm.

What are your opinions on `path: manual/testcase` vs `copy: manual/testcase` vs `source: manual/testcase`?
And what about `command: gen.py` vs `run: gen.py` vs `exec: gen.py`?

mpsijm:
Some alternatives for `inline` that come to my mind:
- `implicit`, because they're not _explicitly_ listed in `generators.yaml`
- `unlisted`, for a similar reason, although it is still "listed" in some way (albeit with an empty value)
-

thorehusfeldt:
I like _implicit_.

Here is my suggestion:
```cue
#testcase: #command |      // same as { command: #command }
	null |                 // same as { path: null }
	{
		command?: #command // invocation of a generator
		path?:    #path |  // copy generators/path.{ext} to path/to/testcase{.ext} 
			null           // leave path/to/testcase{.ext} alone
		in?:   string      // explicitly given contents for .in
		ans?:  string
		desc?: string
		hint?: string
		#image_contents?
		#file_config
	} 

// one of:
#image_contents: {png: string} | {jpg: string} | {jpeg: string} | {pdf: string}
```
* The `#image_contents` pattern ensures that at most one of those four keys can appear; no need for the tool to enforce that but we might as well spec it.
* Only two values are allowed to be `null`
* I think we should also require at least one of `command`, `in`, `path` to exist (the CUE expression doesn’t do that.)

RagnarGrootKoerkamp:
* `implicit` sgtm
* Will have to think more about `(exec, copy)` vs `(command, path)`. I'm starting to get used to the latter.
* I don't see the need for a separate `image_path` if we have `path`.
* After some thought, I like the explicit `path: null` addition. This way we can indeed explicitly mix implicit and generated cases. On the other hand, I don't think we really need this on top of `testcase: null`, so not sure it's worth the added complexity.
* Yes, one of `command`, `in`, and `path` makes sense.

thorehusfeldt:
>  On the other hand, I don't think we really need this on top of `testcase: null`, so not sure it's worth the added complexity.

I am very happy to let it go, but earlier you expressed sympathy being backwards compatible with:
```yaml
sample:
  data:
    '1':
    '2':
```
or even  what I call “surprised Dipsy”:
```yaml
sample:
  data:
    - '':
    - '':
```
That is exactly what the `#testcase: null` rule allows. Otherwise you’d need to rewrite old `generator.yaml`s to

```yaml
sample:
  data:
    '1':
      path:
    '2':
      path:
    '3':
      path:
```
I dislike both (and want to at least deprecate them, if not excise.)

thorehusfeldt:
Here’s a stricter specification:
```yaml
import "list"
// testcase v. 0.3

#testcase: this= #command | {	
    exec?: #command // invocation of a generator
    copy?:    #path   // copy generators/path.{ext} to path/to/testcase{.ext} 
    in?:   string      // explicitly given contents for .in
    #AtLeastOne: true & list.MinItems([ for label, _ in this if list.Contains(["exec", "copy", "in"], label) {label}], 1)
    ans?:  string
    desc?: string
    hint?: string
    #file_config
} 
```
* disallows `null` in `path:`
* gets rid of `#image_path`
* uses `exec` and `copy` just to see how that looks
* uses the field pattern from https://cuetorials.com/patterns/fields/ to ensure “at least one of”, but it’s very clumsy (CUE has to evolve before this becomes readable.)

thorehusfeldt:
In case anyone cares, here’s my best shot at specifying `testdata.yaml`:
```cue
package problemformat

import "strings"

#testdata_settings: {
	on_reject?: "break" | "continue"
	grading?: "default" | "custom"
	grader_flags?: string
	if grading != "custom" { grader_flags? : #default_grader_flags }
	input_validator_flags?: string
	output_validator_flags?: string
	accept_score?: #score
	reject_score?: #score
	range?: string // could be much improved, good CUE exercise
}

#score: =~ "^-?([0-9]+|[0-9]*.[0-9]+)$" // matches "1", "06", "21", ".4", "1.2", even negatives
#filename: =~ "^[a-zA-Z0-9][a-zA-Z0-9_.-]*[a-zA-Z0-9]$"
#path: =~ "[a-zA-Z0-9_.-/]*"

// Default grader
// --------------

#default_grader_flags: this={
    string  
    _as_struct: { for w in strings.Split(this, " ")  { (w): _ } }  // convert to struct and ...
    _as_struct: #valid_default_grader_fields                       // ... validate its fields
    }

// Default grader flags (converted to fields of a CUE struct for validation)
#valid_default_grader_fields: {
    #verdict_aggregation_mode?   // at most one verdict aggregation mode
    #score_aggregation_mode?     // at most one score aggregation mode
    ignore_sample?: _            // two more optional flags
    accept_if_any_accepted?: _
}

#verdict_aggregation_mode: {first_error: _ } | *{worst_error: _  } | {always_accept: _ }
#score_aggregation_mode:  {min: _ } | {max: _ } | *{sum: _ } | {avg: _ }
```

mpsijm:
> Maybe we should also require at least one of `command`, `in`, `path` to exist in the struct

We should be clear about which combinations we do and do not allow. E.g., it does not make sense to provide both `command`/`exec` and `in` (or `ans`). However, `path`/`copy` can be used in combination with any other key, because it will just copy all files with known extensions that are not already provided by other keys.

mpsijm:
Rebase note: in commit 0fe4934, `input` was renamed to `command`, but the CUE specification no longer existed? Not 100% sure if this specification should still be there (and it's probably out-of-date by now), so just letting you know :stuck_out_tongue:

thorehusfeldt:
* the combination `command` and `ans` makes sense to me, in fact I’m working on a problem using exactly that. (Think of this: a generator that creates a large, random no-instance to 3-Sat. I _do_ want to call `command: random -n 20 -m 28 --satisfiable 0`, but I also want to specify `ans: 0`. It both increases problem quality (because `bt validate` now has a ground truth that even the default `solution` is run against) and speeds up generation (because `solution` needn’t be run). 
* I think I agree that `command` and `in` makes no sense. I’m in favour of disallowing it.

thorehusfeldt:
@mpsijm : Regarding the missing/rotting CUE spec, I moved those activities into relative safety at https://github.com/thorehusfeldt/BAPCtools/blob/spec/support/generators_yaml_schema_improved.cue and continue noodling around there. When the decisions about 
* allowing `#testcase: null`? (I weakly think no)
* allowing `#testcase: path: null`? (I weakly think no)
* `command`, `path` versus `exec`, `copy`? (I have no opinion but note (a) that `include` is a verb, so more consistent with `exec` and `copy` while (b) `exec` is offputtingly unixy, so maybe it’s either `execute`/`copy` or `exec`/`cp`.)

are made, I’ll edit my CUE file and send @RagnarGrootKoerkamp a pull request. Once that is done, I’ll re-sync the (ugly) JSON file.

mpsijm:
> the combination `command` and `ans` makes sense to me, in fact I’m working on a problem using exactly that. [...]

That makes sense, nice example! :slightly_smiling_face: In that case, the only mutually exclusive keys should be `in` and `command`/`exec`, and specifying `ans` should always be possible.

> [...] speeds up generation (because `solution` needn’t be run).

As long as `bt validate` actually checks that the reference solution generates the same answer, I'm fine with that :slightly_smiling_face: 

> I moved [the CUE specs] into relative safety [...]

Ahh all right, missed that, thanks! I've rebased it away again :smile:

> When the decisions about [...]:

- allowing `#testcase: null`?
  → I'd say "yes", but with a warning (for backwards compatibility).
- allowing `#testcase: path: null`?
  → I'd say "no", since this is new syntax, and when a user wants to do something more complex than just running a generator command, they probably also want to be explicit about their paths.
- `command`, `path` versus `exec`, `copy`?
  → (a) I don't know where you got `include` from? But I do agree that it's nice to be consistent with using verbs vs. nouns, which also relates to the argument to make the keys self-explanatory in what they _do_, instead of just what they _are_.
  → (b) I don't care too much about "unixicity" (nice term btw :stuck_out_tongue:). For new users it's nice that keys are written out in full (`execute` rather than `exec`), but for power users it's nice that keys are short so you don't need to type much. :joy: 
  → Another suggestion for `command`/`exec`: how about we call this key `generator` instead? That is also more consistent with the other keys that take a `#command`, namely `solution` and `visualizer`. Still a noun instead of a verb, but I think that makes sense here.
  → For `path`/`copy`, I tried thinking of a name that also explains more "what it is" in a noun instead of a verb, but nothing obvious comes to mind. I can come up with phrases, but not with (preferably short) single words. The best I got was `predefined`, but I'm not really a fan of that either.

thorehusfeldt:
I guess I’m at

```cue
// rfc-th-003
#testcase: 
	#command              // same as create: #command
	null |                // deprecated
	{
		create?: #command // invocation of a generator
		copy?:   #path 
		in?:     string   // explicitly given contents for .in
		ans?:    string
		desc?:   string
		hint?:   string
		#file_config
	} 
```

mpsijm:
2. Ah of course, thanks for the reminder :slightly_smiling_face:
3. Good point :joy: 

I'm fine with `create`, it looks semantically clearer to me than `exec`(`ute`). I'm neutral on `command`.

> And advantage of `path` over `copy` is that it is less likely to be confused with what `include` does.

I haven't really worked with test groups yet (except the default `sample` and `secret`), so I also don't have an opinion about this.

---

Perhaps the specification starts to get at a point where we can ask feedback from other users of BAPCtools? I had a discussion with Ragnar about the syntax of #271, and bottom line is that we would want to try things out in real contests, but it's a hassle to change the syntax once it's merged to `master`. And if our test subjects are also indifferent about `create` vs. `command or `path` vs. `copy`, we'll just pick either and be done with it :joy:

RagnarGrootKoerkamp:
> [...] speeds up generation (because solution needn’t be run).

> As long as bt validate actually checks that the reference solution generates the same answer, I'm fine with that 🙂

In the current implementation that's not the case. You have to `bt run` for that, but actually I think that's OK. the `solution:` is only used to generate ans files and validate their syntax usually. Verifying correctness you usually do with `bt run` and ensuring all solutions give the same answer.


On naming:
* I think `create` isn't very specific. `command`, `path`, and `in` are all ways to 'create` a testcase
* `copy` being confusing with `include` is a good point
* I think my preference is back to what we have: `command:` and `path:`. `command` is quite descriptive (I mean, clearly the command will be run). `path` slightly less so, but consistent with `command`.
* "First _run_ the **command**, then *copy* from **path**, and then *write* the **in**'
* Only other suggestion could be `source` instead of `path` but source is actually less specific.

RagnarGrootKoerkamp:
Another point: we could change the order of operations to
1. `path`
2. `in`
3. `command`
so that the generator could read the other files as input. But I kinda prefer to explicitly run the `command` first to avoid this situation and simplify these hidden dependencies.

mpsijm:
> You have to `bt run` for that [...]

Makes sense :smile:

> On naming:

I'm fine with settling for `command` and `path` :slightly_smiling_face: 

>the order of operations

Note here that `command` and `in` should probably be mutually exclusive, see https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/272#issuecomment-1621321311. Hidden dependencies indeed sounds like a bad idea. In my mind, the order of operations would be "first handle all keys except `path`, then let `path` copy all files with known extensions that have not yet been generated by other keys" (and then finally, if no `.ans` file is created yet, run `solution` on the `.in` file).

@thorehusfeldt one more note about the CUE specification: if a testcase is specified as only a string (i.e., `#testcase: #command`), this string may also be a `#path`. Currently, this is disambiguated from `#command` thanks to the `.in` suffix, but I assume that disambiguation will still be possible when we leave out the suffix :slightly_smiling_face:

RagnarGrootKoerkamp:
I've replaced the .in disambiguation with a warning/error. It still works but shouldn't be used imo. The fact that it ends in .in is ugle since it actually copies all extensions.

thorehusfeldt:
After some head-scratching and soul-searching, I now _like_ `generate` (as a verb).

I’m now at
```cue
// rfc-th-004
#testcase: 
    #command |            // same as create: #command
    null |                // deprecated
    {
	generate?: #command // invoke a generator
	copy?:     #path // copy from path
	["in" | "ans" | "desc" | "hint" ]: string // give contents explicitly
	#file_config
    } 
```
We could add 
```cue
numexist(>0, generate, copy, in) // create an .in-file
numexist(<=1, generate, in)      // don't use both `in` and `generate`
```
This CUE syntax is not implemented yet. Not sure whether that belongs to the specification, though. (What we _actually_ want is that the `copy`-directive points to a path where a file with extension `.in` is present, and we can’t specify that anyway.)

RagnarGrootKoerkamp:
> numexist(>0, generate, copy, in) // create an .in-file

yes, that seems like a good lower bound.

> numexist(<=1, generate, in)      // don't use both `in` and `generate`

I'm not 100% convinced there are no usescases for having both. You could generate a `.hint` or `.desc` while hardcoding the `.in`. But then again this is quite obscure. I think I'm fine either way (with/without this restriction).

> `(copy, generate)` works for me. Slight inconsistency is that BAPCtools does the `copy`, while BAPCtools `executes` the `command` to `generate` things. But `generate` for sure is clear.

thorehusfeldt:
> I'm not 100% convinced there are no usescases for having both [`generate` and `in`]

You’re right. I retract my suggestion.

RagnarGrootKoerkamp:
Ok pushed
* rename `path` => `copy`
* rename `command` => `generate`
* require at least one of `generate`, `copy`, `in`

This also improves the code IMO, because `path` is used all over while `copy` is much more unique, and same for `generate` being more specific than `command` (because ie `solution: <command>` also includes a command).

thorehusfeldt:
I believe this is consistent with what I have at https://github.com/thorehusfeldt/BAPCtools/blob/spec/support/schemas/generators.cue 

(Except I don’t know if `#testcase: #path` is allowed.)

RagnarGrootKoerkamp:
From my side:
* `#testcase: #path`: (4): Out of spec and not in BAPCtools
* `#testcase: null`: Either (2) or (3), but I think long term (3) [out-of-spec but supported for legacy] makes most sense.

mpsijm:
I tried the current implementation from a problem setter's perspective, on BAPC 2022 problem Adjusted Average. Some remarks:

- When running `bt generate` without changing the current `generators.yaml`, I got:
  ```
  ERROR: Use the new `copy: path/to/case` key instead of manual/break-value-overlap.in.
  ERROR: Use the new `copy: path/to/case` key instead of manual/take-0.in.
  ```
  But it still copied over the input file and generated the answer from it. That feels weird from a user's perspective, because "error" feels like it should abort processing that test case entirely. Why not make it a warning instead? That will still give exit code `1` in the end, which probably is the end-goal here.

- After that, I started applying the new syntax for `copy`. I like how both of these work as intended:
  ```yaml
  - break-value-overlap: {copy: manual/break-value-overlap}
  - take-0:
      copy: manual/take-0
  ```

- Writing `take-0: manual/take-0` gives `adjustedaverage/generators/manual/take-0 does not exist.`, which makes sense I guess, because we no longer support the old path syntax anyway. However, doing the opposite and writing `take-0: {copy: manual/take-0.in}` throws an exception:
  ```
  ERROR: `copy: manual/take-0.in` should not include the extension.
  Traceback (most recent call last):
    File "/home/maarten/bin/bt", line 1049, in <module>
      main()
    File "/home/maarten/bin/bt", line 1045, in main
      run_parsed_arguments(parser.parse_args())
    File "/home/maarten/bin/bt", line 896, in run_parsed_arguments
      success &= generate.generate(problem)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/maarten/git/contest/BAPCtools/bin/generate.py", line 1759, in generate
      config.run()
    File "/home/maarten/git/contest/BAPCtools/bin/generate.py", line 1533, in run
      p.done()
    File "/home/maarten/git/contest/BAPCtools/bin/parallel.py", line 151, in done
      raise self.first_error
    File "/home/maarten/git/contest/BAPCtools/bin/parallel.py", line 80, in _worker
      self.f(task)
    File "/home/maarten/git/contest/BAPCtools/bin/generate.py", line 1526, in <lambda>
      p = parallel.Parallel(lambda t: t.listed and t.generate(self.problem, self, bar))
                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/maarten/git/contest/BAPCtools/bin/generate.py", line 478, in generate
      cwd = problem.tmpdir / 'data' / t.input_hash
            ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
  TypeError: unsupported operand type(s) for /: 'PosixPath' and 'NoneType'
  ```
  Granted, the `ERROR:` before the stack trace already shows what's going wrong, but I had at least expected that it wouldn't crash. :stuck_out_tongue: It looks like `t.input_hash` (= `self.input_hash` in `TestcaseRule.__init__`) is never set in this case. And maybe it shouldn't indeed, but then it should silently abort processing this test case and not crash.

- I tried to specify a literal test case like this:
  ```yaml
  - min:
      in: |
        2 1 0
        0 0
      ans: 0
  ```
  But this gave me a `FATAL ERROR: .ans must be of type str, found int: 0`. Can we allow `int` and `float` for `.in` and `.ans`? (looks like this check is is done at `bin/generate.py:429`, but it requires some extra branching or an extra map with allowed types for the known data types, not sure what would be best there)

RagnarGrootKoerkamp:
* Made `use new copy syntax` a warning.
* Prevented this python crash, and added an error during `generate` phase (instead of parsing phase) that the `{copy}.in` file was not found. If you included `.in` (or another extension) in the `copy: path.in`, this will clearly show it's looking for `path.in.in`, on top of the `should not include extension` error during parsing

mpsijm:
Right, in that case, I'm fine with having to quote the contents of `.in` and `.ans`. The error is clear enough anyway :slightly_smiling_face:

mpsijm:
Another bug report :innocent: When I:
- Have an existing `generators.yaml` with auto-numbered test cases
- Add a line somewhere in the middle
- Run `bt generate`
- I get the following stack trace:
```
Traceback (most recent call last):
  File "/home/maarten/bin/bt", line 1049, in <module>
    main()
  File "/home/maarten/bin/bt", line 1045, in main
    run_parsed_arguments(parser.parse_args())
  File "/home/maarten/bin/bt", line 896, in run_parsed_arguments
    success &= generate.generate(problem)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maarten/git/contest/BAPCtools/bin/generate.py", line 1770, in generate
    config.run()
  File "/home/maarten/git/contest/BAPCtools/bin/generate.py", line 1554, in run
    p.done()
  File "/home/maarten/git/contest/BAPCtools/bin/parallel.py", line 177, in done
    raise self.first_error
  File "/home/maarten/git/contest/BAPCtools/bin/parallel.py", line 106, in _worker
    self.f(task)
  File "/home/maarten/git/contest/BAPCtools/bin/generate.py", line 1547, in <lambda>
    p = parallel.Parallel(lambda t: not t.listed and t.generate(self.problem, self, bar))
                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maarten/git/contest/BAPCtools/bin/generate.py", line 477, in generate
    ext_file = t.copy.with_suffix(ext)
               ^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'with_suffix'
```
This probably has something to do with stale information about which test cases are listed and which are not? :thinking:

RagnarGrootKoerkamp:
fixed

mpsijm:
That works, thanks! :smile:

mpsijm:
Woops, I thought that would fix it, but apparently it didn't :upside_down_face: I now get the following stack trace:
```
Traceback (most recent call last):
  File "/home/maarten/bin/bt", line 1049, in <module>
    main()
  File "/home/maarten/bin/bt", line 1045, in main
    run_parsed_arguments(parser.parse_args())
  File "/home/maarten/bin/bt", line 896, in run_parsed_arguments
    success &= generate.generate(problem)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maarten/git/contest/BAPCtools/bin/generate.py", line 1770, in generate
    config.run()
  File "/home/maarten/git/contest/BAPCtools/bin/generate.py", line 1544, in run
    p.done()
  File "/home/maarten/git/contest/BAPCtools/bin/parallel.py", line 177, in done
    raise self.first_error
  File "/home/maarten/git/contest/BAPCtools/bin/parallel.py", line 106, in _worker
    self.f(task)
  File "/home/maarten/git/contest/BAPCtools/bin/generate.py", line 1537, in <lambda>
    p = parallel.Parallel(lambda t: t.listed and t.generate(self.problem, self, bar))
                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maarten/git/contest/BAPCtools/bin/generate.py", line 836, in generate
    add_testdata_to_cache()
  File "/home/maarten/git/contest/BAPCtools/bin/generate.py", line 686, in add_testdata_to_cache
    test_hash = hash_file(target_infile)
                ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maarten/git/contest/BAPCtools/bin/util.py", line 922, in hash_file
    assert file.is_file(), f"File {file} does not exist"
AssertionError: File adjustedaverage/data/secret/34-rand.in does not exist
```
Now tested on BAPC 2022 Adjusted Average, by duplicating the `rand` testcase on line 68 and changing one of them to have `{seed:0}` instead of `{seed:1}`. The problem here is that the 34th test case was only just now added, so it indeed did not exist yet :sweat_smile:

doc/generators.yaml
```diff
@@ -81,17 +81,17 @@ data:
     data:
       "1": stdout.py 1 # prints `1` to stdout, which is piped to 1.in
 
-      # To indicate a manual testcase already present in the corresponding directory, specify it with an empty value.
+      # To indicate a unlisted testcase already present in the corresponding directory, specify it with an empty value.
       # Prefer specifying a path to a .in file as in case `3` below.
       # E.g., this indicates that `data/sample/2.in` (and optionally `data/sample/2.ans`) are existing non-generated files.
       "2":
-      # The copy: key indicates a manual testcase that will be copied
+      # The copy: key indicates a unlisted testcase that will be copied
```
mpsijm, line 88: Shouldn't this still be `manual`? :thinking: Same for `manual_cases` below, and in `skel/generators/generators.yamlL45`

RagnarGrootKoerkamp: thanks for checking :) it was inevitable this would end up too greedy on the search-and-replace

thorehusfeldt:
Is anything blocking the inclusion of the generator improvements? I’m really happy with it.

(My own suggestion to restrict the testcase names in the problem package specification was rejected, so maybe somebody else feels it’s too much of a restriction on BAPC-tools that its generators can only ever construct problem-unique testcase names. In that case we’d want to rethink it. But I still like the restriction and don’t see a problem with BAPCtools being more strict.)

RagnarGrootKoerkamp:
Yes I think this is good to go. Will merge now so we can use/test/improve it during NWERC development.

Re. unique test case names: the only thing that is restricted now is that if you do `include: name`, then `name` must be the unique name of a testcase or testgroup.

Issue #217, opened 2022-09-20
Pin runs to core


RagnarGrootKoerkamp:
java and kotlin can use more cpu time than user time because they seem to do JIT in separate threads.

Pinning them to a single core would prevent that, probably using https://docs.python.org/3.7/library/os.html?highlight=affinity#os.sched_setaffinity.

Not super trivial because only testcase runs should be pinned, and not compilation. And we must make sure to never pin two runs to the same cpu. Also it's not clear to me yet whether we can pin processes directly on creation

RagnarGrootKoerkamp:
Simply putting the following in `limit_setter` works, but we need to only pin test cases, not compilations.

```
        # Pin everything to CPU 0 when only running 1 job in parallel.
        if pin and config.args.jobs == 1:
            os.sched_setaffinity(0, [0])
```

mzuenni:
added with ee9fcc5f143802d792b93db9d0dd0e1ab6cb95b0

PR #285, opened 2023-06-30
Pin #217 

mzuenni:
Implemented pinning, but I have some notes:

-  I am not sure if it is a good thing to add `cores` to so many functions...
- right now only submissions in fuzz/run are pinned but not during solution generation
- For interactive problems, all programs are pinned to the same core

### DIFF
--- a/bin/parallel.py
+++ b/bin/parallel.py
@@ -34,28 +34,26 @@ class Parallel:
 
         if self.num_threads:
             if pin:
+                # only use available cores
                 cores = list(os.sched_getaffinity(0))
-                if len(cores) > self.num_threads:
+                if self.num_threads > len(cores):
                     self.num_threads = len(cores)
-                    util.log(f'{num_threads} jobs requested but only {len(cores)} cores found. Using only {len(cores)} jobs!')
+
+                # sort cores by reversed id. If num_threads << len(cores) this ensures that we 
+                # use different physical cores instead of hyperthreads
                 num_bits = max(map(lambda c: c.bit_length(), cores))
                 cores.sort(key=lambda c: util.bit_reverse(c, num_bits))
 
             self.threads = []
             for i in range(self.num_threads):
-                args = [cores[i]] if pin else []
+                args = [{cores[i]}] if pin else []
                 t = threading.Thread(target=self._worker, args=args, daemon=True)
                 t.start()
                 self.threads.append(t)
 
             signal.signal(signal.SIGINT, self._interrupt_handler)
 
-    def _worker(self, pin=False):
-        if pin != False:
-            #TODO store this threadlocal...
-            #TODO read this in limit_setter
-            #storage = threading.local()
-            #storage.__setattr__('pin', pin)
+    def _worker(self, cores=False):
         while True:
             with self.mutex:
                 # if self.abort we need no item in the queue and can stop
@@ -77,7 +75,10 @@ class Parallel:
             # store the first exception for later
             try:
                 current_error = None
-                self.f(task)
+                if cores is False:
+                    self.f(task)
+                else:
+                    self.f(task, cores)
             except Exception as e:
                 self.stop()
                 current_error = e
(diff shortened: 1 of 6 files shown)


RagnarGrootKoerkamp:
Do you observe differences in timings? Ie more stable and/or faster?

bin/parallel.py
```diff
@@ -31,15 +33,27 @@ def __init__(self, f, num_threads=True):
         self.finish = False
 
         if self.num_threads:
+            if pin:
+                # only use available cores
+                cores = list(os.sched_getaffinity(0))
```
RagnarGrootKoerkamp, line 38: likely this only works on linuxy systems

bin/parallel.py
```diff
@@ -31,15 +33,27 @@ def __init__(self, f, num_threads=True):
...
+                if self.num_threads > len(cores):
+                    self.num_threads = len(cores)
```
RagnarGrootKoerkamp, line 40: do 1 less to give some space to the OS. could even pin bapctools itself to that thread, but maybe not needed.

bin/parallel.py
```diff
@@ -31,15 +33,27 @@ def __init__(self, f, num_threads=True):
...
+
+                # sort cores by reversed id. If num_threads << len(cores) this ensures that we 
+                # use different physical cores instead of hyperthreads
+                num_bits = max(map(lambda c: c.bit_length(), cores))
+                cores.sort(key=lambda c: util.bit_reverse(c, num_bits))
```
RagnarGrootKoerkamp, line 45: so on my 6-core 12-thread machine, i think 1 and 7 are on the same core, and 1..=6 are on distinct cores. 
I also always assumed hyperthreads would be consecutive but apparently not. (You can confirm this using `cat /proc/cpuinfo` or so i think.)

RagnarGrootKoerkamp (commented):
Overall the code looks ok i suppose. as you say, it's mostly passing data through a lot of layers which ideed does not look so pretty :/

bin/parallel.py
mzuenni: hm you seem to be right...

RagnarGrootKoerkamp:
Not sure exactly is 'thread local global variables' are a thing in python, but maybe you use those to pass the `cores` info?

Actually, you could possibly just pin each thread to a core the moment it's spawned. At least for programs the 'affinity' is inherited I think, and this may also work for threads? That would for sure simplify things, and running the python part of the threads on those cores as well doesn't seem like a drawback? (the python anyway is idle while the programs run. Apart from interactive problems)

mzuenni:
As I understand the doc `setaffinity` is set for a `pid` and threads run under the same pid... But it still seems to work as far as I can see o.O

RagnarGrootKoerkamp:
Hmm, `pid = 0` may work?

mzuenni:
yes, that is what I tried and what seems to work... but I did not expect it to work... anyway I prefer this and will change it

mzuenni:
I kept the code in the `limit_setter` even though we don`t use it right now...

RagnarGrootKoerkamp:
Ok LGTM and works on my machine. Didn't see a speed difference on the one small problem i tested though. Maybe we should bump the default parellelism (`config.args.default_jobs`) from `os.cpu_count()/2` to `os.cpu_count()-1`? Or even better make it the number of physical cores instead of hyperthreads (if we can easily get that without extra libraries).

mzuenni:
That doesn't seem to be easy... the only way I found so far is `psutil`.

RagnarGrootKoerkamp:
yeah... nvm then. All looks good, but I'm still not 100% convinced we actually need this. Can you make some test to show that timings become faster/more stable? otherwise there's not much point

mzuenni:
It speeds up java and interactive problems by ~50%

RagnarGrootKoerkamp:
Oh that's pretty good! Ready to merge?

mzuenni:
Yes, should be fine :)

RagnarGrootKoerkamp:
Oh it needs rebasing

mzuenni:
are you sure? ^^'

RagnarGrootKoerkamp:
> This branch cannot be rebased due to conflicts

PR #288, opened 2023-07-04
Parallel fuzz

mzuenni:
Allow more parallel fuzzing. (I am not sure if there is a better way to use the bar class for this...)

### DIFF
--- a/bin/parallel.py
+++ b/bin/parallel.py
@@ -15,6 +15,7 @@ class ParallelItem:
         self.priority = priority
         self.id = id
 
+    # Note: heapq uses a min heap, so higher priorities are 'smaller'.
     def __lt__(self, other):
         if self.priority != other.priority:
             # python priority queue is a min heap but larger priority
--- a/bin/parallel.py
+++ b/bin/parallel.py
@@ -1,11 +1,22 @@
 #!/usr/bin/env python3
 import threading
 import signal
+import heapq
 
 import os
 
 import config
 import util
+
+
+class ParallelItem:
+    def __init__(self, task, priority):
+        self.task = task
+        self.priority = priority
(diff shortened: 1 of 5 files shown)


RagnarGrootKoerkamp:
* Can you provide some high level context of these changes? The end-to-end diff on github is a mess.. (if only it wouldn't highlight indentation changes so much :/)
* It seems that you did 2 major rewrites of `fuzz.py` :sweat_smile:  kinda hard to review quickly :upside_down_face: 
* The changes to `ProgressBar` look good; that's mostly simplification and cleanup by using higher level `with` API right.

mzuenni:
I put the old fuzz code in 3 classes:
1.  The `Fuzz` class which handles:
    - building of generators, validators, submission 
    - maintain the task queue
    - adding new `GeneratorTask`s to the queue
    - rewriting the `generator.yaml`
2. the `GeneratorTask` class which handles:
    - generating a new testcase
    - validating the testcases
    - adding all submission to the queue
3. The `SubmissionTask` class which handles:
    - running a submission on a testcase
    - report that the `GeneratorTask` has to be written to the  `generator.yaml` (if necessary)
 
Both kinds of Tasks will additionally callback to the `Fuzz` class after completion which adds more generators if necessary

bin/parallel.py
```diff
@@ -1,12 +1,21 @@
...
 import config
 import util
 
+class ParallelItem:
+    def __init__(self, task, priority):
+        self.task = task
+        self.priority = priority
+
+    def __lt__(self, other):
+        return self.priority > other.priority
```
RagnarGrootKoerkamp: ok this is confusing ;) Please add a comment.

PR #291, opened 2023-07-08
Fix some frenchy errors

chrisgdt:
Hello !

I was experimenting the new language feature (#256) and noticed a weird error the second time I called `bt pdf` (not the first time) :
```
! Undefined control sequence.
<argument> ...{ \vspace {.4pt} \hfill \sffamilyFB 
                                                  \relax \fontsize {14.4}{18...
l.1 ...itel][l]{Allumettes}\hfill }{}{section.0.1}
                                                  
!  ==> Fatal error occurred, no output PDF file produced!
```

After some investigations, I found out that the babel package set up in French renames all `\sffamily` to `\sffamilyFB`, so once I compile in French, some cache files (more precisely .aux and .lop) contain this new command and are not updated anymore to gain some time. However, the `sffamilyFB` command does not exist with babel en, de, nl, da.

That is why I PR this to force `sffamilyFB` to be defined to handle cache latex files if French has previously been used.

I haven't noticed anything wrong with other languages yet (nl, fr, en, da, de) but we should keep this babel behavior in mind.

### DIFF
--- a/latex/bapc.cls
+++ b/latex/bapc.cls
@@ -333,6 +333,8 @@
 % The following are required for the overall layout:
 %-------------------------------------------------------------------------------
 \let\sffamilyFB\sffamily% Prevent error if babel fr loads \sffamilyFB in cache
+\let\ttfamilyFB\ttfamily
+\let\rmfamilyFB\rmfamily
 \usepackage[\langbabel]{babel}
 \usepackage{lmodern}
 \usepackage{graphicx}
--- a/latex/bapc.cls
+++ b/latex/bapc.cls
@@ -332,6 +332,7 @@
 %-------------------------------------------------------------------------------
 % The following are required for the overall layout:
 %-------------------------------------------------------------------------------
+\let\sffamilyFB\sffamily% Prevent error if babel fr loads \sffamilyFB in cache
 \usepackage[\langbabel]{babel}
 \usepackage{lmodern}
 \usepackage{graphicx}
--- a/latex/lang/fr.tex
+++ b/latex/lang/fr.tex
@@ -1,7 +1,7 @@
 \newcommand{\langbabel}{french}
 
 % bapc.cls
-\newcommand{\langblank}{Cette page est intentionnellement laissée vierge.}
+\newcommand{\langblank}{Cette page est intentionnellement laissée vide.}
 \newcommand{\langtimelimitmissing}{LIMITE DE TEMPS MANQUANTE}
 \newcommand{\langproblemname}{Nom du problème}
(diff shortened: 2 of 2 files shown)


chrisgdt:
I take advantage of this PR to modify the blank page formulation, to be more faithful to what Wikipedia states :
https://fr.wikipedia.org/wiki/Page_laiss%C3%A9e_intentionnellement_vide

RagnarGrootKoerkamp:
Ah so you're saying this happens when you first build in french, and then in a different language.

Maybe a cleaner solution is to make separate cache directories for each language, but it feels like that might give a lot of unnecessary overhead -- not sure.

Anyway, for now, this fix looks fine to me.

chrisgdt:
> Maybe a cleaner solution is to make separate cache directories for each language, but it feels like that might give a lot of unnecessary overhead -- not sure.
Yeah it was my first though but I had a huge laziness when I noticed that it was fixable with one single line ahah

You are totally right about rmfamily and ttfamily, those are not currently used here so no error occured but I will add a commit for this, just in case, thank you !

RagnarGrootKoerkamp:
ok i'll merge. We can separate the caches later if we feel it's needed.

thorehusfeldt:
[chrisgdt](https://github.com/chrisgdt): what happens when you use proper French typography in a solutions slide, i.e., « like this »? Don’t we run into unholy font encoding issues? (Danish also uses angled quotation marks,  but »like this«.)

chrisgdt:
@thorehusfeldt I have not tried to compile solutions in French yet (or at least, I haven't noticed anything off). In the case of quotes, a proprer use of \enquote, from csquotes package, with the babel package usually adapts quotes to the language set in babel.

I won't be able to test it right now since I don't have access to my pc, but I will do it as soon as I can ! I let you know :)

thorehusfeldt:
I think I found the culprit.

`latex/solutions_base.tex` has 
```latex
\usepackage[T1, OT1]{fontenc}
```
but should have
```latex
\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc} 
```
in order to be consistent with `bapc.cls`. This makes a difference for symbols like «, », and „, which frequently appear in explanatory body text.

RagnarGrootKoerkamp:
Hmm right. I do remember adding OT1 at some point for a reason but I'm no expert on any of this so we can do the change you propose and see how it goes.

thorehusfeldt:
Ideally, these macros (used by contest, solution, and statement source files) should 
1. all `input{}` the same snippet
2. be consistent with what Kattis eventually does to the `problem.xy.tex`
I used to understand these things very well.

Issue #289, opened 2023-07-04
Improve debugging failed input/output validation


RagnarGrootKoerkamp:
Sometimes the input/output validation error message isn't very clear and a look at the actual generated files is helpful.

Maybe these errors could include the path to the temporary directory where the cases are stored for easier manual inspection.

See also #225 for an easier way to see all validator output in the terminal.

Issue #264, opened 2023-06-21
Use symlink on windows


RagnarGrootKoerkamp:
In Windows 10, symlinks are now supported by turning on developer mode (https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/).

This requires python 3.8 (https://csatlas.com/python-create-symlink/) but that should be fine, and simplifies logic.

@mzuenni

PR #292, opened 2023-07-11
Links

mzuenni:
- Use symlinks on windows if possible #264 
- display a (shortened) path to the invalid file if a format validator fails #289

### DIFF
--- a/bin/export.py
+++ b/bin/export.py
@@ -243,7 +243,10 @@ def export_contest():
     if not has_ryaml:
         for key in ('duration', 'scoreboard_freeze_duration'):
             if key in data:
-                data[key] = str(datetime.timedelta(seconds=data[key]))
+                value = data[key]
+                if isinstance(value, int):
+                    str(datetime.timedelta(seconds=data[key]))
+                data[key] = value;
 
     verbose("Uploading contest.yaml:")
     verbose(data)
--- a/bin/run.py
+++ b/bin/run.py
@@ -163,7 +163,7 @@ class Testcase:
                     file = self.in_path
                 if validator_type == 'output_format':
                     file = self.ans_path
-                data += f'in {shorten_path(self.problem, file.parent) / file.name}\n'
+                data += f'{Style.RESET_ALL}in {shorten_path(self.problem, file.parent) / file.name}\n'
             else:
                 data = ret.err
 
--- a/bin/run.py
+++ b/bin/run.py
@@ -158,6 +158,7 @@ class Testcase:
                     data = ret.err
                 elif ret.out:
                     data = ret.out
+                data += f'in {shorten_path(self.problem, self.in_path.parent) / self.in_path.name}\n'
             else:
                 data = ret.err
 
--- a/bin/run.py
+++ b/bin/run.py
@@ -158,7 +158,12 @@ class Testcase:
                     data = ret.err
                 elif ret.out:
                     data = ret.out
-                data += f'in {shorten_path(self.problem, self.in_path.parent) / self.in_path.name}\n'
+
+                if validator_type == 'input_format':
+                    file = self.in_path
+                if validator_type == 'output_format':
+                    file = self.ans_path
+                data += f'in {shorten_path(self.problem, file.parent) / file.name}\n'
             else:
                 data = ret.err
 
--- a/bin/util.py
+++ b/bin/util.py
@@ -11,6 +11,7 @@ import re
 import threading
 import signal
 import hashlib
+import tempfile
 
 from pathlib import Path
 from colorama import Fore, Style
@@ -45,7 +46,6 @@ def is_bsd():
 
 if not is_windows():
     import resource
-
 
 def debug(*msg):
     print(Fore.CYAN, end='', file=sys.stderr)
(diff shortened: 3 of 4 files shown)


bin/util.py
```diff
@@ -550,18 +559,32 @@ def strip_newline(s):
     else:
         return s
 
+# check if windows supports symlinks
+if is_windows() or True:
```
RagnarGrootKoerkamp, line 563: can you make this lazy and only do it the first time `ensure_symlink` is called? (And remove the `or True`)

RagnarGrootKoerkamp: In that case you could even just try-except the actual link to be made

bin/util.py
```diff
@@ -550,18 +559,32 @@ def strip_newline(s):
...
+    link_parent = Path(tempfile.gettempdir()) / 'bapctools'
+    link_dest = link_parent / 'dir'
+    link_dest.mkdir(parents=True, exist_ok=True)
+    link_src = link_parent / 'link'
+    if link_src.exists() or link_src.is_symlink():
+        link_src.unlink()
+    try:
+        link_src.symlink_to(link_dest, True)
+        windows_can_symlink = True
+    except:
```
RagnarGrootKoerkamp, line 573: Can you catch a specific error class?

bin/util.py
```diff
@@ -550,18 +559,32 @@ def strip_newline(s):
...
+        windows_can_symlink = False
+        warn('Please enable the developer mode in Windows!')
```
RagnarGrootKoerkamp, line 575: `... to enable symlinks`?
And add a link to instructions on how to do this. This warning should be rare so may as well intruct on how to fix.

bin/util.py
```diff
@@ -575,16 +598,15 @@ def ensure_symlink(link, target, output=False, relative=False):
...
         else:
             link.unlink()
     if relative:
         # Rewrite target to be relative to link.
-        rel_target = os.path.relpath(target, link.parent)
-        os.symlink(rel_target, link)
+        link.symlink_to(target.relative_to(link.parent), target.is_dir())
     else:
-        link.symlink_to(target.resolve())
+        link.symlink_to(target.resolve(), target.is_dir())
```
RagnarGrootKoerkamp, line 609: add a comment for what is this second args

bin/util.py
```diff
@@ -411,6 +411,15 @@ def resolve_path_argument(problem, path, type, suffixes=[]):
     return None
 
 
+def shorten_path(problem, path):
```
RagnarGrootKoerkamp, line 414: ah this is neat! Can you explain in a comment?

Normally I'd favour a more structured path, but since that path already contains a hash replacing it by a single shorter hash sgtm

bin/run.py
```diff
@@ -158,6 +158,12 @@ def validate_format(
                     data = ret.err
                 elif ret.out:
                     data = ret.out
+
+                if validator_type == 'input_format':
+                    file = self.in_path
+                if validator_type == 'output_format':
+                    file = self.ans_path
+                data += f'{Style.RESET_ALL}in {shorten_path(self.problem, file.parent) / file.name}\n'
```
RagnarGrootKoerkamp, line 166: I suppose you tested this and it looks decent in the terminal?

bin/export.py
```diff
@@ -243,7 +243,10 @@ def export_contest():
     if not has_ryaml:
         for key in ('duration', 'scoreboard_freeze_duration'):
             if key in data:
-                data[key] = str(datetime.timedelta(seconds=data[key]))
+                value = data[key]
+                if isinstance(value, int):
+                    str(datetime.timedelta(seconds=data[key]))
+                data[key] = value;
```
RagnarGrootKoerkamp, line 249: hmm are strings supported here?

RagnarGrootKoerkamp:
Some small questions/comments to clarify things. Overall this looks nice and makes me happy :)

bin/export.py
mzuenni: pyyaml seems to pars "0:xx:xx" as string but "x:xx:xx" as int

bin/run.py
mzuenni: yea thats why i shortened the paths, looks like this:
```
String length 1000001 out of range [2, 1000001)! Line: 1, Char: 0
((()()((((()((((()((()()(()()))()))))(()())())((((()(((())))((()))((()(()(([...]
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
in /tmp/bapctools_98c90c/loopinvariant/links/6eefaf/testcase.in
```

bin/util.py
mzuenni: I think its cleaner to always request the developer mode right at startup (and be more annoying to get more people to enable it)
We could also just enable it ourselves but that feels weird, is likely hacky, and might need changes from time to time...

bin/export.py
RagnarGrootKoerkamp: huh what?? it parses an `hh:mm:ss` duration as a time in seconds??

PR #294, opened 2023-07-18
Constraints

mzuenni:
Improved Constraints latex parser and added support for multiple statements


RagnarGrootKoerkamp:
Ok so you actually wrote a simple latex parser :) Impressive!

I don't see anything wrong with it on first sight so I'll just merge.

PR #297, opened 2023-07-27
More noise

mzuenni:
We should always warn if people do stuff the wrong way. (However, I am not sure how to fix the tests... the default solution shouldn't be required if there is no generstor.yaml)

### DIFF
--- a/bin/generate.py
+++ b/bin/generate.py
@@ -263,7 +263,7 @@ solution: /{solution.relative_to(problem.path)}'''
             solution_short_path = solution.relative_to(problem.path / 'submissions')
             log(
                     f'''No solution specified. Using {solution_short_path} instead. Use
-generate --default_solution /{solution.relative_to(problem.path)}
+--default_solution {solution.relative_to(problem.path)}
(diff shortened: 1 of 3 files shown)


bin/generate.py
```diff
@@ -228,12 +229,52 @@ def run_interactive(self, problem, bar, cwd, t):
...
+            return False
+        if generator_config.has_yaml:
+            # Note: we explicitly random shuffle the submission that's used to generate answers to
+            # encourage setting it in generators.yaml.
+            solution = random.choice(solutions)
+            solution_short_path = solution.relative_to(problem.path / 'submissions')
+            warn(
+                f'''No solution specified. Using randomly chosen {solution_short_path} instead.
+Setting the solution in generators/generators.yaml:
+> solution: /{solution.relative_to(problem.path)}'''
```
RagnarGrootKoerkamp, line 258: `Set this solution` and remove the `>` for easier copy pasting

RagnarGrootKoerkamp (approved):

PR #298, opened 2023-08-05
[export] update_problems_yaml: Write `name` as `str` instead of `dict`

mpsijm:
Follow-up of #256 (#109).

When running `bt update_problems_yaml`, the contest's `problems.yaml` looked like this:
```yaml
- id: helloworld
  label: A
  name: 
    en: Hello World
  rgb: '#000000'
  time_limit: 1.0
```
However, DOMjudge doesn't eat this yet ([source](https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Service/ImportExportService.php#L290)). Therefore, I propose that the `name` key in `~contest/problems.yaml` remains a `str` instead of a `dict`.
No code changes are needed for `~problem/problem.yaml`, because there is no code that automatically updates it. The user can manually set `name` to be either a `dict` or a `str`, to their choice.

(My first attempt was to overwrite `problem.settings.name` to be a `str` in `Problem._determine_statement_languages` if the only language is English, but that would require extra branches in other places. So far, I think that `problems.yaml` is the only place where this distinction matters, so I switched to the approach of only converting the name back to a `str` in `update_problems_yaml`.)

### DIFF
--- a/bin/export.py
+++ b/bin/export.py
@@ -248,7 +248,7 @@ def export_contest():
                 # (YAML 1.2 parses it as a string.)
                 if isinstance(value, int):
                     str(datetime.timedelta(seconds=data[key]))
-                data[key] = value;
+                data[key] = value
 
     verbose("Uploading contest.yaml:")
     verbose(data)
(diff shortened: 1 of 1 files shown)


mpsijm:
Thanks for merging! :smile: For future reference, documenting the conversation we had in private chat: 
- `problems.yaml` is indeed DOMjudge-only, so it makes sense to only use syntax here that DOMjudge actually support (i.e., force `name: str`, defaulting to the English name)
- There is some duplication between `problem{,s}.yaml`, but not all:
  - `name` is duplicate with `name` (obviously)
  - `timelimit` is duplicate with [`limits.time_limit](https://github.com/Kattis/problem-package-format/blob/master/spec/2023-07-draft.md#limits)
  - `rgb`/`color` and `label` are DOMjudge-specific, so we will still need to keep `problems.yaml` around for at least these fields. In the future (i.e., once the Kattis spec is final and implemented in DOMjudge), we can remove `name` and `timelimit` from `problems.yaml`.

PR #299, opened 2023-08-05
[problem] Rename `Problem.name` to `Problem.id`

mpsijm:
The `Problem.name` field left me confused just a bit too often, as I thought it would contain the human-readable name of the problem instead of the problem's ID/"short name" :innocent:

### DIFF
--- a/bin/export.py
+++ b/bin/export.py
@@ -49,9 +49,7 @@ def build_samples_zip(problems):
 
         attachments_dir = problem.path / 'attachments'
         if problem.interactive and not attachments_dir.is_dir():
-            util.error(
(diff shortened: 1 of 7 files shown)


thorehusfeldt:
Related recent discussion at https://github.com/Kattis/problem-package-format/issues/99#event-9957209194 . There is more-or-less up-to-date terminology on “my” fork: https://problem-package-format-copy.readthedocs.io/en/latest/format.html#directory-structure .

I think the consensus it refer to this concept (which also give the root-level directory name) the _(problem) package name_.  So an alternative would be `Problem.package_name` .

PR #300, opened 2023-08-29
CUE schema in feat/cue

thorehusfeldt:
Significantly rewritten CUE schema for `generators.yaml`.

Fixes:
1. use # convention for schema structs
2. rename generator to #testcase, directory to #testgroup
3. allow in, ans, copy, hint, desc as keys in #testcase
4. restrict testgroups below data to sample, secret, invalid_inputs
5. specify contents of `testdata.yaml` in separate schema (even though this may soon change and be much simplified.)

Contentious:
i. Disallow empty/null value for `solution` and `#testgroup`, fix example `generators.yaml` accordingly

Missing:
<s>a. specify `include`</s>
b. add a bash script that `cue vet`s all `generators.yaml` files in the repo against this schema.


RagnarGrootKoerkamp:
Thanks! Some remaining TODOs are in #301

PR #302, opened 2023-09-02
Update and improve JSON schema

thorehusfeldt:
* Specify generators (note: disallowing `c.py` as a name)
* Allow `data/invalid_inputs`
* Disallow `null` for `solution`
* Disallow empty testcase
* More correct example paths (in particular `/`)

Improved titles and descriptions

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -6,4 +6,5 @@ jobs:
     container: ragnargrootkoerkamp/bapctools
     steps:
       - uses: actions/checkout@v2
+      - run: cue vet doc/generators.yaml support/schemas/*cue -d '#Generators'
       - run: pytest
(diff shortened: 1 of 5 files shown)


thorehusfeldt:
I have added the required regexes to the JSON schema, which makes the editing experience a lot better! (Should have done this earlier, clear usability improvement.)

Hoping that we converge on disallowing `c.py` as a generator name I have also unified all the names to `name`. In particular, testgroups, testcases, and generators all follow the same naming convention. (No `.`, and single-letter strings allowed, otherwise like `filename`.)

RagnarGrootKoerkamp:
Yes, this looks good! So both the CUE and JSON are now consistent on generator/testcase/testgroup names not containing dots, but allowing single-char names. That sounds very reasonable to me :+1:

thorehusfeldt:
When you merge, note that `doc/generators.yaml` should show you an error in your editor (because `c.py`) and `cue` should complain:

```bash
BAPCtools git:(ragnar) ✗ cue vet doc/generators.yaml support/schemas/*.cue -d "#Generators"
generators."c.py": field not allowed:
    ./doc/generators.yaml:70:4
    ./support/schemas/generators.cue:56:14
    ./support/schemas/generators.cue:60:15
 ```
(So to finalise this PR, I suggest you reproduce the above on your machine, fix `doc/generators.yaml` and _then_ merge. Just to ensure that the schemata actually _do_  anything.)

RagnarGrootKoerkamp:
https://github.com/RagnarGrootKoerkamp/BAPCtools/actions/runs/6064144214/job/16452257911 success

PR #306, opened 2023-09-13
[parallel] If not num_threads, use queue instead of tail recursion

mpsijm:
Previously, running `bt fuzz --jobs 0` would crash with a stack overflow after 244 iterations (±4 stack frames per iteration). This happened because after every GeneratorTask, a new task was started by calling Parallel.put, which in turn called its lambda, which called the new GeneratorTask, which called Parallel.put again in finish_task, etc.

I've rewritten the task handling in the case of `not num_threads` by moving it all to Parallel.done. Parallel.put now puts the task in the same priority queue as in the case where we _do_ want parallelism (num_threads > 0). The fuzzer should now be able to run more than 244 iterations again 😄

I created a PR instead of simply pushing to `master`, since I haven't touched all of this parallel stuff that often, so some extra eyes would be nice :slightly_smiling_face: 

<details>
<summary>To quickly reproduce the bug with `bt fuzz -j0`, I added a minimal generator to the test problem "hellowholeworld", which makes 244 iterations take about 2 minutes:</summary>

```diff
diff --git a/test/problems/hellowholeworld/generators/gen.py b/test/problems/hellowholeworld/generators/gen.py
new file mode 100644
index 0000000..829afc9
--- /dev/null
+++ b/test/problems/hellowholeworld/generators/gen.py
@@ -0,0 +1,6 @@
+#!/usr/bin/env python3
+import random
+import sys
+
+random.seed(sys.argv[1])
+print(random.randint(1, 100))
diff --git a/test/problems/hellowholeworld/generators/generators.yaml b/test/problems/hellowholeworld/generators/generators.yaml
new file mode 100644
index 0000000..2f90cce
--- /dev/null
+++ b/test/problems/hellowholeworld/generators/generators.yaml
@@ -0,0 +1,9 @@
+solution: /submissions/accepted/test-hello.py
+data:
+  sample:
+    data:
+      "1":
+        in: "1"
+  secret:
+    data:
+      - "": gen.py {seed}
```

</details>

### DIFF
--- a/bin/fuzz.py
+++ b/bin/fuzz.py
@@ -88,9 +88,8 @@ class GeneratorTask:
                     return False
                 localbar.done()
             else:
-                if not target_ansfile.is_file():
-                    bar.error(f'{self.i}: {ansfile.name} does not exist and was not generated.')
-                    return False
+                bar.error(f'{self.i}: {ansfile.name} was not generated.')
+                return False
         else:
             if not testcase.ans_path.is_file():
                 testcase.ans_path.write_text('')
--- a/bin/fuzz.py
+++ b/bin/fuzz.py
@@ -188,25 +188,16 @@ class Fuzz:
         self.tasks = 0
         self.queue = parallel.Parallel(lambda task: task.run(bar), pin=True)
 
-        if self.queue.num_threads:
-            # pool of ids used for generators
-            self.tmp_ids = 2 * max(1, self.queue.num_threads) + 1
-            self.free_tmp_id = {*range(self.tmp_ids)}
-            self.tmp_id_count = [0] * self.tmp_ids
+        # pool of ids used for generators
+        self.tmp_ids = 2 * max(1, self.queue.num_threads) + 1
+        self.free_tmp_id = {*range(self.tmp_ids)}
+        self.tmp_id_count = [0] * self.tmp_ids
 
-            # add first generator task
-            self.finish_task()
+        # add first generator task
+        self.finish_task()
 
-            # wait for the queue to run empty (after config.args.time)
-            self.queue.join()
-        else:
-            self.tmp_ids = -1
-            while time.monotonic() - self.start_time <= config.args.time:
-                testcase_rule = self.testcase_rules[self.iteration % len(self.testcase_rules)]
-                self.iteration += 1
-                self.queue.put(GeneratorTask(self, testcase_rule, self.iteration, None))
-                self.queue.join()
-
+        # wait for the queue to run empty (after config.args.time)
+        self.queue.join()
         # At this point, no new tasks may be started anymore.
         self.queue.done()
         bar.done()
--- a/bin/fuzz.py
+++ b/bin/fuzz.py
@@ -186,7 +186,7 @@ class Fuzz:
         self.start_time = time.monotonic()
         self.iteration = 0
         self.tasks = 0
-        self.queue = parallel.Parallel(lambda task: task.run(bar), pin=True)
+        self.queue = parallel.new_queue(lambda task: task.run(bar), pin=True)
 
         # pool of ids used for generators
         self.tmp_ids = 2 * max(1, self.queue.num_threads) + 1
--- a/bin/parallel.py
+++ b/bin/parallel.py
@@ -121,30 +121,24 @@ class Parallel:
 
     # Add one task. Higher priority => done first
     def put(self, task, priority=0):
+        # no task should be added after .done() was called
+        assert not self.finish
+
+        # no task will be handled after self.abort so skip adding
+        if self.abort:
+            return
+
         if not self.num_threads:
-            # no task should be added after .done() was called
-            assert not self.finish
-            # no task will be handled after self.abort
-            if not self.abort:
-                if self.pin:
-                    cores = list(os.sched_getaffinity(0))
-                    os.sched_setaffinity(0, {cores[0]})
-                    self.f(task)
-                    os.sched_setaffinity(0, cores)
-                else:
-                    self.f(task)
+            self.total_tasks += 1
+            heapq.heappush(self.tasks, ParallelItem(task, priority, self.total_tasks))
             return
 
         with self.mutex:
-            # no task should be added after .done() was called
-            assert not self.finish
-            # no task will be handled after self.abort so skip adding
-            if not self.abort:
-                # mark task as to be done and notify workers
-                self.missing += 1
-                self.total_tasks += 1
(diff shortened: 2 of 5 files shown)


bin/parallel.py
```diff
@@ -121,30 +121,24 @@ def _interrupt_handler(self, sig, frame):
 
     # Add one task. Higher priority => done first
     def put(self, task, priority=0):
+        # no task should be added after .done() was called
+        assert not self.finish
```
mzuenni, line 125: This is now outside of the mutex in the parallel case, which could lead to race conditions?
(same with the abort check) I think its cleaner if it remains inside the locked environment in the parallel case.

bin/fuzz.py
```diff
@@ -55,8 +55,7 @@ def _run(self, bar):
         localbar = bar.start(f'{self.i}: generate')
         result = self.generator.run(bar, cwd, name, self.seed)
```
RagnarGrootKoerkamp, line 56: Should we pass `localbar` here? (not rhetoric)

bin/parallel.py
```diff
@@ -158,6 +152,18 @@ def join(self):
...
+                    cores = list(os.sched_getaffinity(0))
+                    os.sched_setaffinity(0, {cores[0]})
+                    self.f(task.task)
+                    os.sched_setaffinity(0, cores)
+                else:
+                    self.f(task.task)
+
         self.finish = True
 
         if not self.num_threads:
```
RagnarGrootKoerkamp, line 169: Can you merge this with the `if` above, and duplicate the `self.finish = true`

bin/parallel.py
```diff
@@ -158,6 +152,18 @@ def join(self):
...
+                task = heapq.heappop(self.tasks)
+                if self.pin:
+                    cores = list(os.sched_getaffinity(0))
+                    os.sched_setaffinity(0, {cores[0]})
+                    self.f(task.task)
+                    os.sched_setaffinity(0, cores)
+                else:
+                    self.f(task.task)
+
         self.finish = True
```
RagnarGrootKoerkamp, line 167: Hrmmm; so in the parallel case:
- jobs start running immediately after the first `push`
- jobs can add additional jobs
- when calling `done()`, `finish` is set to true and immediately no more jobs can be added.

But now in the non-parallel case:
- jobs start on `done`
- these jobs can add more jobs
- only after they're all done, `finish` is set to true.

Probably this is OK? At least this should be documented clearly in a comment here.

bin/parallel.py
mpsijm: Would it make sense to wrap _everything_ inside the mutex? In the non-parallel case, this would be a no-op. Since there's a lot of duplication, this could clean things up a bit (although I will definitely add a comment about the race conditions) :slightly_smiling_face:

bin/fuzz.py
mpsijm: Oh huh, that is a good question :thinking: `bar.start` returns a copy, so probably yes?

bin/parallel.py
mpsijm: Yes, that is indeed the intention :slightly_smiling_face: This hangs on the assumption that no jobs will be added after the call to `Parallel.done()` anyway, which I think is a fair assumption to make :stuck_out_tongue: 

Documenting this is indeed a good idea, thanks :smile:

mpsijm:
Better late than never :smile: 

While trying to resolve your feedback, I came to the conclusion that having one class that's doing very different things in the sequential vs. the parallel case is not very maintainable. So, I decided to split the two functionalities into two subclasses of a common abstract superclass.

The consequence of this is that @mzuenni's fix for the fuzzer in #327 is no longer necessary: the user of `parallel.create` can always transparently treat it as if it was a parallel queue, even though it's secretly a sequential queue when `args.jobs == 0` :smile:

bin/fuzz.py
mpsijm: Fixed in #327: https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/327/files#diff-445d79520fde92e79b3572d53d445cf7c57b3d3c58e98aef451519a2c2c766beR56

bin/fuzz.py
```diff
@@ -86,9 +86,8 @@ def _run(self, bar):
                     return False
                 localbar.done()
             else:
-                if not target_ansfile.is_file():
-                    bar.error(f'{self.i}: {ansfile.name} does not exist and was not generated.')
-                    return False
+                bar.error(f'{self.i}: {ansfile.name} does not exist and was not generated.')
```
RagnarGrootKoerkamp, line 89: Update message `{ansfile.name} was not generated.`

bin/fuzz.py
```diff
@@ -185,27 +184,18 @@ def run(self):
...
         self.iteration = 0
         self.tasks = 0
-        self.queue = parallel.Parallel(lambda task: task.run(bar), pin=True)
-
-        if self.queue.num_threads:
-            # pool of ids used for generators
-            self.tmp_ids = 2 * max(1, self.queue.num_threads) + 1
-            self.free_tmp_id = {*range(self.tmp_ids)}
-            self.tmp_id_count = [0] * self.tmp_ids
+        self.queue = parallel.create(lambda task: task.run(bar), pin=True)
```
RagnarGrootKoerkamp, line 187: `create` is not a very descriptive name. Better `new` or `new_queue` or so

bin/parallel.py
```diff
@@ -26,74 +25,119 @@ def __lt__(self, other):
...
+        # no task will be handled after self.abort() so skip adding
+        if self.aborted:
+            return
+
+        self.total_tasks += 1
+        heapq.heappush(self.tasks, QueueItem(task, priority, self.total_tasks))
+
+    # Execute all tasks.
+    def done(self):
+        if self.pin:
```
RagnarGrootKoerkamp, line 80: I don't think pinning really makes much sense for the sequential queue, so these `if`s here can just be dropped I'd say

RagnarGrootKoerkamp (approved):
Mostly LGTM; just some small comments.

Still not entirely convinced we actually need this but happy to merge either way.

bin/fuzz.py
mpsijm: Done!

bin/fuzz.py
mpsijm: Fair point, renamed to `new_queue` :slightly_smiling_face:

bin/parallel.py
mpsijm: I do think this makes sense actually, because I want my timings to be consistent, especially in sequential mode :stuck_out_tongue:

RagnarGrootKoerkamp (approved):

PR #327, opened 2023-12-21
[parallel] If not num_threads dont use tail recursion

mzuenni:
It fixes the same problem as @mpsijm in #306, but I think the problem is in fuzz and not how the queue behaves.

### DIFF
--- a/bin/fuzz.py
+++ b/bin/fuzz.py
@@ -187,16 +187,25 @@ class Fuzz:
         self.tasks = 0
         self.queue = parallel.Parallel(lambda task: task.run(bar), pin=True)
 
-        # pool of ids used for generators
-        self.tmp_ids = 2 * max(1, self.queue.num_threads) + 1
-        self.free_tmp_id = {*range(self.tmp_ids)}
-        self.tmp_id_count = [0] * self.tmp_ids
+        if self.queue.num_threads:
+            # pool of ids used for generators
+            self.tmp_ids = 2 * max(1, self.queue.num_threads) + 1
+            self.free_tmp_id = {*range(self.tmp_ids)}
+            self.tmp_id_count = [0] * self.tmp_ids
 
-        # add first generator task
-        self.finish_task()
+            # add first generator task
+            self.finish_task()
 
-        # wait for the queue to run empty (after config.args.time)
-        self.queue.join()
+            # wait for the queue to run empty (after config.args.time)
+            self.queue.join()
+        else:
+            self.tmp_ids = -1
+            while time.monotonic() - self.start_time <= config.args.time:
+                testcase_rule = self.testcase_rules[self.iteration % len(self.testcase_rules)]
+                self.iteration += 1
+                self.queue.put(GeneratorTask(self, testcase_rule, self.iteration, None))
+                self.queue.join()
+
         # At this point, no new tasks may be started anymore.
         self.queue.done()
         bar.done()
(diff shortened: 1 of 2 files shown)


bin/fuzz.py
```diff
@@ -188,16 +187,25 @@ def run(self):
...
-        self.finish_task()
+            # wait for the queue to run empty (after config.args.time)
+            self.queue.join()
+        else:
+            self.tmp_ids = -1
+            while time.monotonic() - self.start_time <= config.args.time:
+                testcase_rule = self.testcase_rules[self.iteration % len(self.testcase_rules)]
+                self.iteration += 1
+                self.queue.put(GeneratorTask(self, testcase_rule, self.iteration, None))
+                self.queue.join()
```
RagnarGrootKoerkamp, line 207: Is there a cleaner solution that avoids the queue alltogether? Or is this slightly hacky workaround the shortest way of writing it? (It's OK as is, but slightly weird.)

bin/fuzz.py
```diff
@@ -188,16 +187,25 @@ def run(self):
...
+
+            # add first generator task
+            self.finish_task()
 
-        # add first generator task
-        self.finish_task()
+            # wait for the queue to run empty (after config.args.time)
+            self.queue.join()
+        else:
+            self.tmp_ids = -1
```
RagnarGrootKoerkamp, line 202: Comment here that this is the single-threaded case. Maybe this should also include num_threads=1? Not sure.

RagnarGrootKoerkamp (commented):
Thanks! Mostly lgtm

bin/fuzz.py
mpsijm: A cleaner solution imo is still my fix in #306. This PR looks more like a workaround for the usage of `Parallel` in the fuzzer, but if any other part of BAPCtools ever runs into a stack overflow when using `Parallel`, we'd have to re-implement a similar fix there again. But in #306, users of `Parallel` don't have to think about the difference between the single-threaded and the multi-threaded case :slightly_smiling_face: 

(and yes, I will spend some time soon™ to resolve your comments at #306 and test things through :innocent:)

mzuenni: But the bug is not in Parallel the bug is the tail recursion in fuzz? The parallel class works exactly as intended IMO.

bin/fuzz.py
mzuenni: `num_threads=1` should work fine since we have one thread (besides the main thread locked by `join`) that handles elements in the queue

bin/fuzz.py
RagnarGrootKoerkamp: Hmm either way works.
The Parallel class isn't really broken, and to me it feels like doing the work in 'put' is cleaner than doing the work in 'done', but it would be nice if you never have to check `if self.queue.num_threads` manually outside `Parallel`

mzuenni: Yeah, but the bug occurs just because of the special way I implemented the parallel fuzz... Maybe there is a cleaner way to do this... but i think this is the best fix for now

RagnarGrootKoerkamp:
I'll just merge this and then we can merge #306 later if we feel like it.

bin/fuzz.py
RagnarGrootKoerkamp: Oh I merged now but my point here was still open: Can you avoid the queue and just call GeneratorTask.run() or whatever directly?

mzuenni: the GeneratorTask will write to the queue anyway... This is my best solution right now

PR #328, opened 2024-01-01
display more debug paths

mzuenni:
This changes the following things:

- When generation fails clearly display what went wrong i.e. solution,visualization,generation...
- When generation fails and  `-v` is given display a shortened path to the tmp directory where generated stuff is
- Add the rule to the `meta_.yaml` to make it easier to find the failing rule in the `generator.yaml` (if auto numbering is used and multiple cases have the same base name this can be hard)

### DIFF
--- a/bin/generate.py
+++ b/bin/generate.py
@@ -95,6 +95,7 @@ class Invocation:
         commands = string.split()
         command = commands[0]
         self.args = commands[1:]
+        self.problem = problem
 
         # The original command string, used for caching invocation results.
         self.command_string = string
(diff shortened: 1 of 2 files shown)


bin/generate.py
```diff
@@ -170,9 +171,11 @@ def run(self, bar, cwd, name, seed, retries=1):
 
         if result.ok is not True:
             if retries > 1:
-                bar.error(f'Failed {retry + 1} times', result.err)
+                bar.debug(f'{Style.RESET_ALL}-> {shorten_path(self.problem, cwd)}')
+                bar.error(f'generator failed {retry + 1} times', result.err)
             else:
-                bar.error(f'Failed', result.err)
+                bar.debug(f'{Style.RESET_ALL}-> {shorten_path(self.problem, cwd)}')
+                bar.error(f'generator failed', result.err)
```
RagnarGrootKoerkamp, line 178: Can you consistently capatalize these?

RagnarGrootKoerkamp (approved):
looks quite nice! One small nit and then I'll merge :)

Issue #326, opened 2023-12-20
Warn if \problemtitle differs from problem.settings.name[language]


thorehusfeldt:
When `problem.en.tex` contains
```tex
\problemname{Hello Gaia!}
```
but `problem.yaml` contains
```
name:
  en: Hello World!
```
there currently is no warning. BAPC instead expects the author to write
```tex
\problemname{\problemyamlname}
```
but there can be reasons to avoid that macro, typically due to other tooling. (In particular, it is one of the few details that prevents a BAPC-compatible problem package from being `problemtools/verifyproblem`-ed.)

There is extensive parsing of TeX-code already being done in `bt constraints`, but the above check could be done quite easily and catch 99% of cases.

A more aggressive solution would be to actually edit `problem.en.tex` after `bt new_problem` and `bt rename_problem`.

RagnarGrootKoerkamp:
Sounds reasonable to add a warning/log for this indeed.

I think the most common usecase for this is when the latex name includes additional markup that we don't want to expose in the yaml. But it's probably fine to just warn for these cases anyway since accidental mismatches are worse than a few extra warnings.

mzuenni:
We could also check if the expected string is a substring of the actual string... That would reduce the number of warnings but would also suppress the warning for some types of typos.

mzuenni:
It would not war for markup like this: `\texttt{IX}th Problem` because `IXth Problem` is still a substring. On the other hand it would no longer warn for `IXth Poblem` or stuff like that...

thorehusfeldt:
Ah, _that_ kind of substring. Now it makes sense.

thorehusfeldt:
Minimal solution is now a concrete pull request, please comment there.

PR #329, opened 2024-01-15
326 warn if problemtitle differs from problemsettingsnamelanguage

thorehusfeldt:
For #326, compare problemtitle in `problem.yaml` and various `{lang}.tex` files.

The extraction of `\problemname{foo}` in `problem.en.tex` is handled by a new method in `bin/latex.py`. This is very bare-bones and would be confused by each of the following

```tex
% \problenname{No warning produced, though outcommented}
```

```tex
\problemname{
   foo
}
```
Somebody should run this on a large repository of existing problems and see how many false positive warnings it generates.

It warns if no `problemname` is found in the tex source, which makes the BAPCtools testsuite unhappy. Maybe that warning shouldn’t be there. On the other hand, the specification actually requires it:

https://github.com/Kattis/problem-package-format/blob/af5288b5c21d8a65ace08296ce7aecad8fdefdd2/spec/2023-07-draft.md?plain=1#L282

(and in the first line of the code!) so maybe it should be much stricter.

### DIFF
--- a/bin/problem.py
+++ b/bin/problem.py
@@ -78,7 +78,7 @@ class Problem:
             with open(self.path / 'problem_statement' / f'problem.{lang}.tex') as texfile:
                 match texname := latex.get_argument_for_command(texfile, 'problemname'):
                     case None:
-                        warn(rf"No \problemname found in problem.{lang}.tex")
+                        error(rf"No \problemname found in problem.{lang}.tex")
                         continue
                     case '\problemname':
                         continue
--- a/bin/problem.py
+++ b/bin/problem.py
@@ -58,19 +58,19 @@ class Problem:
         """
         if isinstance(self.settings.name, str):
             self.settings.name = {'en': self.settings.name}
-        yamlnames = set(self.settings.name)
-        texfiles = set(
+        yamllangs = set(self.settings.name)
+        texlangs = set(
             path.suffixes[0][1:] for path in glob(self.path, 'problem_statement/problem.*.tex')
         )
-        for lang in texfiles - yamlnames:
+        for lang in texlangs - yamllangs:
             error(
                 f"{self.name}: Found problem.{lang}.tex, but no corresponding name in problem.yaml."
             )
-        for lang in yamlnames - texfiles:
+        for lang in yamllangs - texlangs:
             error(
                 f"{self.name}: Found name for language {lang} in problem.yaml, but not problem.{lang}.tex."
             )
-        return sorted(texfiles & yamlnames)
+        return sorted(texlangs & yamllangs)
 
     def _read_settings(self):
         # some defaults
--- a/bin/problem.py
+++ b/bin/problem.py
@@ -7,6 +7,7 @@ import sys
 from pathlib import Path
 
 import config
+import latex
 import parallel
 import program
 import run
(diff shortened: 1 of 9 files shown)


bin/problem.py
```diff
@@ -58,19 +59,31 @@ def _determine_statement_languages(self):
...
             error(
                 f"{self.name}: Found problem.{lang}.tex, but no corresponding name in problem.yaml."
             )
-        for lang in yamlnames - texfiles:
+        for lang in yamllangs - texlangs:
             error(
                 f"{self.name}: Found name for language {lang} in problem.yaml, but not problem.{lang}.tex."
             )
-        return sorted(texfiles & yamlnames)
+        for lang in texlangs & yamllangs:
```
RagnarGrootKoerkamp, line 74: Could you add a quick comment above here that this for loop is for checking problem name consistency?

bin/problem.py
```diff
@@ -58,19 +59,31 @@ def _determine_statement_languages(self):
...
+            unnormalised_yamlname = self.settings.name[lang]
+            yamlname = ' '.join(unnormalised_yamlname.split())
+            with open(self.path / 'problem_statement' / f'problem.{lang}.tex') as texfile:
+                match tex_problem_name := latex.get_argument_for_command(texfile, 'problemname'):
+                    case None:
+                        warn(rf"No \problemname found in problem.{lang}.tex")
+                    case s if s not in ["\\problemyamlname", yamlname]:
```
RagnarGrootKoerkamp, line 81: As discussed in the issue, I think it's more reasonable to ensure that the yamlname is a substring of the texname.

The typical reason they differ is because of added typesetting, which shouldn't appear in the yamlname.

RagnarGrootKoerkamp (commented):
mostly lgtm. Just 2 tiny things.

And indeed, would be nice if you could fix the empty `problem.en.tex` in `tests/` so that tests pass again.

RagnarGrootKoerkamp:
For us we pretty much always use `\problemyamlname` unless the formatting is different, but maybe you're right.

We could also allow some way to disable the check, like `\problemname{something different} % DISTINCTNAME`. Then the check can be strict and users can simply disable the warning.

@mpsijm @mzuenni wdyt?

thorehusfeldt:
Another idea is to disable the check as soon as `texname` contains `\`.

RagnarGrootKoerkamp:
Oh, I like that! Feel free to implement.

thorehusfeldt:
Can someone grep `\\problemname\{^\\}`(or something like that) in the full BAPC- and NWERC-library? Then we could get a better feeling for what kind of formatting actually appears in real life.

RagnarGrootKoerkamp:
For BAPC we have:
- `\problemname{\texttt{git mv}}`
- `\problemname{\texttt{apt upgrade}}`

For NWERC we didn't deviate in the past 3 years.

bin/problem.py
thorehusfeldt: Solved differently, see below

thorehusfeldt:
I did toy with the idea of removing matching `\foo{` … `}` from `\problemtitle` but I don’t think this issue deserves that much attention. I think I’m done with my TODOs now.

RagnarGrootKoerkamp:
> I did toy with the idea of removing matching `\foo{` … `}` from `\problemtitle` but I don’t think this issue deserves that much attention.

Agreed.
> I think I’m done with my TODOs now.

> Do note that [62a41b1](https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/329/commits/62a41b1cfe364f51c13da9a356bd73875c30fc2c) escalated `warn` to `error`, which is consistent with the specification (and now also with BAPCtools’s test suite.)

:+1: 

Thanks! Merging :)

PR #330, opened 2024-01-15
Answer Format Validation

thorehusfeldt:
This addresses the issue of separating output validation from validation of the format of the default answer file. Very much work in progress, but worth looking at (and commenting on.)


test/problems/fltcmp/input_validators/validate.py
```diff
@@ -0,0 +1,19 @@
+#! /usr/bin/env python3
+
+import re
+import sys
+
+line = sys.stdin.readline()
+if not re.match("([1-9][0-9]*)\n", line):
+    print("Positive integer expected", file=sys.stderr)
+    sys.exit(43)
```
RagnarGrootKoerkamp, line 9: something is weird here. Either the sys.exit should go or the code directly below. I suspect the below should not be indented?

bin/problem.py
```diff
@@ -343,56 +343,58 @@ def build_program(p):
...
     def validators(problem, validator_type, check_constraints=False):
-        assert validator_type in ['input_format', 'output_format', 'output']
-
-        # For custom validation, treat 'output' and 'output_format' validators the same.
-        if problem.settings.validation != 'default' and validator_type == 'output_format':
-            validator_type = 'output'
+        """
+        Args:
+            validator_type: 'answer', 'output', 'input'
+            check_constraints: True if the validao
```
RagnarGrootKoerkamp, line 349: incomplete comment

bin/problem.py
```diff
@@ -343,56 +343,58 @@ def build_program(p):
...
+                }
+        paths_for_type = {
+                vtype: [g for sdir in sdirs for g in glob(problem.path / sdir, '*')]
+                for vtype, sdirs in subdirs_for_type.items()
+                }
+
+        # Handle default output validation
+        if problem.settings.validation == 'default':
+            if paths_for_type['output']:
+                error("Validation is default but custom output validator exists")
```
RagnarGrootKoerkamp, line 376: Either make this fatal or include that it is ignored maybe?

bin/problem.py
```diff
@@ -343,56 +343,58 @@ def build_program(p):
...
+                ok = False
```
RagnarGrootKoerkamp, line 377: Hmm; so in this case we do build the default output validator but then anyway end up returning `False`. Can we just bail early? Or do you think it's ugly to have to repeatedly update the cache to `False` here and elsewhere we have `ok = False`?
(Or maybe the code following here already doesn't really do much anyway once `ok=False` so it's fine.)

bin/problem.py
```diff
@@ -343,56 +343,58 @@ def build_program(p):
...
+            paths_for_type['output'] = [ config.tools_root / 'support' / 'default_output_validator.cpp' ]
+
+        paths = paths_for_type[validator_type]
+
+
+        # Check that the proper number of validators is present
+        match validator_type, len(paths):
```
RagnarGrootKoerkamp, line 384: oh I never knew `match` is a thing; cool!
This is very clean now :)

bin/problem.py
```diff
@@ -419,38 +421,33 @@ def has_constraints_checking(f):
...
-                    check_constraints=check_constraints,
-                )
-                for path in paths
-            ]
+        validator_dispatcher = {
+                'input': validate.InputValidator,
+                'answer': validate.AnswerValidator,
+                'output': validate.OutputValidator
+                }
+        skip_double_build_warning = check_constraints or not paths_for_type['answer']
```
RagnarGrootKoerkamp, line 433: I think the `check_constraints or` part of this actually isn't needed anymore. This was needed because I was passing some custom `-Dxyz` flags to the C++ compiler for constraints checking but those are now not used anymore.

RagnarGrootKoerkamp (commented):
Looks very nice!
Only some small comments. Overall I see many improvements to code clarity and simplicity :tada:

test/problems/fltcmp/input_validators/validate.py
thorehusfeldt: fixed, thank you

bin/problem.py
thorehusfeldt: Decision: Tell user we’re ignoring their validator

bin/problem.py
thorehusfeldt: I didn’t want to touch this.

RagnarGrootKoerkamp: Fair enough; I'll make an issue to clean this up later.

bin/problem.py
thorehusfeldt: We _could_ bail early, but one of the first things I did to the code (in particular because I wanted to make sure the caching worked) is to only return at the end. It’s prettier and safer this way.

doc/validation.md
```diff
@@ -2,11 +2,20 @@
 
 BAPCtools does 3 types of validation, one more than the spec (https://www.kattis.com/problem-package-format/spec/problem_package_format#input-validators):
 
+1. input validation, which validates the `.in` file for every test case, typically for syntax, format, and ranges
```
RagnarGrootKoerkamp, line 5: add correctness here? Or some other word that contains e.g. checking that a graph is connected.

doc/validation.md
```diff
@@ -2,11 +2,20 @@
...
+2. answer validation, which validates the `.ans` file for every test case, typically for syntax, format, and ranges, but not for correctness
+3. ouput validation, which checks correctness and running time for the output of every submission.
```
RagnarGrootKoerkamp, line 7: add that this is only available/allowed when `validation` is not `default`

doc/validation.md
```diff
@@ -2,11 +2,20 @@
...
+
+Input and answer validation run on the  _files_ in `data/*`; their purpose is to ensure problem quality.
+Output validation runs on the output of the author submissions in `submission` (and eventually on solver submissions when the problem is hosted on a judge system);
```
RagnarGrootKoerkamp, line 10: `submissionS`, plural

doc/validation.md
```diff
@@ -2,11 +2,20 @@
...
+the purpose of output validation is to check correctness of submitted problems.
```
RagnarGrootKoerkamp, line 11: corectness of submissions?

doc/validation.md
```diff
@@ -2,11 +2,20 @@
...
+
+
 ## Common parameters/settings
 
 These are some things that hold for all types of validation mentioned below.
 
-- For each testcase, all validators are run in lexicographic order. If one
+- For each testcase, all validators in the same directory are run in lexicographic order. If one
```
RagnarGrootKoerkamp, line 18: 'of the same type'?

doc/validation.md
```diff
@@ -19,90 +28,82 @@ These are some things that hold for all types of validation mentioned below.
...
-input_validator <input_validator_flags> [--contraints_file <path>] < <testcase>.in
+input_validator [input_validator_flags] < testcase.in
 ```
 
-- The testcase input is given on `stdin`.
+## Answer validation
 
-## Output format validation (out-of-spec)
+BAPCtools allows (in fact, encourages) the validation of the `.ans`-file of each testcase.
+As for input validatio, every program in `answer_validators` is a validator, and all validator must pass.
```
RagnarGrootKoerkamp, line 52: `validator`

doc/validation.md
```diff
@@ -19,90 +28,82 @@ These are some things that hold for all types of validation mentioned below.
...
+Answer validators receive the testcase answer file on standard input, as
+```
+answer_validator /path/to/testcase.in [output_validator_flags] < tc.ans
```
RagnarGrootKoerkamp, line 55: `testcase.ans`

doc/validation.md
```diff
@@ -19,90 +28,82 @@ These are some things that hold for all types of validation mentioned below.
...
-  that `stdin` contains a single integer. Slightly more advanced would be read
-  `n` from the `.in` file (first argument). Then read `n` lines from `stdin`
-  and verify that they contain integers separated by newlines.
-- Custom validation (`validation: custom`): Validate the `.ans` file on
-  `stdin` as you would validate a team submission. This assumes that the
-  `.ans` file in indeed present in the `data/` directory and a valid solution
-  to the problem. While this is not technically required otherwise, this has
-  never been false not the case. See 'output validation' below for an example.
+In particular, note the flags `case_sensitive` and `space_change_sensitive`, 
+which allows an output validator to be less lenient about the format of `.ans` files than about submission output.
```
RagnarGrootKoerkamp, line 70: less lenient => more strict

RagnarGrootKoerkamp (commented):
Just some minor commends on the docs; code all lgtm :)

doc/validation.md
thorehusfeldt: Could you rephrase? I’m not sure what you mean.

test/problems/hello/parked/test-exitcode.c
mpsijm, line 1: Why are these three submissions parked?

doc/validation.md
```diff
@@ -19,90 +31,82 @@ These are some things that hold for all types of validation mentioned below.
...
 ```
-input_validator <input_validator_flags> [--contraints_file <path>] < <testcase>.in
+input_validator [input_validator_flags] < testcase.in
 ```
 
-- The testcase input is given on `stdin`.
+## Answer validation
 
-## Output format validation (out-of-spec)
+BAPCtools allows (in fact, encourages) the validation of the `.ans`-file of each testcase.
```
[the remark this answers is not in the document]
thorehusfeldt: There’s even more in `/skel`; here’s a partial list:

1.  `/data` should die
2. the `.ctd` and `.viva` templates in `output_validators` are now `answer_validators`
3. generators/stdout.py  is obsolete

test/problems/hello/parked/test-exitcode.c
thorehusfeldt: My mistake. Will fix.

doc/validation.md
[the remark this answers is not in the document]
thorehusfeldt: I think updating `/skel` is a separate Issue (because there are also details resulting from `generators` and from multiple language support.)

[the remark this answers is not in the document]
thorehusfeldt: Started a collection in #332

doc/validation.md
thorehusfeldt: Yes, I tried to be precise but maybe this needs to be even better.

Answer validation _always_ happens during `bt validate`; and it _always_ uses the(!) output validator (which must uniquely exist, and _may_ be the default output validator.) If there are answer validators, these are also run during answer validation.  Maybe it’s more stringent to just say “`bt answer`” (when that command exists, maybe in a few days) instead of using the term _answer validation_.

Some sentences that use these terms:
1. Answer validation is a _process_ (namely, the one triggered by `bt validate`). Answer validators are _programs_ (namely, the ones residing in `answer_validators`). 
2. Answer validation involves the answer validators and the output validator. 
3. Output validation involves the output validator.

Output validation happens during `bt run` (and when the problem is _live_).

RagnarGrootKoerkamp:
Merged :tada: 
Thanks again :)

PR #336, opened 2024-02-13
update skel

### LINKED ISSUE #332: Update problem skeleton
The problem skeleton in `/skel` is getting a bit long in the tooth. Here are some issues:

1. Problem name in `problem.yaml` is not longer up to date with multilingual problem title support: https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/689d1ab993ead6ab7b4c1a91f718708bc0c81e8f/skel/problem/problem.yaml#L1
2. There should be no `data` directory
3. `generators/stdout.py` is obsolete. A good small _random_ generator in python (taking seed) would make sense as an example instead.
4. the `generators/generator/` symlink should just be removed.
5. `generators/generator.yaml` should contain an `invalid` input.
6. there should be an `answer_validators` (presumably filled with just the old `output_validators`

There may be more.

mzuenni:
Fixes #332


skel/problem/generators/example.py
```diff
@@ -0,0 +1,19 @@
...
+import sys
+import random
+
+#Init seed with first argument
+random.seed(int(sys.argv[1]))
+
+#Read the second... arguments and evaluate them as python. Example call:
+#example.py {seed} list(range(3,6)) + [2**i for i in range(5)]
+# =>               [3, 4, 5]        + [1, 2, 4, 8, 16]
+list = eval(' '.join(sys.argv[2:]).encode('ascii').decode('unicode_escape'))
```
mpsijm, line 11: I like the idea of having a canonical Python generator, but I'm not sure if we should advocate the use of `eval` here :stuck_out_tongue: I'm also fine with a generator that simply takes the remaining varargs passed to the generator and shuffles them, for instance:

```py
list = list(map(int, sys.argv[2:]))
```

Also, we probably shouldn't assign to `list`, since that's a built-in function in Python :stuck_out_tongue: (perhaps `values` or `numbers`?)

skel/problem/generators/generator
mpsijm, line 1: If we remove the symlink to the input validator, the examples in the skeleton `generators.yaml` also won't work :upside_down_face: Why not keep this symlink?

skel/problem/problem.yaml
```diff
@@ -1,4 +1,5 @@
-name: {%problemname%}
+name:
+  en: {%problemname%}
```
mpsijm, line 2: This currently generates:
```
name:
  en: {'en': 'Test'}
```
Something should unwrap this somewhere, I think? :stuck_out_tongue:

skel/problem/generators/generators.yaml
```diff
@@ -60,3 +60,36 @@ data:
...
+    # (The output validator is also called with these)
+    data:
+      #'dont_start_with_whitespace':
+      #    in: "<valid testcase>"
+      #    ans: " <valid answer>"
+  invalid_outputs:
+    data:
+      # Add valid testcase with answer and invalid team output to ensure that theout_validator correctly rejects them.
+      # (The output validator is also called with these)
+      data:
```
mpsijm, line 91: I think there's one `data:` too many here?

skel/problem/generators/generators.yaml
```diff
@@ -60,3 +60,36 @@ data:
...
+  invalid_answers:
+    # Add valid testcase with invalid answers to ensure that the answer_validators correctly rejects them.
+    # (The output validator is also called with these)
+    data:
+      #'dont_start_with_whitespace':
+      #    in: "<valid testcase>"
+      #    ans: " <valid answer>"
+  invalid_outputs:
+    data:
+      # Add valid testcase with answer and invalid team output to ensure that theout_validator correctly rejects them.
```
mpsijm, line 89:
```suggestion
      # Add valid testcase with answer and invalid team output to ensure that the output_validator correctly rejects them.
```

skel/problem/generators/generators.yaml
```diff
@@ -60,3 +60,36 @@ data:
       #- large: generator {seed} -n 991
       #- large: generator {seed} -n 999
       #- large: generator {seed} -n 1000
+
+  invalid_inputs:
+    # Add invalid testcase to ensure that input_validators correctly rejects them.
+    data:
+      'empty':
```
mpsijm, line 67: The testcase names don't require quotes, because the names by themselves are valid YAML keys. (This also holds for the commented-out `'dont_start_with_whitespace'`)

skel/problem/generators/example.py
```diff
@@ -0,0 +1,19 @@
...
+
+#Shuffle the list
+random.shuffle(list)
+
+#Print in default format i.e. one line with the number of elements and then 
+#the space separated elements
```
mpsijm, line 17: Can you remove trailing spaces from the comments, and add a space after the `#` sign, to be PEP8-compliant? While I'm here anyway, I have a suggestion for rephrasing this particular comment :slightly_smiling_face: 

```suggestion
# Print in standard format, i.e. one line with the number of elements,
# followed by the space-separated elements.
```

skel/problem/answer_validators/answer_validator/answer_validator.cpp
```diff
@@ -0,0 +1,22 @@
...
+// You should verify the grammar of the answer file.
+// See input_validator.cpp for information on how to use the Validator class.
+// Furthermore you should check simple properties of the answer.
+
+// TODO: Remove these comments, and summarize your answer validator.
+
+int main(int argc, char *argv[]) {
+    // Set up the input and answer streams.
+    std::ifstream in(argv[1]);
+    // std::ifstream ans(argv[2]); // Only for custom checker.
```
mpsijm, line 15: This line should probably be removed from the answer validator, since the answer file is passed via stdin, and not as an argument?

mpsijm (changes requested):
Thank you for updating the skeleton! :smile: Since I'll be setting up for Freshmen Programming Contests soon, I planned to make similar changes, but you beat me to it :slightly_smiling_face: 

I reviewed this by checking out your branch, running `bt new_problem`, and trying to run commands like `bt generate` and `bt run` to see what happens. That resulted in some small issues here and there:

skel/problem/generators/generator
mpsijm: I now see that @thorehusfeldt mentioned this in #332:

> 4. the `generators/generator/` symlink should just be removed.

What was the reasoning behind this again?

mzuenni: I don't know, maybe because reusing a validator as generator is a bit ugly?

skel/problem/problem.yaml
mzuenni: hmm this one is tricky... my best idea right now is to either undo the change or to ensure that new_problem always asks for the english name...

mpsijm: It looks like `bt new_problem` asks for the English name already, sounds good to me :slightly_smiling_face: 

```
$ bt new_problem 
problem name (en): 
```

skel/problem/generators/generator
mpsijm: That does ring a bell. I agree that it's a bit ugly, but I'd still use it to quickly generate an initial set of test data to have a minimum viable problem as quickly as possible. I can always add this if others disagree about keeping the symlink, but then we do need to update `generators.yaml` to not refer to the no-longer-existing symlinked `generator` :stuck_out_tongue:

mzuenni: Right now the `generator.yaml` only mentions that this feature exists and gives examples how to use it. That is still perfectly fine without the link. And i am not too big of a fan myself regarding the fancy features/reusing the validator... for many Problems the feature are not that useful anyway since there are additional constraints like $\sum a_i\leq10^5$ or that it should be a graph etc.

RagnarGrootKoerkamp: I find it quite useful to quickly generate a bunch of cases early on for simple problems, just to have a start. And typically there are some simpler problems where this type of input can cover ~half the cases, which is nice to have without needing any further code.

I could live with dropping the symlink; doesn't matter too much in the end.

mpsijm: > Right now the `generator.yaml` only mentions that this feature exists and gives examples how to use it.

That's true, that's good enough for me then :slightly_smiling_face: 

https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/5cc831a080a6b605c8cadca2862c0a224aee5fcd/skel/problem/generators/generators.yaml#L24-L25

Furthermore, I agree with you both that (1) reusing the input validator as generator will not be satisfactory for the more complicated problems, but (2) for simpler problems it's a huge time saver (and again, for these problems, we can always manually add the symlink :smile:)

Ready for merging in my opinion! :rocket:

mpsijm (approved):

skel/problem/generators/example.py
RagnarGrootKoerkamp: Hmm, I kinda like the `eval` suggestion though. For problems with small input, being able to write `2**10-1 10**6` is really useful and maybe not something users realize themselves. (In that case maybe rename to `eval.py`?)

But again this also doesn't matter too much.

mpsijm: I agree that it can be nice, but again, I don't like having it in the default skeleton because `eval`-ing arbitrary user input isn't really the best idea. If someone likes to use this for their own problems, they can quickly add such a generator themselves (similar to the symlink from `input_validator` to `generator`) :slightly_smiling_face:

mzuenni: I am not sure if `eval`-ing makes it much worse when you already execute random code inside the generator? :p

mpsijm: That's fair :joy:

Either which one we choose, this generator mostly serves as an example (hence the name `example.py`) and is not intended for production use. Its fate may end up being the same as `stdout.py`, which is, being removed or unused for 90% of the problems...

PR #337, opened 2024-02-14
update return code usage

mzuenni:


bin/testcase.py
```diff
@@ -268,7 +268,7 @@ def validate_format(
                 warn_instead_of_error=warn_instead_of_error,
             )
 
-            if ok is True:
+            if ok:
```
RagnarGrootKoerkamp, line 271: hmm just use `ret` here directly?

bin/util.py
```diff
@@ -835,9 +845,22 @@ def _try_wait(self, wait_flags):
...
             return (pid, sts)
 
+def default_exec_code_map(code):
+    if code == 0:
+        return ExecCode.ACCEPTED
+    if code == -9:
+        return ExecCode.TIMEOUT
+    return ExecCode.ERROR
+
+def icpc_exec_code_map(code):
```
RagnarGrootKoerkamp, line 855: Rename to `validator_exec_code_map`?

bin/util.py
```diff
@@ -835,9 +845,22 @@ def _try_wait(self, wait_flags):
...
+    if code == config.RTV_AC:
+        return ExecCode.ACCEPTED
+    if code == config.RTV_WA:
+        return ExecCode.REJECTED
+    return ExecCode.ERROR
```
RagnarGrootKoerkamp, line 860: Hmmm, not sure whether we currently handle timeout for validators and/or whether we should.
Probably it's useful for users to know the error was a validation timeout rather than a generic 'error'?

bin/util.py
```diff
@@ -748,9 +749,18 @@ def get_memory_limit(kwargs=None):
         kwargs.pop('memory')
     return memory_limit
 
+class ExecCode(Enum):
```
RagnarGrootKoerkamp, line 752: `Code` to me means the integer value. Rename to `ExitStatus`? `ExecStatus`?

bin/util.py
```diff
@@ -748,9 +749,18 @@ def get_memory_limit(kwargs=None):
...
+    ERROR = 3
+    TIMEOUT = 4
+
+    def __bool__(self):
+        return self == ExecCode.ACCEPTED
 
 class ExecResult:
-    def __init__(self, ok, duration, timeout_expired, err, out, verdict=None, print_verdict=None):
+    def __init__(self, code, ok, duration, timeout_expired, err, out, verdict=None, print_verdict=None):
+        self.code = code
```
RagnarGrootKoerkamp, line 763: Add some comments here.
- `code` is the return code (or None) (maybe rename to `returncode`, similar to `process.returncode`?)
- `ok` is the `ExecCode` above, but would be clearer to rename this to `status` or so also probably?

bin/util.py
```diff
@@ -927,7 +950,7 @@ def interrupt_handler(sig, frame):
     def maybe_crop(s):
         return crop_output(s) if crop else s
 
-    ok = True if process.returncode == expect else process.returncode
+    ok = exec_code_map(process.returncode)
```
RagnarGrootKoerkamp, line 953: same here, `ok` is not really a good name anymore

RagnarGrootKoerkamp (commented):
Looks great overall! Just some comments regarding docs and naming

RagnarGrootKoerkamp:
(hopefully) pushed a fix for failing CI

bin/util.py
mzuenni: it is probably useful, but the main reason to add this was because generators etc. handled this and i thought it would fit better here

bin/util.py
mzuenni: can you easily refactor this? ^^' (there are other things that are called `.ok` and i didn't want to break them...)

bin/testcase.py
```diff
@@ -231,17 +232,15 @@ def validate_format(
             flags = args if flags is None else flags + args
 
             ret = validator.run(self, mode=mode, constraints=constraints, args=flags)
-            if ret.ok is not True and ret.ok != config.RTV_WA:
-                bar.log(f"Expected exit code {config.RTV_AC} or {config.RTV_WA}, got {ret.ok}")
-                ret.ok = config.RTV_WA
-            ok = ret.ok == True
+            if ret.ok == ExecStatus.ERROR:
+                bar.log(f"Unxpected exit code!")
```
mpsijm, line 236: We could still print the exit code here (and perhaps even still mention that it should be either `config.RTV_AC` or `config.RTV_WA`):
```suggestion
                bar.log(f'Expected exit code {config.RTV_AC} or {config.RTV_WA}, got ${ret.returncode}')
```

---

Also, it's somewhat unfortunate that [Python uses `returncode`](https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess.returncode) as name for what I know as the "exit code" of a program... I wanted to propose renaming `ExecResult.returncode` to `.exit_code`, but on the Python side, it's probably best to stay consistent with the standard library. However, in user-facing messages, I'm in favour of writing "exit code" :slightly_smiling_face:

bin/util.py
mpsijm: It feels kind of weird to have the `ExecResult.ok` (to be renamed to `ExecCode.status` field have two semantics: one as a "boolean-able" and one as an actual enumeration :thinking: If we're rethinking the terminology here anyway, I would propose that we could separate the two? For example:
```py
class ExecResult:
    def __init__(self, returncode, status, duration, ...):
        self.returncode = returncode
        self.status = status
        ...
    
    def is_ok(self):
        self.status == ExecStatus.ACCEPTED
```
Then we also don't need the coercion method `ExecStatus.__bool__` anymore :slightly_smiling_face:

RagnarGrootKoerkamp: I don't mind the implicit conversion to bool so much though; it makes call sites nice and simple. But ok with either.

RagnarGrootKoerkamp: After renaming .ok to .status I feel like maybe this is sightly more appropriate now. as you wish.

mpsijm: I think I'm fine with the coercion to boolean if we check in the constructor of `ExecResult` that `status` really is one of the enum values, and not something silly like `None` :slightly_smiling_face: (note that the contructor currently still uses `ok` instead of `status` :stuck_out_tongue:

RagnarGrootKoerkamp: Oh I did that but forgot to save the file :)

Checking type is a good idea indeed.

RagnarGrootKoerkamp: grrr;`pytest` is a bit broken for me:
```
FATAL ERROR: Did not find problem.yaml. Are you running this from a problem directory?
/usr/lib/python3.11/site-packages/_pytest/main.py:306: PluggyTeardownRaisedWarning: A plugin raised an exception during an old-style hookwrapper teardown.
Plugin: terminalreporter, Hook: pytest_sessionfinish
SystemExit: 1
For more information see https://pluggy.readthedocs.io/en/stable/api_reference.html#pluggy.PluggyTeardownRaisedWarning
  config.hook.pytest_sessionfinish(
```

not sure what's wrong, and CI is having some different weird issue now...

bin/testcase.py
```diff
@@ -231,17 +238,15 @@ def validate_format(
...
-            if ret.ok is not True and ret.ok != config.RTV_WA:
-                bar.log(f"Expected exit code {config.RTV_AC} or {config.RTV_WA}, got {ret.ok}")
-                ret.ok = config.RTV_WA
-            ok = ret.ok == True
+            if ret.status == ExecStatus.ERROR:
+                bar.log(f"Unxpected exit code!")
 
-            validator_accepted.append(ok)
-            message = validator.name + (' accepted' if ok else ' rejected')
+            validator_accepted.append(True if ret.status else False)
```
mpsijm, line 244: Tiny optimization using the boolean coercion :smile:
```suggestion
            validator_accepted.append(bool(ret.status))
```

bin/testcase.py
mzuenni: we can display the actual exit code but we should not display what we expect since that changes depending on the type of validator?

mpsijm: Fair point, this looks good to me then :smile:

mpsijm (approved):

RagnarGrootKoerkamp (approved):

bin/testcase.py
```diff
@@ -266,14 +272,21 @@ def validate_format(
...
 
-            bar.part_done(
-                ret.status or expect_rejection,
-                message,
-                data=data,
-                warn_instead_of_error=warn_instead_of_error,
-            )
+            if expect_rejection:
+                bar.debug(
+                    message,
```
RagnarGrootKoerkamp, line 277: Hmm, so this causes fewer lines to be printed for invalid cases right? Fine with me i suppose

RagnarGrootKoerkamp (approved):

bin/testcase.py
mzuenni: yes

PR #339, opened 2024-02-15
fix hash

mzuenni:


RagnarGrootKoerkamp (approved):

bin/util.py
```diff
@@ -999,6 +1000,21 @@ def inc_label(label):
         label = label[:x] + 'A' + label[x + 1 :]
     return 'A' + label
 
+def combine_hashes(list):
```
mpsijm, line 1003: Can you rename `list` to not clash with the built-in `list`?

mpsijm (approved):

PR #340, opened 2024-02-16
[ci] Run pre-commit hooks on CI

mpsijm:
This should avoid the random formatting changes that we now require sometimes :smile: 

As a reminder for fellow contributors, please follow https://github.com/RagnarGrootKoerkamp/BAPCtools#contributing--style-guide to install the pre-commit hooks locally :slightly_smiling_face:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,10 +1,17 @@
 name: CI
 on: [push, pull_request]
+
 jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: pre-commit/action@v3.0.1
+
   test:
     runs-on: ubuntu-latest
     container: ragnargrootkoerkamp/bapctools
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yaml
```diff
@@ -1,10 +1,17 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: pre-commit/action@v3.0.1
+
   test:
     runs-on: ubuntu-latest
     container: ragnargrootkoerkamp/bapctools
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
```
RagnarGrootKoerkamp, line 15: Would it be more efficient to merge the jobs and run precommit as part of test? Or is that ugly?

mpsijm: The `checkout` action runs in less than 1 second, it looks like, so it's not too inefficient :slightly_smiling_face: Also, splitting up the jobs allows them to run in parallel. I was even doubting to split off the `cue vet` command, but that would require a double pull of BAPCtools, so let's keep that together with `pytest` to be slightly nice towards the GitHub Actions runners :stuck_out_tongue:

RagnarGrootKoerkamp: allright! feel free to merge

Issue #314, opened 2023-09-21
Add UUIDs


RagnarGrootKoerkamp:
See https://github.com/Kattis/problem-package-format/issues/124; we can add them in `bt new_problem`

mzuenni:
we may also want to warn (or just add) missing ids? maybe in `bt export` or already earlier?

PR #342, opened 2024-02-17
add uuid support

mzuenni:
#314


bin/util.py
```diff
@@ -1049,3 +1050,17 @@ def hash_file_or_dir(file_or_dir, buffer_size=65536):
...
+
+def generate_problem_uuid():
+    uuid = bytearray(secrets.token_bytes(16))
+    # mark this as v8 uuid (custom uuid) variant 0
+    uuid[6] &= 0b0000_1111
+    uuid[6] |= 0b1000_0000
+    uuid[8] &= 0b0011_1111
+    # format as uuid
+    uuid = uuid.hex()
+    uuid = f'{uuid[0:8]}-{uuid[8:12]}-{uuid[12:16]}-{uuid[16:20]}-{uuid[20:32]}'
```
RagnarGrootKoerkamp, line 1063: there must be some library for this, but probably that's more work than just having this code anyway

RagnarGrootKoerkamp (approved):
looks good, thanks!

I'm wondering if we could actually use this for more things. Ideas such as:
- use UUID instead of problemname for the tmpdir? so that multiple instances of a problem share tmpdir?
- Warn if two problems in different dirs have the same UUID?

But don't really think those are worth the benefit for now.

mzuenni:
I think I prefer the hashed directory path as tmpdir, we would need that as a fallback anyway?

PR #343, opened 2024-02-18
fixes

mzuenni:
this "fixes" a few things

- create tmpdir before writing .uuid 
- refactor licenses and use `VALIDATION_MODES`
- remove unofficial `.gif` support
- warn for deprecated files on export
- ensure that only testcase files including `.in` and `.ans` are exported
- include tescases for interactive problems

### DIFF
--- a/bin/export.py
+++ b/bin/export.py
@@ -104,37 +104,32 @@ def build_samples_zip(problems, statement_language):
 def build_problem_zip(problem, output, statement_language):
     """Make DOMjudge ZIP file for specified problem."""
 
-    files = [
-        ('domjudge-problem.ini', False),  # DEPRECATED, may be removed at some point.
-        ('problem.yaml', True),
-        ('problem_statement/*', True),
-        ('data/secret/**/*.in', True),
-        ('data/secret/**/*.ans', True),
-        ('submissions/accepted/**/*', True),
-        ('submissions/*/**/*', False),
-        ('attachments/**/*', False),
+    deprecated = [  # may be removed at some point.
+        'domjudge-problem.ini',
     ]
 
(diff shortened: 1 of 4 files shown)


bin/export.py
```diff
@@ -104,37 +104,32 @@ def build_samples_zip(problems, statement_language):
...
     ]
 
-    for ext in config.KNOWN_DATA_EXTENSIONS:
-        files += [
-            (f'data/sample/**/*{ext}', False),
-            (f'data/secret/**/*{ext}', False),
-        ]
+    testcases = [
+        ('data/secret/**/*.in', True),
+        ('data/sample/**/*.in', not problem.interactive),
```
RagnarGrootKoerkamp, line 121: Hmm I think by default we should include the sample inputs, so they can actually be run as samples?
Otherwise we should generatlly disallow `sample/*.{in,ans}` for interactive problems and make a big warning here if they are present.

bin/export.py
```diff
@@ -165,6 +171,25 @@ def build_problem_zip(problem, output, statement_language):
                     out = problem.name / out
                 copyfiles.add((f, out))
 
+    for pattern, required in testcases:
```
RagnarGrootKoerkamp, line 174: There's a bit much duplication here but I suppose we'll live with it. Can you add a comment what this loop does do differently than the one above?

bin/export.py
mzuenni: They are now included (unlike before) but not required for interactive problems?

RagnarGrootKoerkamp: Oh right I was completely misreading the code. Lack of sleep is showing up. All's good to go them :)

RagnarGrootKoerkamp (approved):

PR #344, opened 2024-02-20
fix validation

mzuenni:
correctly verify which of the `.in` `.ans` and `.out` files inside the `invalid_*` directories should be rejected

### DIFF
--- a/bin/problem.py
+++ b/bin/problem.py
@@ -224,7 +224,7 @@ class Problem:
             for prefix in {
                 validate.Mode.INPUT: ['secret', 'sample'],
                 validate.Mode.ANSWER: ['secret', 'sample'],
-                validate.Mode.INVALID: ['bad', 'invalid_*'],
+                validate.Mode.INVALID: config.INVALID_CASE_DIRECTORIES,
             }[mode]:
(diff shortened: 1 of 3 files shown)


thorehusfeldt:
Here is a toy program to test invalidation (and constraints):
[pickanumber.zip](https://github.com/RagnarGrootKoerkamp/BAPCtools/files/14355700/pickanumber.zip)

It has several validators:
```bash
pickanumber> tree *_validator*
answer_validators
└── answerformat.py
input_validators
├── invala
│   ├── validate.cpp
│   └── validation.h
└── invalb
    ├── validate.cpp
    └── validation.h
output_validator
└── validate.py

5 directories, 6 files
```

You can run
```
bt generate --no-validators
bt validate --v --invalid
```

Here is the expected output:

image

Note, e.g., that testcase `neg_b` passes the first input validator (`invala`) only to be rejected by the other `invalb`, which is therefore green. Testcases named `actually_valid` should trigger red lines; they were (deliberately) misplaced into `invalid_` in order to check that invalidation works. Every validator is run on every testcase at most once.

thorehusfeldt:
Looks good to me.

Issue #305, opened 2023-09-12
Rethink `.gitignore` of `data/`


RagnarGrootKoerkamp:
**Current situation:**

Setting `gitignore_generated: true` in `generators.yaml` writes `data/.gitignore` when generating, which contains all generated cases, i.e.:
```
.gitignore
secret/1-min.*
secret/2-max.*
bad/empty.*
```

This works nicely, and allows having sample files directly in `data/sample/*`.

**New situation:**

With the new generator spec, having 'inline' source files directly in `data/` is discouraged. It would much simpler and much more consistent to simply ignore all `data` directories in their entirety, and generate warnings for any files in `data/` that are not generated.

A) When to ignore files?
1. Always
2. Only when `gitignore_generated: true` is set

My preference: A1, always, since that's what we always use in practice anyway, and then we can drop the `gitignore_generated` flag and simplify code since there are fewer different usages.

B) Some approaches for ignoring files:
1. Let user put `problems/*/data/` in toplevel `.gitignore` manually.
2. Generate `problem/.gitignore` containing `data/`
3. Generate `problem/data/.gitignore` containing `*`; using the file we currently also use.

My preference: B3, to have automation, but not mess around with a possible `problem/.gitignore` file that could already exist. We can claim ownership of `problem/data/.gitignore` without issues IMO.

C) Some approaches for warning about non-generated files:
1. Always
2. When `gitignore_generate: true` is set
3. When `problem/data/.gitignore` contains `*`, or some specific BAPCtools-generated string/comment.

My preference: C1, again to simplify the implementation and possible use cases.

thorehusfeldt:
B1.

mzuenni:
B3. we want to claim ownership of the whole data dir anyway?

RagnarGrootKoerkamp:
(although not completely independent I think A, Band C each need their own answer ;)

mpsijm:
My preferences are the same as Ragnar and mzuenni: A1, B3, C1 :slightly_smiling_face:

RagnarGrootKoerkamp:
One things that we may want to decide on separately is whether to gitignore samples. It's kinda convenient to have them committed when building PDFs, so they don't have to be re-generated each time `bt pdf` is run somewhere.

RagnarGrootKoerkamp:
[...]

PR #347, opened 2024-02-27
changed gitignore handling

mzuenni:
#305


mpsijm:
Thanks for simplifying the implementation, but I think I still don't like having `.gitignore` files in the root of my problem directory :sweat_smile: I think we need more opinions here :stuck_out_tongue:

mzuenni:
it should be easy to change now ^^ (as long as the `gitignore` stays in the problem folder). I just hope that i found all places that needed a change

mpsijm:
> it should be easy to change now ^^ (as long as the `gitignore` stays in the problem folder).

But it's not easy to change when I don't have a `.gitignore` file in my problem folders, and BAPCtools keeps regenerating it :sweat_smile:

mzuenni:
I meant it should be easy to change the code if we decide for the self ignoring gitignore in `/data`.

bin/generate.py
```diff
@@ -1699,60 +1700,24 @@ def generate_dir_unlisted(d):
...
         if gitignorefile.is_file():
-            text = gitignorefile.read_text()
-            if not text.startswith('# GENERATED BY BAPCtools'):
-                warn('Not overwriting existing data/.gitignore file.')
-                return
+            # if there is any rule fore data/ we expect that the user knows
+            # what he does.
+            if gitignorefile.read_text().find('data/') == -1:
+                with gitignorefile.open("a") as f:
+                    f.write(content)
```
RagnarGrootKoerkamp, line 1715: Maybe write a preceding newline to not create bugs when `f` doesn't end in a newline.

skel/problem/.gitignore
```diff
@@ -0,0 +1,3 @@
+#GENERATED BY BAPCtools
```
RagnarGrootKoerkamp, line 1: I think only writing the .gitignore once `bt generate` has been invoked is preferable. That way we don't ignore things if the user doesn't use `bt generate`.

RagnarGrootKoerkamp:
I also like putting this in the problem .gitignore. Indeed it's a bit redundant but I agree it's more a problem than a contest setting, and anyway these files are automatically generated so there is little overhead.

So this settles A1 and B2 from #305. That only leaves the cleanup of unknown-file handling (C).

RagnarGrootKoerkamp (approved):

Issue #304, opened 2023-09-06
Generators: reconsider -f --samples defaults


thorehusfeldt:
The generators framework work very smoothly, but I now constantly get tripped up in the following dance:

```bash
> bt generate
PROBLEM closestpairscoring
Generate: sample/2                    SKIPPED: 2.ans; supply -f --samples to overwrite
Generate: sample/1                    SKIPPED: 1.ans; supply -f --samples to overwrite
Generate: sample/3                    SKIPPED: 3.ans; supply -f --samples to overwrite
^[[A^CFATAL ERROR: Running interrupted[######################---------------------------------------------------------------------------------- 11/48]
> bt generate -f --samples
PROBLEM closestpairscoring
Generate: sample/2                    CHANGED 2.ans
Generate: sample/1                    CHANGED 1.ans
Generate: sample/3                    CHANGED 3.ans
```
The `-f --samples` tradition targets a workflow where sample typically are provided _as files_ in `data/samples`, but the new framework much prefers to have everything in `generators/generators.yaml`.

This confuses two recommendations for what “the expected workflow” is; an author that follows best practice should be supported by default settings.

Suggestions:
1. drop the default `SKIPPED` behaviour
2. introduce `config.I_know_what_Im_doing_shut_up` or `config.this_contest_has_some_people_with_dirty_samples` somewhere (presumably with a better name)
3. make BAPC check if _all files in `data` are mentioned in `generators.yaml`_. If so, assume the author knows what they’re doing and set the default to `-f --sample`. (This is very hard to document an can lead to unexpected behaviour.)

RagnarGrootKoerkamp:
1. I think dropping the `--samples` part of `-f --samples` is OK and we can simply warn when this happens. If you care about specific sample output you can hardcode/copy them. (Otherwise there is also usually `git` to recover what you lost.)
2. I'm not a fan of introducing such flags. Good default behavior for power users is more convenient.
3. See #305. Indeed I think we should (always/by default) check that all files in `data/` are generated, to simplify towards a single workflow of generating everything.

mzuenni:
Current idea for `bt generate`:

1. remove all files in `tmp/problemdir/removed` that are older than ~10min
2. identify all files that are up to date
3. move all files that are not up to date to `tmp/problemdir/removed` (and print a warning with the path to this dir if this happens)
4. generate all missing files

- This should resolve the `-f --sample` behavior by just always regenerating them
[...]

PR #348, opened 2024-02-29
New generate

mzuenni:
#304

### DIFF
--- a/bin/generate.py
+++ b/bin/generate.py
@@ -392,6 +392,9 @@ class TestcaseRule(Rule):
         assert is_testcase(yaml)
         assert config.COMPILED_FILE_NAME_REGEX.fullmatch(name + '.in')
 
+        # if False rule will be skipped during generation
+        self.ok = True
+
         if name.endswith('.in'):
             error(f'Testcase names should not end with \'.in\': {parent.path / name}')
             name = name[:-3]
@@ -434,7 +437,11 @@ class TestcaseRule(Rule):
 
         if yaml is None:
             intarget = self.path.parent / (self.path.name + '.in')
-            fatal(f'Empty yaml key (Testcases must be generated not mentioned): {intarget}')
+            error(
+                f'Empty yaml key (Testcases must be generated not mentioned): {intarget}. Skipping.'
+            )
+            self.ok = False
+            return
         else:
             check_type('testcase', yaml, [str, dict])
             if isinstance(yaml, str):
@@ -455,7 +462,9 @@ class TestcaseRule(Rule):
             if 'generate' in yaml:
                 check_type('generate', yaml['generate'], str)
                 if len(yaml['generate']) == 0:
-                    fatal(f'`generate` must not be empty')
+                    error(f'`generate` must not be empty. Skipping.')
+                    self.ok = False
+                    return
 
                 self.generator = GeneratorInvocation(problem, yaml['generate'])
 
@@ -504,7 +513,9 @@ class TestcaseRule(Rule):
         # Warn/Error for unknown keys.
         for key in yaml:
             if key in RESERVED_TESTCASE_KEYS:
-                fatal('Testcase must not contain reserved key {key}.')
+                error('Testcase must not contain reserved key {key}. Skipping.')
+                self.ok = False
+                return
             if key not in KNOWN_TESTCASE_KEYS:
                 if config.args.action == 'generate':
                     log(f'Unknown testcase level key: {key} in {self.path}')
@@ -526,9 +537,11 @@ class TestcaseRule(Rule):
         if self.hash in generator_config.rules_cache:
             # This is fatal to prevent crashes later on when running generators in parallel.
             # https://github.com/RagnarGrootKoerkamp/BAPCtools/issues/310
-            fatal(
-                f'Found identical input at {generator_config.rules_cache[self.hash]} and {self.path}'
+            error(
+                f'Found identical input at {generator_config.rules_cache[self.hash]} and {self.path}. Skipping.'
             )
+            self.ok = False
+            return
         generator_config.rules_cache[self.hash] = self.path
 
     def generate(t, problem, generator_config, parent_bar):
@@ -537,6 +550,9 @@ class TestcaseRule(Rule):
         t.generate_success = False
 
         # Some early checks.
+        if not t.ok:
+            bar.done(False, f'Invalid yaml entry.')
+            return
         if t.generator and t.generator.program is None:
             bar.done(False, f'Generator didn\'t build.')
             return
--- a/bin/generate.py
+++ b/bin/generate.py
@@ -1247,7 +1247,7 @@ class GeneratorConfig:
     # Parse generators.yaml.
     def __init__(self, problem):
         self.problem = problem
-        yaml_path = self.problem.path / 'generators/generators.yaml'
+        yaml_path = self.problem.path / 'generators' / 'generators.yaml'
         self.ok = True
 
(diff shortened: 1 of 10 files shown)


bin/tools.py
```diff
@@ -505,28 +496,7 @@ def build_parser():
     genparser.add_argument(
         '--timeout', '-t', type=int, help='Override the default timeout. Default: 30.'
     )
-    genparser.add_argument(
```
RagnarGrootKoerkamp, line 508: Hmmm, I do like not accidentally overwriting samples, for example with more/less precise floating point numbers.

On the other hand, we assume everything is in git anyway so any diffs will clearly show up and hard to miss. Removing this probably ends up simpler, more predictable, and anyway safe enough.

bin/tools.py
```diff
@@ -505,28 +496,7 @@ def build_parser():
...
-        '--samples',
-        action='store_true',
-        help='Overwrite the samples as well, in combination with -f.',
-    )
-
-    genparser_group = genparser.add_mutually_exclusive_group()
-    genparser_group.add_argument(
```
RagnarGrootKoerkamp, line 515: The one benefit of `--add-unlisted` is that one can just dump files in `data/secret`, run `bt generate --add-unlisted` and then they're automatically moved to `generators/manual/` *and* a rule in `generators.yaml` is added.
Without this command, each new testcase needs two actions: 1) writing the file, and 2) adding the name of the file to `generators.yaml`, which is slightly redundant.

Nevertheless, I suppose we're all getting more and more used to `bt generate` now and the new workflow is not terrible, and also rarely used.
If needed we could add a new command `bt add_manual_generated_files generators/manual/*.in` which updates the yaml.

doc/commands.md
```diff
@@ -234,8 +215,8 @@ This table contains:
 This may look like:
 
 ```
-problem               yaml ini tex sol    Ival Oval    sample secret    AC  WA TLE    cpp java py2 py3   comment
-A appealtotheaudience    Y   Y   Y   N       Y    Y         2     30     4   4   2      2    0   0   2
+problem               time yaml tex sol   val: I A O    sample secret bad    AC  WA TLE subs   cpp py java kt   comment
```
RagnarGrootKoerkamp, line 218: side note: maybe `bad` can be replaced by `inv: I A O` just like `val`. Although maybe that's more verbose than needed.

bin/tools.py
mzuenni: We can add a separate command but the workflow should be:

- dump the files in `generator/???`
- add the rules to the `generator.yaml` with the new commnd

but moving from `data` is just weird since you are not supposed to touch `data`?

RagnarGrootKoerkamp: Yeah I agree; I'm just used to the old behaviour :P Removing and cleaning up is good :)

mzuenni:
@mpsijm i tried to start the rewriting but for me the parsing of the yaml and the actual generation is to entangled and confusing, do you want to rewrite it from scratch or do you have a better understanding?

RagnarGrootKoerkamp:
The 'problem' with `bt stat` not including `generators.yaml` content is that this requires (re)generating all testcases for an entire contest even if you just want to have a quick overview of how much is done everywhere. I use this quite a lot in fact since I usually care about progress for most problems but don't work on all of them (or at least don't have the latest version of testdata everywhere).

RagnarGrootKoerkamp:
What exactly would you like to clean up / rewrite in `generate.py`? If things are working now I don't think it's super useful to spend time simplifying it. (I do agree it's complicated, and if you think that prevents making progress on it that's a valid reason.)

mzuenni:
My main problem with the current code is that stuff is not cleanly separated (or I just understand it wrong).
What I would want is a `ParsedConfig` object which stores data and only parses it for errors and a generate function that manages the content of `data` based on the `ParsedConfig` object

The `ParsedConfig` should have fields like this:
2. a default submission
3. a visualizer
4. a list of generators
5. a list of testcase rules
6. a list of includes for testcases
7. ...?

The generate function should then:

1. build all programmes (parallel)
2. generate all directories (sequential)
3. generate everything in the `tmp/data/` dir (parallel)
4. match files in the cache with files in `data` (sequential)
5. move unmatched files to `tmp/trash`  (sequential)
6. move missing files to `data`  (sequential)
7. add symlinks for includes  (sequential)

So my main goal is that `generator.yaml` parsing should be completely independent of the data folder.

doc/commands.md
mpsijm: The `bad` stat is currently the sum of all invalid directories. I think that's fine, I don't use it much anyway :stuck_out_tongue: 

(note that I updated `bt stats` on `master`, thanks @mzuenni for updating the documentation :smile: :heart:)

RagnarGrootKoerkamp: At least we can rename to `invalid` or `inv` now

mzuenni:
ok, I hope we can reuse a lot of code that is already there... we only need to untangle parsing the yaml from generating the files...
(right now `generate --add` does not work correctly... for some reason known_testcases is empty?)

mzuenni:
yes... and maybe even some new bugs :)

bin/generate.py
```diff
@@ -1553,10 +1487,40 @@ def generate_dir(d):
...
 
-        self.update_gitignore_file()
+    def remove(self, src):
+        if self.trashdir is None:
+            self.trashdir = self.problem.tmpdir / secrets.token_hex(4)
+        dst = self.trashdir / src.resolve().relative_to((self.problem.path / 'data').resolve())
+        dst.parent.mkdir(parents=True, exist_ok=True)
+        shutil.move(src, dst)
+
+    def remove_unknown(self, path, bar):
```
RagnarGrootKoerkamp, line 1471: this could use some comments. But also it doesn't seem to be called yet.

bin/generate.py
```diff
@@ -436,13 +433,8 @@ def __init__(self, problem, generator_config, key, name: str, yaml, parent):
...
 
         if yaml is None:
-            self.inline = True
-            yaml = dict()
             intarget = self.path.parent / (self.path.name + '.in')
-            target_infile = problem.path / 'data' / intarget
-            if not target_infile.exists():
-                fatal(f'Could not find .in for inline case {intarget}')
-            self.hash = hash_file(target_infile)
+            fatal(f'Testcases should not be placed in the data directory: {intarget}')
```
RagnarGrootKoerkamp, line 437: Also add that the yaml is empty.

mzuenni:
I hope this is working now (we may still want to rewrite stuff at some point...) can someone check? @RagnarGrootKoerkamp @mpsijm

bin/generate.py
```diff
@@ -1654,11 +1654,17 @@ def generate_samples(problem):
...
     gen_config = GeneratorConfig(problem)
     if gen_config.has_yaml:
-        # TODO handle includes flag
         if gen_config.ok:
-            return {
-                problem.path / 'data' / x.parent / (x.name + '.in') for x in gen_config.known_cases
-            }
+            if includes:
+                return {
+                    problem.path / 'data' / x.parent / (x.name + '.in')
```
RagnarGrootKoerkamp, line 1660: The set automatically dedups right?

mzuenni:
One open question is:
What exactly should happen if we call `bt generate` with a set of testcases...

RagnarGrootKoerkamp:
Ideally: Files corresponding to these cases are touched and nothing else.

(Throwing away *all* unknown cases could also be OK, but preferably not I'd say.)

PR #351, opened 2024-03-05
Validate generators.yaml better

thorehusfeldt:
1. Update the schema for `testdata.yaml` to the `2023-07-draft` version.
This involves retiring some outdated `testdata.yaml` entries in `generators.yaml`-files, such as 
```yaml
testdata.yaml:
        on_reject: break
        accept_score: "25"
        range: 0 25
        grader_flags: min
```
2. Split the CUE schema in two, one things that are specified in the official problem package format (valid filenames, testdata settings), one for BAPC’s generators.
3. Generator names can’t include `.`, so generators like `main.cpp` in `identity` are renamed to `main_cpp`
4. `data` cannot be `null`, so removed some empty productions from generator in `identity`
5. For the same reason, added a single testcase to the `invalid_*` part of the problem skeleton
6. Removed universally `invalid_input` from the skeleton


skel/problem/generators/generators.yaml
```diff
@@ -62,32 +62,32 @@ data:
...
 
   invalid_answers:
     # Add valid testcase with invalid answers to ensure that the answer_validators correctly rejects them.
     # (The output validator is also called with these)
     data:
-      #dont_start_with_whitespace:
+      #leading_whitespace:
       #    in: "<valid testcase>"
       #    ans: " <valid answer>"
+      in: 2 2
```
RagnarGrootKoerkamp, line 79: hmm not sure how I feel about hardcoding these problem specific examples into the skel. Just putting `data: {}` is also fine I'd say.


This one is missing a testcase name.

skel/problem/generators/generators.yaml
```diff
@@ -62,32 +62,32 @@ data:
...
-          in: "\n"
-      random:
-          in: YVRtr&*teTsRjs8ZC2%kN*T63V@jJq!d
-      not_printable_1:
-          in: "\x7f"
-      not_printable_2:
-          in: "\xe2\x82\xac"
-      unicode:
-          in: "¯\\_(ツ)_/¯"
+      too_many_tokens:
```
RagnarGrootKoerkamp, line 65: Can you drop this first one (`too_many_tokens`) since it's not universally a bad input.

RagnarGrootKoerkamp (commented):
Some small comments on the skel; otherwise lgtm

skel/problem/generators/generators.yaml
```diff
@@ -3,7 +3,7 @@
 
 data:
   sample:
-    data:
+    data: []
```
mpsijm, line 6: Good idea to make the skeleton spec-compliant like this! Do you think we should explain that a problem setter should remove the `[]` to make use of the commented-out examples below this? I guess this could lead to confusion, especially to people who are new to BAPCtools :slightly_smiling_face:

thorehusfeldt:
@RagnarGrootKoerkamp : I reverted my suggestions for rewriting the `skel/problem/generatators/generators.yaml`. There are too many issues involved that are orthogonal to the ambition of this particular pull request. Now I only replaced the empty values in `data:` with `data: []` to make the skeleton conform.

This is ready to be merged.

skel/problem/generators/generators.yaml
thorehusfeldt: I have no opinion either way; and just did the minimally intrusive edit. Others should retain the editorial responsibility for the skeleton. (Which contains _a lot_ of information, for instance about repurposing symlinked input validators as generators, which I believe is at least as confusing. I think it’s good enough as it is, and the didactics are a different issue in any case.)

PR #353, opened 2024-03-05
#271 [generate] Add count

### LINKED ISSUE #271: Generators: Add `testcase: count: X` for multiple testcase copies

mzuenni:
tried to implement count... its a bit hacky. Fixes #271


bin/generate.py
```diff
@@ -1320,7 +1346,9 @@ def parse(key, name, yaml, parent):
                                     f'Unnumbered testcases must not have an empty key: {Path("data") / d.path / child_name}/\'\''
                                 )
                         c = parse(child_key, child_name, child_yaml, d)
-                        if c is not None:
+                        if isinstance(c, Sequence):
```
RagnarGrootKoerkamp, line 1349: where does this come from?

bin/generate.py
```diff
@@ -1261,18 +1275,30 @@ def parse(key, name, yaml, parent):
...
+
                 if not config.COMPILED_FILE_NAME_REGEX.fullmatch(name + '.in'):
                     error(f'Testcase \'{parent.path}/{name}.in\' has an invalid name.')
                     return None
 
-                # If a list of testcases was passed and this one is not in it, skip it.
-                if not self.process_testcase(parent.path / name):
-                    return None
+                ts = []
+                for i in range(count):
```
RagnarGrootKoerkamp, line 1287: rename i to count_index?

bin/generate.py
```diff
@@ -1261,18 +1275,30 @@ def parse(key, name, yaml, parent):
...
+                    # If a list of testcases was passed and this one is not in it, skip it.
+                    if not self.process_testcase(parent.path / name):
```
RagnarGrootKoerkamp, line 1289: Can you make it so that both `name` and `curname` trigger processing?

bin/generate.py
```diff
@@ -1261,18 +1275,30 @@ def parse(key, name, yaml, parent):
...
+                        continue
+
+                    curname = name if count == 1 else name + f'-{i+1:0{len(str(count))}}'
+
+                    t = TestcaseRule(self.problem, self, key, curname, yaml, parent, i)
+                    if t.path in self.known_cases:
+                        error(f'{t.path} was already parsed')
```
RagnarGrootKoerkamp, line 1296: Maybe clarify this a bit; something like `Counted testcase name already appears in generators.yaml`

RagnarGrootKoerkamp (commented):
Looks pretty decent I'd say! Pretty straightforward actually.

The code you added lgtm; I suppose the bigger question is if there are now other places that also need changing that we didn't think of :sweat_smile:

RagnarGrootKoerkamp:
Hmm this makes a test fail. Didn't check yet but maybe the seeding subtly changed? Would be best to avoid that if we can.

Also this will need some tests of its own and docs.

bin/generate.py
mzuenni: a single entry can now return multiple rules i.e. if count > 1

RagnarGrootKoerkamp: Ahh right; can you add that as a comment at the top of the `parse` function?

But `ts = []; ...; return ts` should just return a `list` right? No need to import `Sequence`?

bin/generate.py
mzuenni: i think only curname should trigger it?

bin/generate.py
mzuenni: The error was there before and we also display them as if it would have been multiple rules?

bin/generate.py
RagnarGrootKoerkamp: Hmm; `curname` should trigger for sure.
One case for `name` triggering it: if you increase the count, those are not files yet so you can't use globbing, and then writing `name` is simpler. But anyway this is rarely used to anything is fine.

mzuenni: but thats also the case if you add new rules?

RagnarGrootKoerkamp: ah good  point. `curname` only is good then.

mzuenni:
maybe it would be nice to be able to use count "inline" i.e. combining it with this:
```- name: gen {seed}```
instead of needing to write this:
```
- name:
    generator: gen {seed}
    count: 5
```

**Comment by RagnarGrootKoerkamp** [comment]
Hmm, maybe something like `{seed:count=5}`? But that's already abused for changing the seed value.

Or completely replace `{seed}` by `{count=N}`?

On the other hand this is already a shorthand to generate multiple cases; spelling them out a bit more verbosely to clearly signal that this creates multiple cases may not be so bad?

**Comment by mzuenni** [comment]
beeing explicit with count is good it is just unusual to write the `generate` key as its the default else :D

**Comment by RagnarGrootKoerkamp** [comment]
Fair point. One option that we shouldn't do is
```yaml
- name: <gen>
  count: cnt
```

but that is super ugly and also doesn't work for non-numbered cases.

mzuenni:
that also doesn't work with the yaml parser if i remember correctly

mpsijm:
Can you also add this key to the `generators.yaml` in the documentation folder, and perhaps also in the commented-out part of the skeleton? :slightly_smiling_face:

mzuenni:
can someone update the cue to include count?

thorehusfeldt:
> can someone update the cue to include count? 

Yes. As soon as I understand the syntax.

Here’s the current regex for `command`:

```cue
// A command invokes a generator, like "tree --n 5".
// The regex restricts occurrences of curly-bracketed expressions
// to things like "tree --random --seed {seed:5}"
command: !="" & (=~"^[^{}]*(\\{seed(:[0-9]+)?\\}[^{}]*)*$")
```

RagnarGrootKoerkamp:
The syntax is as here https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/353#issuecomment-1981027148
Just a new count keyword alongside generator, with a value between 1 and 100 inclusive.

thorehusfeldt:
By “alongside generator” do you mean a new key for a `#testcase` struct/map/dict, alongside `generate`, such as
```yaml
testcase_name:
  generate: foo --bar 2 {seed}
  count: 23
```
If so, is the following allowed:
```yaml
testcase_name:
  in: "34 5"
  count: 23
``` 
or can `count` only  appear in a `#testcase` when `generate` occurs?

thorehusfeldt:
I added my best interpretation of the syntax, but remain confused. In `doc/generators.yaml` we have

```yaml
      12-counted:
          - generate: graph {seed:3}
          - count: 2                       # generate two testcases at once
```

Is this just wrong? (My schema complains; maybe that’s a good thing.)

mzuenni:
copy is only valid if there is  `generate: <command>` entry where command contains `{seed}` or `{seed:[0-9]+}` (not sure if its possible to check the seed).

The entry looks good to me? (the indention is also right or?)

thorehusfeldt:
CUE actually allows
```cue         
    generate?: string
    if generate != _|_ {
        if generate =~ "{seed(:[0-9]+)?}" {
          count?: int & >= 1 & <= 100
        }
    }
```
which is exactly the rule “make `count` available if `generate` is available and conforms to this-and-that regex”. But I decide against this; let’s have a readable but not maximally invasive schema. (CUE syntax for conditional optional fields will change on this point anyway.)

thorehusfeldt:
… now we just need to get 351, 353 (which both manipulate the schema) merged… I’ll stand way back and cover my ears.

RagnarGrootKoerkamp:
So currently the cue allows `count` even if there is no `generate:`. Fine by me.

@mzuenni How does `generate.py` handle this? Does it  already warn if `count` is set alongside `in:` (or anything else than `generate:`)?

Thinking about it, there *are* in fact usecases for `count:` alongside `in:`, for example when the interactor internally generates a random seed and we just need 100 empty input files. So we should either support that or warn that it's not handled.

mzuenni:
We generate a parse error if count is used without a generator with seed.
Yes, there are use cases for this, however, we currently forbid identical rules anyway and count is no exception to this as it just duplicates the rule.

If you want I can lift that restriction for count by including count_index in the hash for the cache entry. We then could also propose count as a solution if we find identical rules instead of just skipping those duplicates?

mzuenni:
Actually my prefered way right now would be this:

- I fix the parallel issue for duplicated testcases (shouldn't be hard)
- We still warn for duplicate testcases
- We say that if you want to suppress that error you should use count.
- We allow count for all kinds

RagnarGrootKoerkamp:
I was gonna write that we can postpone generalizing count until we need it, but that plan sounds nice, so go ahead :)

mzuenni:
Shouldn't be too complicated :)

bin/generate.py
```diff
@@ -532,47 +541,92 @@ def __init__(self, problem, generator_config, key, name: str, yaml, parent):
...
-
         # E.g. bapctmp/problem/data/<hash>.in
         cwd = problem.tmpdir / 'data' / t.hash
         cwd.mkdir(parents=True, exist_ok=True)
         infile = cwd / 'testcase.in'
         ansfile = cwd / 'testcase.ans'
         meta_path = cwd / 'meta_.yaml'
 
+        # Returns False when some files were skipped.
+        def copy_generated():
```
RagnarGrootKoerkamp, line 588: Do you want to move this into `def generate_includes()` below? Since you comment on the parallel invocation that includes and copies happen in step 2, after dirs and testcases?

bin/generate.py
```diff
@@ -456,16 +456,16 @@ def __init__(self, problem, generator_config, key, name: str, yaml, parent, coun
             # checks
             if not any(x in yaml for x in ['generate', 'copy', 'in', 'interaction']):
                 self.parse_error = 'Testcase requires at least one key in "generate", "copy", "in", "interaction". Skipping.'
-                config.n_error += 1
+                generator_config.n_parse_error += 1
```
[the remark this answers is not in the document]
mzuenni: we may parse the yaml but dont run the generators (for example for bt stat) and than it would be weird if we dont display errors but exit with non zero exit code

bin/generate.py
mzuenni: i dont know what you mean? ^^'

bin/generate.py
```diff
@@ -1035,6 +1057,7 @@ def generate(d, problem, generator_config, bar):
             bar.log(f'REMOVED: testdata.yaml')
         bar.done()
 
+    def generate_includes(d, problem, generator_config, bar):
```
RagnarGrootKoerkamp, line 1060: Re. the other comment: here you split `generate` into `generate_includes`. Shouldn't the `copy_generated` bit live inside here, as further down you write:
```
        def generate_copies_and_includes(d):
            p.join()
            d.generate_includes(self.problem, self, bar)
```

mzuenni: copy generated is the function that moves files from tmp to data (the function was always there)

mzuenni:
I think right now everything works as expected?

RagnarGrootKoerkamp:
Can you add some count rules to the identity test problem? Also for both generate: and in: and copy: rules

PR #355, opened 2024-03-08
Apply validator flags from testdata.yaml

thorehusfeldt:
Read flags for validators from testdata.yaml, such as

```yaml
input_validator_flags:
  format_validator: case_sensitive
  topology: --max_connected_components 1
output_validator_flags: float_tolerance 1e-5
---
input_validator_flags: min_n 100 max_n 1000
```

and apply them correctly. 

Interesting details that need to be kept in mind:
- [x] Input validators can (but don’t have to) be specified by validator name
- [x] There are inheritance rules (the closest applicable parental flags “win”)
- [x] The empty string are valid flags (relevant for inheritance)
- [x] Flags apply to included testcases at the point of their inclusion (rather than their origin)
- [x] Answer validators use output validator flags


bin/problem.py
```diff
@@ -172,17 +172,73 @@ def _read_settings(self):
...
+                    flags = p._testdata_yamls[f]
+                else:
+                    p._testdata_yamls[f] = flags = read_yaml(f, plain=True)
+
+                # Validate and exctract the flags
+                for k in flags:
+                    match k:
+                        case 'output_validator_flags':
+                            if not isinstance(flags[k], str):
+                                error_handler("ouput_validator_flags must be string")
```
RagnarGrootKoerkamp: This is slightly inconsistent:
`output_validator_flags` is checked for `str` only if `output_validator_flags` comes before `input_validator_flags` in case `key == input_validator_flags`.
I'd say either:
- First loop to check that all keys are known and correct style, and then return `flags[k]`.
- Only check `flags[k]`.

RagnarGrootKoerkamp (commented):
Looks great! just one comment

bin/problem.py
thorehusfeldt: … you mean “only check `flags[key]`, right? (I.e., the method argument, rather than the iteration variable.)

RagnarGrootKoerkamp: yes, indeed

RagnarGrootKoerkamp (approved):

PR #357, opened 2024-03-08
Keep global numbering for test cases with `count: X` (#271)

mzuenni:
not sure if this is the best solution...

### DIFF
--- a/bin/generate.py
+++ b/bin/generate.py
@@ -1332,6 +1332,8 @@ class GeneratorConfig:
                 for count_index in range(count):
                     if count_index > 0:
                         name = name_gen()
+                    if count > 1:
+                        name += f'-{count_index+1:0{len(str(count))}}'
 
                     # If a list of testcases was passed and this one is not in it, skip it.
                     if not self.process_testcase(parent.path / name):
--- a/bin/generate.py
+++ b/bin/generate.py
@@ -1275,6 +1275,20 @@ class GeneratorConfig:
         num_numbered_testcases = 0
         testcase_id = 0
 
+        def parse_count(yaml, warning=False):
+            if yaml is None or 'count' not in yaml or not isinstance(yaml['count'], int):
+                return 1
+            count = yaml['count']
+            if count < 1:
+                if warning:
+                    warn(f'Found count: {count}, increased to 1.')
+                return 1
+            if count > 100:
+                if warning:
+                    warn(f'Found count: {count}, limited to 100.')
+                return 100
+            return count
+
         # Count the number of testcases in the given directory yaml.
         # This parser is quite forgiving,
         def count(yaml):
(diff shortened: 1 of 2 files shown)


bin/generate.py
```diff
@@ -1275,6 +1275,20 @@ def add_known(obj):
         num_numbered_testcases = 0
         testcase_id = 0
 
+        def parse_count(yaml, warning=False):
+            if yaml is None or 'count' not in yaml or not isinstance(yaml['count'], int):
```
mpsijm, line 1279: Do we want to warn if `count` is given, but not an `int`?

bin/generate.py
```diff
@@ -1291,44 +1305,39 @@ def count(yaml):
...
                 ts = []
                 for count_index in range(count):
-                    curname = name if count == 1 else name + f'-{count_index+1:0{len(str(count))}}'
+                    if count_index > 0:
+                        name = name_gen()
 
                     # If a list of testcases was passed and this one is not in it, skip it.
-                    if not self.process_testcase(parent.path / curname):
+                    if not self.process_testcase(parent.path / name):
                         continue
```
mpsijm, line 1338: I have the feeling that this does not really have a proper semantics for test cases with a `count:` :thinking:

bin/generate.py
```diff
@@ -1363,24 +1372,34 @@ def parse(key, name, yaml, parent):
...
+                                    )
+
+                                child_name = next_testgroup_name
                             elif is_testcase(child_yaml):
-                                testcase_id += 1
-                                child_name = numbered_testcase_name(
-                                    child_key, testcase_id, num_numbered_testcases
-                                )
+
+                                def next_testcase_name():
```
mpsijm, line 1386: Perhaps we can move the `next_testgroup_name` and `next_testcase_name` outside of the loops, to make sure that Python does not re-create them with every iteration :slightly_smiling_face:

mpsijm (commented):
I was thinking of a similar approach, it isn't as bad as I feared :smile:

bin/generate.py
mzuenni: testcases with count behave the same way as any other... i.e. have regular names?

bin/generate.py
mzuenni: we need to know child_key?

bin/generate.py
mpsijm: I never actually used `bt generate [testcases...]`, so I'm also not 100% sure what the syntax should be for the testcases that should be generated. Any syntax that I try, results in `bt generate` simply printing "Done" twice, with nothing happening, even though I make changes to the relevant test cases in `generators.yaml` :thinking:

mzuenni: you should give it a path like `data/secret/blub.in` i guess?

bin/generate.py
mzuenni: this happens inside the TestcaseRule

mzuenni:
Hm now count only works if the directory is numbered... i think we should still append the count_index?

mpsijm (approved):
Thanks for implementing this! :smile:

PR #358, opened 2024-03-10
Testgroup verdicts

thorehusfeldt:
Properly determine the verdicts for each testgroup.

- [x] Refactoring: introduce `verdict.py`
- [x] move util.VerdictTable to `verdict.py`
- [x] In `verdict.py`, provide method for `Run.run` to register verdicts per testcase as those are computed, inferring testgroup verdicts as these become determined. This registry overlaps with the current internal functionality of `VerdictTable` and needs to be unified. This is probably just `class VerdictForTestItem(Dict)`, mapping `str` (testcase and testgroup names) to `verdicts.Verdict`.
- [x] Rewrite `Run.run` to leave the inference of the final (root) verdict to `verdict.py` 
- [x] Possibly provide `tree`-like overview of verdicts per testcase in UI
- [ ] Possibly  move `Submission._get_expected_verdicts` and related code to `verdict.py`

### DIFF
--- a/bin/run.py
+++ b/bin/run.py
@@ -6,7 +6,7 @@ import config
 import interactive
 import parallel
 import validate
-from verdicts import Verdicts, Verdict
+from verdicts import Verdicts, Verdict, from_string
 from typing import Type
 
 from util import *
@@ -40,14 +40,15 @@ class Run:
             result = interactive.run_interactive_testcase(
                 self, interaction=interaction, submission_args=submission_args
             )
+            # TODO : this is messed up wrt result.verdict being str
         else:
             result = self.submission.run(self.testcase.in_path, self.out_path)
             if result.duration > self.problem.settings.timelimit:
-                result.verdict = 'TIME_LIMIT_EXCEEDED'
+                result.verdict = Verdict.TIME_LIMIT_EXCEEDED
                 if result.timeout_expired:
                     result.print_verdict_ = 'TLE (aborted)'
             elif result.status == ExecStatus.ERROR:
-                result.verdict = 'RUN_TIME_ERROR'
+                result.verdict = Verdict.RUNTIME_ERROR
                 if config.args.error:
                     result.err = 'Exited with code ' + str(result.returncode) + ':\n' + result.err
                 else:
@@ -59,17 +60,17 @@ class Run:
                 if result is None:
                     error(f'No output validators found for testcase {self.testcase.name}')
                     result = ExecResult(None, ExecStatus.REJECTED, 0, False, None, None)
-                    result.verdict = 'VALIDATOR_CRASH'
+                    result.verdict = Verdict.VALIDATOR_CRASH
                 else:
                     result.duration = duration
 
                     if result.status:
-                        result.verdict = 'ACCEPTED'
+                        result.verdict = Verdict.ACCEPTED
                     elif result.status == ExecStatus.REJECTED:
-                        result.verdict = 'WRONG_ANSWER'
+                        result.verdict = Verdict.WRONG_ANSWER
                     else:
                         config.n_error += 1
-                        result.verdict = 'VALIDATOR_CRASH'
+                        result.verdict = Verdict.VALIDATOR_CRASH
 
             # Delete .out files larger than 1MB.
             if (
@@ -128,29 +129,33 @@ class Submission(program.Program):
         # interactor exits first. Thus, we don't distinguish the different non-AC
         # verdicts.
         if self.problem.interactive and (is_windows() or is_bsd()):
-            wrong_verdicts = ['WRONG_ANSWER', 'TIME_LIMIT_EXCEEDED', 'RUN_TIME_ERROR']
+            wrong_verdicts = [
+                Verdict.WRONG_ANSWER,
+                Verdict.TIME_LIMIT_EXCEEDED,
+                Verdict.RUNTIME_ERROR,
+            ]
             for wrong_verdict in wrong_verdicts:
                 if wrong_verdict in self.expected_verdicts:
                     self.expected_verdicts += wrong_verdicts
                     break
 
-    def _get_expected_verdicts(self):
-        verdicts = []
+    def _get_expected_verdicts(self) -> list[Verdict]:
+        expected_verdicts = []
 
         # Look for '@EXPECTED_RESULTS@: ' in all source files. This should be followed by a comma separated list of the following:
         # - ACCEPTED / CORRECT
         # - WRONG_ANSWER / WRONG-ANSWER / NO-OUTPUT
         # - TIME_LIMIT_EXCEEDED / TIMELIMIT
         # - RUN_TIME_ERROR / RUN-ERROR
-        domjudge_verdict_map = {
-            'CORRECT': 'ACCEPTED',
-            'WRONG-ANSWER': 'WRONG_ANSWER',
-            'TIMELIMIT': 'TIME_LIMIT_EXCEEDED',
-            'RUN-ERROR': 'RUN_TIME_ERROR',
-            'NO-OUTPUT': 'WRONG_ANSWER',
-            'CHECK-MANUALLY': None,
-            'COMPILER-ERROR': None,
-        }
+        domjudge_verdicts = [
+            'CORRECT',
+            'WRONG-ANSWER',
+            'TIMELIMIT',
+            'RUN-ERROR',
+            'NO-OUTPUT',
+            'CHECK-MANUALLY',
+            'COMPILER-ERROR',
+        ]
         # Matching is case insensitive and all source files are checked.
         key = '@EXPECTED_RESULTS@: '
         if self.path.is_file():
@@ -168,16 +173,15 @@ class Submission(program.Program):
                 endpos = text.find('\n', beginpos)
                 arguments = map(str.strip, text[beginpos:endpos].split(','))
                 for arg in arguments:
-                    if arg in domjudge_verdict_map:
-                        arg = domjudge_verdict_map[arg]
-                        if arg is None:
-                            continue
-                    if arg not in config.VERDICTS:
+                    argverd: Verdict = from_string(arg)
+                    if argverd is None:
+                        continue
+                    if argverd not in config.VERDICTS:
                         error(
-                            f'@EXPECTED_RESULTS@: `{arg}` for submission {self.short_path} is not valid'
+                            f'@EXPECTED_RESULTS@: `{argverd}` for submission {self.short_path} is not valid'
                         )
                         continue
-                    verdicts.append(arg)
+                    expected_verdicts.append(argverd)
                 break
             except (UnicodeDecodeError, ValueError):
                 # Skip binary files.
@@ -187,20 +191,18 @@ class Submission(program.Program):
         if len(self.path.parts) >= 3 and self.path.parts[-3] == 'submissions':
             # Submissions in any of config.VERDICTS should not have `@EXPECTED_RESULTS@: `, and vice versa.
             # See https://github.com/DOMjudge/domjudge/issues/1861
-            subdir = self.short_path.parts[0].upper()
+            subdir = from_string(self.short_path.parts[0].upper())
             if subdir in config.VERDICTS:
-                if len(verdicts) != 0:
+                if len(expected_verdicts) != 0:
                     warn(f'@EXPECTED_RESULTS@ in submission {self.short_path} is ignored.')
-                verdicts = [subdir]
+                expected_verdicts = [subdir]
             else:
-                if len(verdicts) == 0:
+                if len(expected_verdicts) == 0:
                     error(
                         f'Submission {self.short_path} must have @EXPECTED_RESULTS@. Defaulting to ACCEPTED.'
                     )
 
-        if len(verdicts) == 0:
-            verdicts = ['ACCEPTED']
-        return verdicts
+        return expected_verdicts or [Verdict.ACCEPTED]
 
     # Run submission on in_path, writing stdout to out_path or stdout if out_path is None.
     # args is used by SubmissionInvocation to pass on additional arguments.
@@ -265,7 +267,7 @@ class Submission(program.Program):
             localbar = bar.start(run)
             result = run.run(localbar)
 
-            if result.verdict == 'ACCEPTED' and not self.problem.interactive:
+            if result.verdict == Verdict.ACCEPTED and not self.problem.interactive:
                 validate.sanity_check(run.out_path, localbar, strict_whitespace=False)
 
             if verdict_table is not None:
@@ -273,7 +275,7 @@ class Submission(program.Program):
             verdicts[run.name] = result.verdict
             verdicts.duration[run.name] = result.duration
 
-            got_expected = result.verdict in ['ACCEPTED'] + self.expected_verdicts
+            got_expected = result.verdict in [Verdict.ACCEPTED] + self.expected_verdicts
 
             # Print stderr whenever something is printed
             if result.out and result.err:
@@ -315,7 +317,7 @@ class Submission(program.Program):
 
             # - for TLE, the run was aborted because the global timeout expired
             # TODO: What does this mean?
-            if result.verdict == 'TIME_LIMIT_EXCEEDED' and not result.timeout_expired:
+            if result.verdict == Verdict.TIME_LIMIT_EXCEEDED and not result.timeout_expired:
                 return
 
         p = parallel.new_queue(lambda run: process_run(run, p), pin=True)
@@ -324,7 +326,7 @@ class Submission(program.Program):
         p.done()
 
         self.verdict = verdicts['.']
-        self.duration = 42
+        self.duration = 42  # TODO this is messed up
 
         # Use a bold summary line if things were printed before.
         if bar.logged:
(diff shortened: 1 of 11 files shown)


bin/verdicts.py
```diff
@@ -0,0 +1,394 @@
...
+class Verdict(Enum):
+    """The verdict of a testcase or testgroup"""
+
+    ACCEPTED = 1
+    WRONG_ANSWER = 2
+    TIME_LIMIT_EXCEEDED = 3
+    RUN_TIME_ERROR = 4
+
+
+def to_char(v: Verdict | None):
```
mzuenni, line 20: can we add a flag to get small letters instead and then remove `VerdictTable.colors` and update `VerdictTable._get_verdict` ?

thorehusfeldt: That’s the direction I want this to go in. But let me hold back a bit before introducing too many dependencies.

bin/verdicts.py
```diff
@@ -0,0 +1,452 @@
...
+
+class Verdict(Enum):
+    """The verdict of a testcase or testgroup"""
+
+    ACCEPTED = 1
+    WRONG_ANSWER = 2
+    TIME_LIMIT_EXCEEDED = 3
+    RUNTIME_ERROR = 4
+    VALIDATOR_CRASH = 5
+    COMPILER_ERROR = 5
```
mzuenni, line 19: why are they both 5?

bin/verdicts.py
```diff
@@ -0,0 +1,452 @@
...
+    }[v]
+
+
+def to_char(v: Verdict | None, lower=False):
+    return {
+        Verdict.ACCEPTED: f'{Fore.GREEN}A{Style.RESET_ALL}',
+        Verdict.WRONG_ANSWER: f'{Fore.RED}W{Style.RESET_ALL}',
+        Verdict.TIME_LIMIT_EXCEEDED: f'{Fore.MAGENTA}T{Style.RESET_ALL}',
+        Verdict.RUNTIME_ERROR: f'{Fore.YELLOW}R{Style.RESET_ALL}',
+        None: f'{Fore.BLUE}?{Style.RESET_ALL}',
```
mzuenni, line 47: Why do to_string and to_char not handle all verdicts?

bin/verdicts.py
```diff
@@ -0,0 +1,452 @@
...
+        return {
+            Verdict.ACCEPTED: 'ACCEPTED',
+            Verdict.WRONG_ANSWER: 'WRONG ANSWER',
+            Verdict.TIME_LIMIT_EXCEEDED: 'TIME LIMIT EXCEEDED',
+            Verdict.RUNTIME_ERROR: 'RUNTIME ERROR',
+            Verdict.COMPILER_ERROR: 'COMPILER ERROR',
+        }[self]
+
+
+def to_string(v: Verdict | None):
```
mzuenni, line 31: wouldnt it be simpler to add something like `get_color` and then implement this as f`{get_colot}{str(v)}{Style.RESET_ALL}` and  f`{get_colot}{str(v)[0]}{Style.RESET_ALL}`

bin/verdicts.py
```diff
@@ -0,0 +1,452 @@
...
+
+def from_string(s: str) -> Verdict:
+    match s:
+        case 'CORRECT' | 'ACCEPTED' | 'AC':
+            return Verdict.ACCEPTED
+        case 'WRONG-ANSWER' | 'WRONG_ANSWER' | 'WA':
+            return Verdict.WRONG_ANSWER
+        case 'TIMELIMIT' | 'TIME_LIMIT_EXCEEDED' | 'TLE':
+            return Verdict.TIME_LIMIT_EXCEEDED
+        case 'RUN-ERROR' | 'RUN_TIME_ERROR' | 'RUNTIME_ERROR' | 'RTE':
```
mpsijm, line 59: I appreciate your effort to fix this long-standing grammatical error, but DOMjudge only supports `RUN-ERROR` and `RUN_TIME_ERROR`: https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Service/SubmissionService.php#L40

Similarly, the short-hands `AC`/`WA`/... will not be supported by DOMjudge (unless we send them a PR, of course).

bin/verdicts.py
```diff
@@ -0,0 +1,452 @@
...
+            return Verdict.RUNTIME_ERROR
+        case 'NO-OUTPUT':
+            return Verdict.WRONG_ANSWER
+        case 'COMPILER-ERROR':
+            return Verdict.COMPILER_ERROR
+        case 'CHECK-MANUALLY':
+            raise NotImplementedError
+        case _:
+            return None
+            # raise ValueError(f"Unknown verdict string {s}")
```
mpsijm, line 69: Why have an explicit error raised for `CHECK-MANUALLY`, while any other verdict simply returns `None`? What's the reason for not simply always raising a `ValueError` for unknown verdict strings?

bin/run.py
```diff
@@ -127,29 +129,33 @@ def __init__(self, problem, path, skip_double_build_warning=False):
...
-        }
+        domjudge_verdicts = [
+            'CORRECT',
+            'WRONG-ANSWER',
+            'TIMELIMIT',
+            'RUN-ERROR',
+            'NO-OUTPUT',
+            #'CHECK-MANUALLY', TODO this is *not* supported, shouldn't pretend it is
+            'COMPILER-ERROR',
+        ]
```
mpsijm, line 158: This list is not needed anymore, now that we have `verdicts.from_string`, right?

bin/run.py
```diff
@@ -319,28 +315,18 @@ def process_run(run, p):
...
-                return
             # - for TLE, the run was aborted because the global timeout expired
-            if result.verdict == 'TIME_LIMIT_EXCEEDED' and not result.timeout_expired:
+            # TODO: What does this mean?
+            if result.verdict == Verdict.TIME_LIMIT_EXCEEDED and not result.timeout_expired:
                 return
 
-            bar.count = None
-            p.abort()
-
```
mpsijm, line 335: RE `TODO: What does this mean?`:

You moved the `bar.count = None` and `p.abort()` up. If you read it like this:
```py
            if result.verdict != 'TIME_LIMIT_EXCEEDED' or result.timeout_expired:
                bar.count = None
                p.abort()
```
So, in the current implementation on `master`, we abort TLE submissions _only_ when the hard timeout expired. This new implementation also aborts submissions that run just a single millisecond over the problem's time limit.

It looks like you moved the code that aborts the remaining test case runs to https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/358/files#diff-dcd3fdd39ac1d0938a32c9c9c9ef70fc8b00f9d336d0676e30f159ee0e8d2314R263-R265. Can you re-add here that TLE submissions only abort once they reach the hard timeout? :slightly_smiling_face:

bin/run.py
```diff
@@ -319,28 +315,18 @@ def process_run(run, p):
...
         p = parallel.new_queue(lambda run: process_run(run, p), pin=True)
         for run in runs:
             p.put(run)
         p.done()
 
-        self.verdict = verdict[1]
-        self.print_verdict = verdict[2]
-        self.duration = max_duration
+        self.verdict = verdicts['.']
+        self.duration = 42  # TODO this is messed up
```
mpsijm, line 329: This can probably be removed? :stuck_out_tongue: I don't see the `42` ending up in my console so I guess that `self.duration` not needed? :joy: 

(it's nice that the `duration` bookkeeping is now part of `verdicts`, btw :smile:)

bin/verdicts.py
```diff
@@ -0,0 +1,452 @@
...
+            verdict = from_string(verdict)
+        if verdict == Verdict.ACCEPTED and duration is not None:
+            self.duration[testcase] = duration
+        self._set_verdict_for_node(testcase, verdict)
+
+    def __getitem__(self, testnode) -> Verdict | None:
+        return self._verdict[testnode]
+
+    def salient_testcase(self) -> str:
+        """The testcase most salient to the root verdict. If self['.'] this is the
```
mpsijm, line 158:
```suggestion
        """The testcase most salient to the root verdict. If self['.'] is AC, this is the
```

bin/run.py
```diff
@@ -354,9 +340,13 @@ def process_run(run, p):
             color = Fore.GREEN if self.verdict in self.expected_verdicts else Fore.RED
             boldcolor = ''
 
+        max_duration = max(d for d in verdicts.duration.values() if d is not None)
+        salient_testcase = verdicts.salient_testcase()
         printed_newline = bar.finalize(
-            message=f'{max_duration:6.3f}s {color}{self.print_verdict:<20}{Style.RESET_ALL} @ {verdict_run.testcase.name}'
+            message=f'{max_duration:6.3f}s {color}{self.verdict:<20}{Style.RESET_ALL} @ {salient_testcase}'
```
mpsijm, line 346: Instead of separately printing the `max_duration` and the name of the `salient_testcase`, shouldn't we print the time of the `salient_testcase` (since the printed time and name ought to be connected)? :thinking: The advantage here is, that the salient test case happens to be the one with the max duration anyway (at least, in the case that the root verdict is AC) :smile:

bin/verdicts.py
```diff
@@ -0,0 +1,452 @@
...
+        if isinstance(verdict, str):
+            verdict = from_string(verdict)
+        if verdict == Verdict.ACCEPTED and duration is not None:
+            self.duration[testcase] = duration
+        self._set_verdict_for_node(testcase, verdict)
+
+    def __getitem__(self, testnode) -> Verdict | None:
+        return self._verdict[testnode]
+
+    def salient_testcase(self) -> str:
```
mpsijm, line 157: Related to my [previous comment](https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/358#discussion_r1527228362): we probably want the salient test case to be the slowest test case when any sub-verdict is TLE as well? Note that TLE may not be the first rejected verdict, especially with non-lazy judging. So, for the verdicts `AAAWWTWTW`, I would expect the salient test case to be the slowest `T`, not simply the first `W`.

bin/verdicts.py
```diff
@@ -0,0 +1,452 @@
...
+    """
+
+    def unknowns_iterator(self, node):
+        """Yield the node's (yet) unknown children in lexicographic order."""
+        for child in sorted(self.children[node]):
+            if self._verdict[child] is not None:
+                continue
+            yield child
+
+    def __init__(self, testcase_list: list[str]):
```
mpsijm, line 111: (nitpick) Can we make `__init__` the first method in the `class`?

bin/verdicts.py
```diff
@@ -0,0 +1,452 @@
...
+        """
+        return node in self.children
+
+    def is_testcase(self, node) -> bool:
+        """Is the given testnode name a testcase (rather than a testgroup)?
+        This assumes nonempty testgroups.
+        """
+        return node not in self.children
+
+    def __setitem__(self, testcase, verdict: str | Verdict, duration=None):
```
mpsijm, line 141: Can `__setitem__` even have a third argument? https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/358/files#diff-dcd3fdd39ac1d0938a32c9c9c9ef70fc8b00f9d336d0676e30f159ee0e8d2314R276 doesn't use it either :stuck_out_tongue:

bin/verdicts.py
```diff
@@ -0,0 +1,452 @@
...
+    def finalize(self, *, print_done=True, message=None):
+        with self:
+            res = super().finalize(print_done=print_done, message=message)
+            self.table._clear(force=True)
+            return res
+
+
+# if __name__ == "__main__":
+#    import doctest
+#    doctest.testmod()
```
mpsijm, line 452: Can this be removed, or can we use this in `pytest` somehow?

bin/verdicts.py
thorehusfeldt: A mistake.

bin/verdicts.py
thorehusfeldt: Oh, but `from_string` is only used to translate Various Other Naming Conventions to the internal `verdicts.Verdict`. Never in the other direction. The ambition is to _allow_ VANC to continue to exist.

bin/verdicts.py
thorehusfeldt: Thank you. Changed back in `d6abe66` to `ValueError`, the `None` was a temporary fix to get various tests to not crash on me.

bin/run.py
thorehusfeldt: Removed in 8e91a088c. Fingers crossed.

bin/verdicts.py
RagnarGrootKoerkamp: This is a tricky one, but since the spec is not quite clear that we break on the first failing case, consistently showing that one would be the simple & straightforward way.
Alternatively we could show the slowest failed case. But this would be a bit inconsistent since it's only due to parallelism that we run any cases other than the first failing one at all :thinking:

bin/verdicts.py
mpsijm: That's fair! In that case, I've updated `Submission._get_expected_verdicts` to enforce compatibility with DOMjudge :slightly_smiling_face:

thorehusfeldt:
@mpsijm : thanks for `[28f9c45](https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/358/commits/28f9c457ef27aabe446db00670f0205adc59806e)`, I was writing almost the same code.

The directory `mixed` remains a problem and currently breaks `pytest` on `hello`.

mpsijm:
@thorehusfeldt No problem! :smile: And it looks like the `mixed` directory is processed just fine, there was just an empty `@EXPECTED_RESULTS@:` string in one of the submissions. :joy: Let's see if that fixes it :slightly_smiling_face:

bin/verdicts.py
mzuenni: maybe we don't want `salient` at this point but, multiple methods for, slowest, most memory, first wa, first rte, first tle and then in the UI decide what we want to display?

RagnarGrootKoerkamp: I think more specific error reporting is nice but it can be done in a separate PR.

mpsijm:
I've re-implemented lazy judging and changed it from "stop as soon as parent verdict is known" to "stop when one of the runs gets aborted" (unless `-v` or `--table` is given, of course). While I was at it, I also changed the displayed run time to be the slowest rejected test case (or, if the submission is AC, the slowest AC test case). Does anybody wants to do a double-check before we merge? :slightly_smiling_face:

bin/run.py
```diff
@@ -251,35 +244,37 @@ def run_all_testcases(
...
-            )
-            if new_verdict > verdict:
-                verdict = new_verdict
-                verdict_run = run
-            max_duration = max(max_duration, result.duration)
-
             if verdict_table is not None:
                 verdict_table.finish_testcase(run.name, result.verdict)
+            verdicts[run.name] = result.verdict
+            verdicts.duration[run.name] = result.duration
```
RagnarGrootKoerkamp, line 275: this also needs a (write) lock.

bin/verdicts.py
```diff
@@ -0,0 +1,463 @@
...
+            case _:
+                return min(
+                    (k, v)
+                    for k, v in sorted(self._verdict.items())
+                    if self.is_testcase(k) and v is not Verdict.ACCEPTED
+                )[0]
+
+    def aggregate(self, testgroup: str) -> Verdict:
+        """The aggregate verdict at the given testgroup.
+        Computes the lexicographically first non-accepted verdict.
```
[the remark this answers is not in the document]
mzuenni: lexicographically first sounds better since it is more consistent/reproduceable.

bin/run.py
```diff
@@ -251,35 +244,37 @@ def run_all_testcases(
...
+            # - table mode
+            #
+            # When the parent verdict is TLE, do continue when the timeout has not been reached.
+            def verdict_and_salient_case_known(parent):
+                if verdicts[str(parent)] is None:
+                    return False
+                if verdicts[str(parent)] == Verdict.TIME_LIMIT_EXCEEDED:
+                    children = c for c in verdicts.children[str(parent)] if verdicts.is_testcase(c)
+                    return any(verdicts.duration[str(c)] > self.problem.settings.timeout for c in children)
+                return True
```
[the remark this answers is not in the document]
mpsijm: So this stops at the first WA? What if the submission turns out to be mixed WA/TLE, I guess you'd still want to know what the slowest test case is in that case, until some run is aborted?

bin/verdicts.py
```diff
@@ -0,0 +1,496 @@
...
+                    return min(
+                        tc
+                        for tc, v in self._verdict.items()
+                        if self.is_testcase(tc) and v != Verdict.ACCEPTED
+                    )
+
+    def slowest_testcase(self) -> str:
+        """The slowest testcase."""
+        with self:
+            return max((v, tc) for tc, v in self.duration.items() if v is not None)[1]
```
mpsijm, line 225: Why only return the name of the slowest test case? At the (only) use site, the duration is needed anyway :slightly_smiling_face: 

(Same for `salient_testcase`)

RagnarGrootKoerkamp: Because this is cleaner, in that it's clear that the shown duration actually corresponds to the testcase instead of being some other duration.

mpsijm (approved):

PR #363, opened 2024-03-26
Feat/aggresive

mzuenni:
This changes the following things:

1. python3 is the default. Python2 is deprecated anyway and even problemtools changed their default
2. we no longer warn for missing UUID, we just add them (this even works without ruayaml)
3. Made the hash more random in case the new `count` keyword is used. (Nothing changed for rules without count)
4. the solution in `generator.yaml` is now mandatory. If it is missing we pick a solution and add the missing entry (again without ruayaml).

Please note that the previous behavior for 4 was buggy. We picked a random submission but overwrote the hashing therefore the `.ans` file weren't necessarily created by the solution we displayed but by some other solution on a previous run...

### DIFF
--- a/bin/generate.py
+++ b/bin/generate.py
@@ -507,8 +507,9 @@ class TestcaseRule(Rule):
 
                     # TODO: Should the seed depend on white space? For now it does, but
                     # leading and trailing whitespace is stripped.
-                    seed_value = self.config.random_salt + yaml['generate'].strip()
-                    self.seed = (int(hash_string(seed_value), 16) + count_index * 2**16) % 2**31
+                    salt = self.config.random_salt + ('' if count_index == 0 else f':{count_index}')
+                    seed_value = salt + yaml['generate'].strip()
+                    self.seed = int(hash_string(salt + yaml['generate'].strip()), 16) % 2**31
                     self.in_is_generated = True
                     self.rule['gen'] = self.generator.command_string
                     if self.generator.uses_seed:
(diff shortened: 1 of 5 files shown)


bin/generate.py
```diff
@@ -245,46 +245,55 @@ def run_interactive(self, problem, bar, cwd, t):
...
         # Use one of the accepted submissions.
         solutions = list(glob(problem.path, 'submissions/accepted/*'))
         if len(solutions) == 0:
             fatal(f'No solution specified and no accepted submissions found.')
-        if generator_config.has_yaml:
-            # Note: we explicitly random shuffle the submission that's used to generate answers to
-            # encourage setting it in generators.yaml.
+
+        # always try to take the same solution to not mess with hashing
+        if stored_solution.is_file():
```
RagnarGrootKoerkamp, line 271: This is only needed when `!generator_config.has_yaml`, right? So can be moved into the `else:` at line 290 below.

RagnarGrootKoerkamp (commented):
All nice changes :)

Good idea to just write stuff that's missing directly; simplifies things a lot as a user since you can just modify what's there instead of coming up with the right syntax yourself.

bin/generate.py
mzuenni: I want to pick the same solution again even when the user deletes the line from the `generator.yaml`

RagnarGrootKoerkamp: ah right sure; feels a bit redundant since you agressively add it back in but doesn't really hurt I suppose :)

PR #364, opened 2024-03-27
[run] #230 Add `--sanitizer` flag

### LINKED ISSUE #230: Compile generators/validators separately from submissions
Not sure how, but it would be nice to set compilation flags/commands for submissions only

mzuenni:
Fixes #230.

### DIFF
--- a/bin/program.py
+++ b/bin/program.py
@@ -63,6 +63,9 @@ def languages():
 def sanitizer():
     global _sanitizer, _program_config_lock
     with _program_config_lock:
+        if _sanitizer is not None:
+            return _sanitizer
+
         # Read sanitizer extra flags
         _sanitizer = parse_yaml(SANITIZER_FLAGS)
 
--- a/bin/program.py
+++ b/bin/program.py
@@ -29,14 +29,20 @@ manual:
     run: '{run}'
 '''
 
+SANITIZER_FLAGS = '''
+cpp:
+    compile: -fsanitize=undefined
+'''
+
 # The cached languages.yaml for the current contest.
 _languages = None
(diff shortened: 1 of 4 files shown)


mpsijm (approved):
Looking good! I also tried running `bt run` on some problem, with and without the `--sanitizer` flag, and it properly recompiles the submissions between the runs :slightly_smiling_face: I don't know how to trigger sanitizer warnings, but that will probably work fine, since this also worked with the old `--cpp_flags=-fsanatize=undefined`. :smile:

bin/program.py
```diff
@@ -29,14 +29,20 @@
     run: '{run}'
 '''
 
+SANITIZER_FLAGS = '''
+cpp:
+    compile: -fsanitize=undefined
```
RagnarGrootKoerkamp, line 34: wanna add `,address` also here?

mzuenni: i get an error if i add this...
```
Exited with code -6:
==618896==ERROR: AddressSanitizer failed to allocate 0xdfff0001000 (15392894357504) bytes at address 2008fff7000 (errno: 12)
==618896==ReserveShadowMemoryRange failed while trying to map 0xdfff0001000 bytes. Perhaps you're using ulimit -v
```

RagnarGrootKoerkamp: ah yes, `ulimit` / memory limit should be disabled for that. But that probably makes sense to do right?

mzuenni: The issue seems to be that we set a memory limit. If we want to have `-fsanitize=address` we have to skip the memory limit. I can add that if you want?

PR #369, opened 2024-04-14
Multipass Support

mzuenni:
TODO:

- [x] parse problem.yaml
- [x] change generate
- [x] change validate
- [x] change run
- [x] change run interactive
- [x] change test
- [x] change test interactive
- [x] handle latex multipass samples
- [x] handle latex interactive multipass samples
- [x] improve verdict ui (show pass)
- [ ] ???


bin/validate.py
```diff
@@ -300,22 +300,25 @@ def run(self, testcase, mode, constraints=None, args=None):
...
+            # mode is actually a run
             path = mode.out_path
+            in_path = mode.in_path
 
         if self.language in Validator.FORMAT_VALIDATOR_LANGUAGES:
             raise ValueError("Invalid output validator language")
 
         cwd, constraints_path, arglist = self._run_helper(testcase, constraints, args)
-        feedbackdir = cwd if isinstance(mode, Mode) else mode.feedbackdir
+        if not isinstance(mode, Mode):
```
RagnarGrootKoerkamp, line 311:
```suggestion
        if isinstance(mode, run.Run):
```

bin/problem.py
```diff
@@ -733,12 +740,13 @@ def validate_data(problem, mode: validate.Mode, constraints: dict | bool | None
...
             case validate.Mode.ANSWER:
                 assert not problem.interactive
+                assert not problem.multipass
                 problem.validators(validate.AnswerValidator, check_constraints=check_constraints)
                 problem.validators(validate.OutputValidator, check_constraints=check_constraints)
                 testcases = problem.testcases(mode=mode)
             case validate.Mode.INVALID:
                 problem.validators(validate.InputValidator)
-                if not problem.interactive:
+                if not problem.interactive and not problem.multipass:
```
RagnarGrootKoerkamp, line 749: I see a lot of `problem.interactive or problem.multipass`. Should we introduce `problem.simple` or so?
(Really not quite sure though; it's hard to find a good name for it and clarity may be lost. Keeping as it is is also fine.)

bin/util.py
```diff
@@ -810,6 +810,27 @@ def write_yaml(data, path, allow_yamllib=False):
         )
 
 
+# Parse validation mode
+def parse_validation(mode):
```
RagnarGrootKoerkamp, line 814: The plan is to drop `validation: custom/default/custom interactive` from `problem.yaml` and infer this via the presence of `output_validator/`, and move `interactive` and `multipass` to a new field `type: [list of, interactive, pass-fail, scoring, multipass]`.

bin/run.py
```diff
@@ -20,18 +21,24 @@ def __init__(self, problem, submission, testcase):
...
+            / 'runs'
+            / self.submission.short_path
+            / self.testcase.short_path.with_suffix('')
         )
-        self.out_path = tmp_path.with_suffix('.out')
-        self.feedbackdir = tmp_path.with_suffix('.feedbackdir')
+
+        self.in_path = self.tmpdir / 'testcase.in'
+        self.out_path = self.tmpdir / 'testcase.out'
+        self.feedbackdir = self.in_path.with_suffix('.feedbackdir')
```
RagnarGrootKoerkamp, line 33:
```suggestion
        self.feedbackdir = self.tmpdir / 'testcase.feedbackdir'
```

for consistency?

bin/run.py
```diff
@@ -40,35 +47,54 @@ def run(self, bar, *, interaction=None, submission_args=None):
...
+                else:
+                    # Overwrite the result with validator returncode and stdout/stderr, but keep the original duration.
+                    duration = result.duration
+                    result = self._validate_output(bar)
+                    if result is None:
+                        error(f'No output validators found for testcase {self.testcase.name}')
+                        result = ExecResult(None, ExecStatus.REJECTED, 0, False, None, None)
                         result.verdict = 'VALIDATOR_CRASH'
+                    else:
+                        result.duration = duration
```
RagnarGrootKoerkamp, line 74: Take a `max` here?

bin/problem.py
mzuenni: i think both makes clearer what's going on... for now. And i am not sure if i have to undo some places later ^^

bin/util.py
mzuenni: yes, we can do that later, i guess?

bin/run.py
mzuenni: the current version is consistent with the other places where we open the feedbackdir, that's what i would prefer?

bin/run.py
mzuenni: ahm I didn't change this... But I also think this is right. Why should the validator run time be relevant here?

bin/validate.py
mzuenni: hm, will undo this again... cyclic imports...

mzuenni:
@RagnarGrootKoerkamp @mpsijm is there more to do here? (besides that this is not properly tested...)

RagnarGrootKoerkamp (approved):
Didn't look super closely but looks good overall!

Maybe we can add a trivial example `multipass` or (and?) `multipass interactive` problem to the `test/problems` directory as well?

bin/interactive.py
```diff
@@ -150,7 +150,7 @@ def get_validator_command():
             validator_err.decode('utf-8', 'replace'),
             exec_res.err,
             verdict,
-            last_pass,
+            last_pass if run.problem.multipass else None,
```
RagnarGrootKoerkamp: It feels cleaner to always set this and instead do this `if` check where it's used/displayed.

bin/run.py
```diff
@@ -41,34 +48,92 @@ def run(self, bar, *, interaction=None, submission_args=None):
...
+                elif result.status == ExecStatus.ERROR:
+                    result.verdict = Verdict.RUNTIME_ERROR
+                    if config.args.error:
+                        result.err = (
+                            'Exited with code ' + str(result.returncode) + ':\n' + result.err
+                        )
+                    else:
+                        result.err = 'Exited with code ' + str(result.returncode)
+
+                if interaction:
```
RagnarGrootKoerkamp, line 78: add a comment here what this is doing, and when/how it is triggered.

Also, can you move this right below `result = self.submission.run(..)`?

Then the if-statement on line 91 can be simplified to just early-break in the `RTE` case above, and a conditional early break in the `TLE` case above.

bin/run.py
```diff
@@ -77,9 +142,39 @@ def run(self, bar, *, interaction=None, submission_args=None):
...
+            bar.warn(f'Validator created nextpass.in for non multipass problem. Ignored.')
+
         self.result = result
         return result
 
+    # check if we should continue after tle
+    def _continue_with_tle(self, verdict, timeout_expired):
+        if not self.problem.multipass:
+            return False
+        elif verdict != Verdict.TIME_LIMIT_EXCEEDED:
```
RagnarGrootKoerkamp, line 156: `elif` => `if` 2x

bin/run.py
```diff
@@ -314,7 +410,12 @@ def process_run(run):
                 color = Fore.GREEN if got_expected else Fore.RED
             timeout = result.duration >= self.problem.settings.timeout
             duration_style = Style.BRIGHT if timeout else ''
-            message = f'{color}{result.verdict.short():>3}{duration_style}{result.duration:6.3f}s{Style.RESET_ALL} @ {run.name:{max_testcase_len}}'
+            passmsg = (
+                f':{Fore.CYAN}{result.pass_id}{Style.RESET_ALL}'
+                if self.problem.multipass and not result
```
RagnarGrootKoerkamp, line 415: `result is None`?

bin/run.py
```diff
@@ -41,34 +48,92 @@ def run(self, bar, *, interaction=None, submission_args=None):
...
+                        if nextpass and nextpass.is_file():
+                            bar.error(f'got WRONG_ANSWER but found nextpass.in', resume=True)
+                            result.verdict = Verdict.VALIDATOR_CRASH
                     else:
                         config.n_error += 1
                         result.verdict = Verdict.VALIDATOR_CRASH
 
+                    if result.verdict != Verdict.ACCEPTED:
+                        break
+                else:
```
RagnarGrootKoerkamp, line 117: `if ...: ... else: break` => can you make it an early break and drop the `else:`?

bin/run.py
```diff
@@ -41,34 +48,92 @@ def run(self, bar, *, interaction=None, submission_args=None):
...
                         result.verdict = Verdict.WRONG_ANSWER
+                        if nextpass and nextpass.is_file():
+                            bar.error(f'got WRONG_ANSWER but found nextpass.in', resume=True)
+                            result.verdict = Verdict.VALIDATOR_CRASH
                     else:
                         config.n_error += 1
                         result.verdict = Verdict.VALIDATOR_CRASH
 
+                    if result.verdict != Verdict.ACCEPTED:
+                        break
```
RagnarGrootKoerkamp, line 116: Maybe also move this break to right below whereever a non-AC verdict is set?

RagnarGrootKoerkamp (commented):
Mostly looks good! Just some comments on style/code clarity really

bin/run.py
```diff
@@ -41,34 +48,92 @@ def run(self, bar, *, interaction=None, submission_args=None):
...
+                        result = ExecResult(
+                            None, ExecStatus.REJECTED, 0, False, None, None, Verdict.VALIDATOR_CRASH
+                        )
+                    elif result.status:
                         result.verdict = Verdict.ACCEPTED
+                        validate.sanity_check(self.out_path, bar, strict_whitespace=False)
                     elif result.status == ExecStatus.REJECTED:
                         result.verdict = Verdict.WRONG_ANSWER
+                        if nextpass and nextpass.is_file():
+                            bar.error(f'got WRONG_ANSWER but found nextpass.in', resume=True)
```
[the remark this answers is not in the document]
mzuenni: The spec says no and there is a good reason for this If you start to stream you cant give AC without triggering the next pass.

bin/run.py
mzuenni: but that would just mean that we have to add a break in every single case?

bin/run.py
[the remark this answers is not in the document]
RagnarGrootKoerkamp: Well I don't really see the problem. You can start writing `nextpass.in`, intending to trigger a next pass by returning `AC`, but then sometime later you realize you want to return `WA` and abort things.

But anyway it's good the spec is on the safe side here :)

bin/interactive.py
mzuenni: I think its cleaner to only set `pass_id` if we are doing a pass (and not judging a normal problem)

bin/run.py
RagnarGrootKoerkamp: Yeah. Maybe it's too much, but I think consistently `break`ing as soon as you know and set a non-AC final verdicts may simplify the code, even if it's a bit longer.

PR #374, opened 2024-05-10
[support] Improve float comparison in default_output_validator

mzuenni:
This is mostly done from my side. However, we should wait how exactly floats are defined in the spec.


RagnarGrootKoerkamp:
I'll review this end of the week. This needs some more attention to detail than I have right now.

test/default_output_validator/default_output_validator.yaml
```diff
@@ -264,7 +322,97 @@ wa:
...
+# ans 0, float compare
+flags: 'float_tolerance 0'
+ans: 0
+ac:
+  - 0
+  - '0.0'
+  - '0.00'
+  - '-0'
+  - '-0.0'
+  - '0e1'
```
RagnarGrootKoerkamp, line 353:
```suggestion
  - '0e1'
  - '0E1'
```

support/default_output_validator.cpp
```diff
@@ -2,199 +2,379 @@
...
+	}
 
-string crop_output(string output) {
-	if(output.size() > 200) {
-		output.resize(200);
-		output += " ...";
+	constexpr bool is_decimal(std::string_view token) {
+		std::size_t dot = token.find('.');
+		if (not is_integer(token.substr(0, dot))) return false;                        // decimals before the dot are *non empty* integers
+		if (dot < token.size() and not is_digits(token.substr(dot + 1))) return false; // decimals only have digits after the dot
```
RagnarGrootKoerkamp, line 66: This allows trailing dot (`0.`). Should we? I don't see a testcase to either AC or WA this, so that can be added.

support/default_output_validator.cpp
```diff
@@ -2,199 +2,379 @@
...
+		try {
+			//std::from_chars for floats is not widely supported ):
+			std::size_t pos = 0;
+			std::string tmp(token);
+			long double res = std::stold(tmp, &pos);
+			assert(pos == token.size());
+			assert(std::isfinite(res));
+			return res;
+		} catch(const std::out_of_range& /**/) {
+			// i dont know how to handle this... return +-inf?
```
RagnarGrootKoerkamp, line 116: maybe just panic? Seems safer than possibly silently ignoring values.

On the other hand if teams submit rediculously long float values we probably shouldn't crash... so maybe just saying it's not a float and doing string comparison is fine. I think when we say 'output a float' we can expect something reasonable and reject unreasonable floats.

support/default_output_validator.cpp
```diff
@@ -2,199 +2,379 @@
...
-				return {false, quick_diff(out, ans)};
-			}
-			if(p2 < w2.size()) return {false, quick_diff(w2, w1)};
+	bool equal(token_view o) const {
+		return token == o.token;
+	}
 
-			// OK if w1 and w2 represent the same (possibly nan/inf) value.
-			if(v1 == v2) continue;
+	bool insensitive_equal(token_view o) const {
```
RagnarGrootKoerkamp, line 145: rename to `case_insensitive_equal` (as opposed to 'float_insensitive_equal' or so)

RagnarGrootKoerkamp (approved):
Code is very clean and clear! Thanks :)

I just have a few small comments.

support/default_output_validator.cpp
```diff
@@ -2,199 +2,379 @@
...
+constexpr int AC = 42;
+constexpr int WA = 43;
+
+constexpr std::string_view CASE_SENSITIVE           = "case_sensitive";
+constexpr std::string_view SPACE_CHANGE_SENSITIVE   = "space_change_sensitive";
+constexpr std::string_view FLOAT_ABSOLUTE_TOLERANCE = "float_absolute_tolerance";
+constexpr std::string_view FLOAT_RELATIVE_TOLERANCE = "float_relative_tolerance";
+constexpr std::string_view FLOAT_TOLERANCE          = "float_tolerance";
+constexpr std::string_view TEXT_ELLIPSIS            = "[...]";
+constexpr std::string_view WHITESPACE               = " \f\n\r\t\v";
```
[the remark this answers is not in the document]
mzuenni: not sure, but those are typically recognized by `is_space()` functions

support/default_output_validator.cpp
```diff
@@ -2,199 +2,379 @@
...
-string strip_newline(string s) {
-	if(not s.empty() and s.back() == '\n') s.pop_back();
-	return s;
-}
+//============================================================================//
+// Utility                                                                    //
+//============================================================================//
+namespace util {
+	constexpr bool is_space(char c) {
+		return WHITESPACE.find(c) != std::string_view::npos;
```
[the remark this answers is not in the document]
mzuenni: i think so... (right now I only use c++17 features)

support/default_output_validator.cpp
mzuenni: yes, trailing dot is typically considered ok.

support/default_output_validator.cpp
mzuenni: panic would be ok for jury tokens (maybe?) but definitely not for team tokens... but yea I have no really god idea for this...

support/default_output_validator.cpp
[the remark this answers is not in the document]
RagnarGrootKoerkamp: allright; works for me then

support/default_output_validator.cpp
RagnarGrootKoerkamp: we leave it as you did then

RagnarGrootKoerkamp:
Feel free to merge! Thanks :)

PR #376, opened 2024-05-10
Latex samples

mzuenni:
#373 
This removes the old notes section, but that one was out of spec and just shortly available so it should be no problem to just remove it again

### DIFF
--- a/latex/bapc.cls
+++ b/latex/bapc.cls
@@ -93,6 +93,7 @@
 \newlistof{listofproblems}{lop}{\listofproblemsname}
 \newlistentry{problem}{lop}{0}
 \newcounter{SampleCount}
+\newcounter{NextSample}
 \renewcommand{\printloptitle}[1]{
 	\hfill
 	\sffamily
@@ -120,6 +121,7 @@
 	\ifdefempty{\timelimit}{}{{\color{darkgray}\fontfamily{qcr}\langtimelimit{}: \timelimit{}s}}
 
 	\setcounter{SampleCount}{0}
+	\setcounter{NextSample}{1}
 	\setcounter{figure}{0}
 
 	\rmfamily
@@ -153,7 +155,7 @@
 
 
 %-------------------------------------------------------------------------------
-% Environments for specifying Input, Output, Interaction, and sample Explanation/Notes.
+% Environments for specifying Input, Output and Interaction
 %-------------------------------------------------------------------------------
 \newenvironment{Input}{
 	\subsection*{\langinput{}}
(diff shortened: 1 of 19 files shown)


mzuenni:
do we want to (re)add an environment `notes` which should be used to add notes for samples?

RagnarGrootKoerkamp:
I don't think a notes section is needed anymore.

Code lgtm.

Can you explain why the fallback samples are needed? Since this only adds things (apart from removing unused notes section), shouldn't it just work for existing packages?

mpsijm:
Regarding the `Notes` section: I don't think that's necessary, you can simply write `\subsection*{Notes}` instead.

Regarding the fallback samples: I don't think any contest overrides `bapc.cls`, so I guess the fallback samples aren't needed?

latex/bapc.cls
```diff
@@ -167,17 +169,6 @@
 \newenvironment{Explanation}{
 	\subsection*{\langexplanation{}}
 }{\par}
```
mpsijm, line 171: And while we're at it, what was the `Explanation` environment used for again?

mzuenni: not sure about that, never used it

mzuenni:
the fallback is needed if you overwrite the `problem.tex` or `contest-problem.tex`. The old workflow was to first include the statement and then include the `samples.tex` and the samples.tex would directly print all problems (we override this for gcpc and nwerc). The new workflow is to first include the `samples.tex` since it only defines commands like `\Sample1` but does not print them. This is needed to make the samples available in the problem statement.

mpsijm:
But, for every sample, you call `fallback_call.append(f'\t\\csname Sample{i+1}\\endcsname\n')`. If the overridden `problem.tex` (or `contest-problem.tex`) forgets to include `samples.tex`, then `\SampleX` also is not available :thinking: 

Just to make sure we're talking about the same piece of code, I'm talking about this part:
```py
    # this is only for backwards compatibility
    samples_data += [
        '% this is only for backwards compatibility\n',
        '\\ifcsname remainingsamples\\endcsname\\else\n',
        ''.join(fallback_call),
        '\\fi\n',
    ]
```
Which I interpret as "if `\remaningsamples` is not defined, print a list of `\SampleX`". But `\remainingsamples` is always defined in `bapc.cls`. And if users _do_ use a different `.cls` file, they'll have to implement their own sample-printing anyway. We also do this for NWERC, so I just tried removing the `fallback_call` and the samples are still in the PDF :slightly_smiling_face:

mzuenni:
yes, we are speaking about the same code.

> And if users do use a different .cls file, they'll have to implement their own sample-printing anyway.

No? you could use `sample.tex` and your own `problem.cls`. (In fact, that's what we are doing for gcpc now because it's easier). And since this is possible and works fine we should include this fallback for backwards compatibility.

latex/bapc.cls
RagnarGrootKoerkamp: let's drop if you didn't already.

RagnarGrootKoerkamp:
ok all sgtm. Maybe just add some comments in the code explaining these details?
(We could even add such usage comments in the generated `.tex` file but that's not all that useful probably.)

Should we also document this somewhere (probably not since it's already in the spec)?
Or at least add/modify some examples for it? In that case probably modify the statement for the `identity` problem to do `\nextsample` twice and then `\remainingsamples` or so

latex/bapc.cls
mzuenni: but we even added this to the lang files... i would prefer to keep it actually.
And even with the new abilities it might be useful?

mzuenni:
Can we merge?

Issue #378, opened 2024-05-12
Automate exhaustive generation of small test cases


roehrt:
Often a problem has a family of test cases which is small enough to be enumerated explicitly.
Currently, enumerating these test cases individually is necessary, as illustrated below:
```yaml
- small-1: small -c 1
- small-2: small -c 2
- small-3: small -c 3
...
```

Introducing a shorthand notation for this process would be awesome.
Building upon the `{seed}` and `lo..hi` syntax, one approach could be:
```yaml
- small: small -c {1..10}
```
Alternatively, for a specified set of tokens provided as a comma-separated list:
```yaml
- small: small -c {some,important,tokens}
```

One alternative would be to allow generators to output multiple test cases in one invocation.
However, I doubt the benefit in comparison to the cost of changing generator semantics.
Also, at least I don't know of a clean way to separate the test cases.

mpsijm:
Thanks for taking the time to write an issue! See also the discussion in https://github.com/RagnarGrootKoerkamp/BAPCtools/issues/271#issuecomment-1611211270 and onwards. A few weeks ago, we discussed this between the BAPCtools developers, and we decided that it's not worth the complexity right now:

![image](https://github.com/RagnarGrootKoerkamp/BAPCtools/assets/9739541/6feb6775-47c5-4b96-9d03-5f7082c1093c)

Nonetheless, let's keep this issue open, since multiple people have asked for this.

mzuenni:
For this we could substitute `{count}` with numbers 1..n when using the count key?

RagnarGrootKoerkamp:
Oh adding the `{count}` variable sounds quite nice and like a natural extension indeed.
This is restrictive enough to still encourage thinking about actually interesting/varied arguments for generators, but general enough to allow enumerating them.

Should it count from `0` instead? Not quite sure.

mpsijm:
Supporting a `{count}` placeholder in generator invocations may be a rabbit hole for more requests. But if we're okay with that, I think it's a nice addition to help generating test cases exhaustively :slightly_smiling_face: 

I think it should start counting from `1`, based on the fact that our [styleguide](https://github.com/nwerc/docs/blob/main/styleguide.md#general-guidelines) says to use 1-based indexing where possible. If you need anything else, you can modify it within the generator itself (e.g. do `count - 1` or `2 ** count` or something else).

mzuenni:
[...]

PR #379, opened 2024-05-13
substitute {count}

mzuenni:
#378

### DIFF
--- a/bin/generate.py
+++ b/bin/generate.py
@@ -432,8 +432,7 @@ class TestcaseRule(Rule):
 
         # Used by `fuzz`
         self.in_is_generated = False
-        self.has_count = count_index is not None
-        self.count_index = count_index if self.has_count else 0
+        self.count_index = count_index
 
         # used to decide if this was supposed to be a duplicate or not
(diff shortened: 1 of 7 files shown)


mzuenni:
@thorehusfeldt it seems like the regex is wrong here? A generator is allowed to have `{name}` (and now) `{count}` any number of times?

RagnarGrootKoerkamp (approved):

mzuenni:
can someone proofread my regex? ^^'
And changes to the schemas have to be submitted somewhere else too, right?

thorehusfeldt:
The regex seems to do what we want it to. (I’ve tested a bunch of them now…) You can always write more no-instances in https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/master/test/generator_yaml/bad_generators.yaml ; the CUE schema should then reject them automagically.

The JSON schema indeed has to be submitted to schemastore.org, and that takes typically a few days. (I’ve done this so far.)

mzuenni:
Can you submit this update again? :)

PR #380, opened 2024-05-19
clean abort on ctrl+c for fuzz

mzuenni:


bin/parallel.py
```diff
@@ -163,6 +163,15 @@ def _worker(self, cores: bool | list[int] = False):
     def _interrupt_handler(self, sig, frame):
         util.fatal('Running interrupted', force=True)
 
+    def _handle_first_error(self):
+        if self.first_error is not None:
+            first_error = self.first_error
+            self.first_error = None
+            # we are the main thread now, so we can handle this
+            if isinstance(first_error, ChildProcessError):
```
RagnarGrootKoerkamp, line 171: can you expand the comment? I don't really understand this if statement and why we must forward to the original interrupt handler here. Why are we the main thread?
In which circumstances does the `if` trigger?

bin/fuzz.py
```diff
@@ -192,6 +195,21 @@ def run(self):
...
 
+        def soft_exit(sig, frame):
+            if self.queue.aborted:
+                fatal('Running interrupted', force=True)
+            else:
+                self.queue.abort()
+                with bar:
+                    bar.clearline()
+                    message(
+                        'Running interrupted (waiting on remaining tasks)\n',
```
RagnarGrootKoerkamp, line 206: add 'Press CTRL+C again to abort'? (that's what the `fatal` above does, right?)

RagnarGrootKoerkamp (approved):

bin/fuzz.py
mzuenni: yes

bin/parallel.py
mzuenni: in `util.py:exec_command` we throw a `ChildProcessError` if the child is terminated by `ctrl+c` and we are not the main process (If we are the main process we call fatal... i don't know why this distinction is made...).
Anyway, this means a `ChildProcessError` is only thrown if we use the parallel queue.
And the thread that calls `join()` or `done()` is supposed to handle this error, but there is nothing to handle we should exit since someone pressed `ctrl+c`.

bin/fuzz.py
mzuenni: not sure if we should print that extra message though... pressing ctrl+c again is somewhat dangerous and can remove the `generator.yaml` completely (this is also the reason why I changed the code in the first place)

RagnarGrootKoerkamp: I see. We could also print a warning for that. Or somehow disable the 2nd ctrl-c completely?

bin/parallel.py
RagnarGrootKoerkamp: I suppose the distinction is so that the main thread can clearly handle errors from child processes somehow.

bin/fuzz.py
mzuenni: i would not disable it if the program hangs for some reason.

RagnarGrootKoerkamp:
I think you can merge

PR #383, opened 2024-07-17
[generate] Better handling of cached visualizer output

mpsijm:
TL;DR: Based on our discussion on Slack, I have made sure that the visualizer output (i.e. files with an extension in `config.KNOWN_VISUALIZER_EXTENSIONS`, or in short: "images") is removed when the `.in` or `.ans` files change, but the images are not removed when the `.in` and `.ans` files are identical between the cache and `data/`.

Longer version:
In https://github.com/RagnarGrootKoerkamp/BAPCtools/commit/c537d9b0359f765046969399b11c032cbf982892, I wanted to make sure that the images are not simply removed when it does not exist in the cache. This may happen when running `bt generate --no-visualizer` or `bt all` with an empty cache (either after a reboot or in a CI environment). As @mzuenni correctly pointed out, this may cause the images in `data/` to be outdated when the `.in`/`.ans` files change. Therefore, this PR makes sure that the images are only kept when the `.in` and `.ans` files are identical between the cache and `data/`. If any of them is different, the images _will_ be removed.

I tested this with a problem that has a visualizer and:
- running various permutations of `bt generate`, `bt generate --no-visualizer`, and `rm -r $(bt tmp)/data`.
- changing the `solution:` in `generators.yaml` (which caused the `.ans` files to change and therefore the images to be removed when running `bt generate --no-visualizer`)
- tested that no extra checks are necessary in `TestcaseRule.link()`, by running `rm -r $(bt tmp)/data/*/*.{pdf,png}` followed by `bt generate --no-visualizer` (for the first commit on this branch)

Let me know if you can think of more scenarios and whether they currently behave as they should :smile:


mzuenni (approved):
Looks good to me, thanks :)

RagnarGrootKoerkamp (approved):

PR #384, opened 2024-08-04
Fix link to CLICS spec in readme

s7nfo:
^

Unsure if this is the exact URL you'd want, but the current one certainly 404's.


RagnarGrootKoerkamp:
thanks :)

Issue #388, opened 2024-08-26
Required python version


RagnarGrootKoerkamp:
The readme says we require 3.6, but it seems we now use a bunch of `match` statements that require 3.10.

mpsijm:
Already fixed in https://github.com/mpsijm/BAPCtools/commit/f33fcd5ebea51966c83fbd2ba0f615b0dc57ca3b, but I'm not done fixing all type errors yet :stuck_out_tongue:

RagnarGrootKoerkamp:
If you plan on merging that within the next ~month, that's fine.

Issue #102, opened 2021-04-13
Consider adding type annotations / type checking


RagnarGrootKoerkamp:
This may improve long term maintainability and give better errors during development.

mzuenni:
I think everythin is typed, so I will close this

PR #390, opened 2024-09-19
Update Python version (3.10)

Nonoze:


RagnarGrootKoerkamp:
Hey! What was the issue exactly?
I think @mpsijm has some unpushed changes somewhere to actually make things compatible again with 3.6.

mpsijm:
Ehhh no, the other way around unfortunately; I also updated the readme to require 3.10 😅 https://github.com/RagnarGrootKoerkamp/BAPCtools/commit/b21db3256ad08862f79652623c72f004631f81e5

Indeed, my branch for #102 is not ready for merging yet, but feel free to cherry-pick the linked commit already 🙂

Nonoze:
Hi, I tried to run it with python 3.8, and there were some errors (line 81
in verdicts.py) because of the function “match/case” I think, I don’t know
if you understand what I mean.

I deleted my pool request because I saw that someone has mentioned it on
the issue part.

Bye and thanks for this great tool !

Nonoze:
Oh ok, no problem it works know (I will still check that this evening)

mpsijm:
Ah, I see now that I made the same comment in https://github.com/RagnarGrootKoerkamp/BAPCtools/issues/388#issuecomment-2310933537 😂 

Hopefully upgrading to Python 3.10 isn't a problem for you! If it is, we can downgrade the match/case to plain if/else to make it compatible with older version of Python again 🙂 No guarantee that it will happen soon though, since we have some contests coming up 😇 But feel free to make a PR 😄

Nonoze:
Haha no worries it’s ok for me (I just had to reinstall my Ubuntu wsl
because of that but it’s done now and I should have done it soon so it’s
not a problem).

Good preparation of the bapc I suppose

PR #391, opened 2024-09-23
interactive.py: fix variable name

arknave:
Per the README, I wanted to let you know I've been using BAPCTools locally to develop ICPC problems for North American Regional contests, the North American Championships, and the World Finals. Thanks for making the tool. 😄 

I'm not 100% clear on the logic here, but this at least allows interactive problems to work on OSX.

Before:
```
$ bt run
Traceback (most recent call last):
  File "bin/bt", line 1079, in <module>
    main()
  File "bin/bt", line 1075, in main
    run_parsed_arguments(parser.parse_args())
  File "bin/bt", line 945, in run_parsed_arguments
    success &= problem.run_submissions()
  File "BAPCtools/bin/problem.py", line 649, in run_submissions
    submission_ok, printed_newline = submission.run_all_testcases(
  File "BAPCtools/bin/run.py", line 421, in run_all_testcases
    p.done()
  File "BAPCtools/bin/parallel.py", line 208, in done
    self._handle_first_error()
  File "BAPCtools/bin/parallel.py", line 173, in _handle_first_error
    raise first_error
  File "BAPCtools/bin/parallel.py", line 150, in _worker
    self.f(task)
  File "BAPCtools/bin/run.py", line 354, in process_run
    result = run.run(localbar)
  File "BAPCtools/bin/run.py", line 47, in run
    result = interactive.run_interactive_testcase(
  File "BAPCtools/bin/interactive.py", line 154, in run_interactive_testcase
    if not result.verdict and not self._continue_with_tle(verdict, max_duration >= timeout):
NameError: name 'result' is not defined
```

After:
```
$ bt run
<works as expected>
```

PR #392, opened 2024-09-23
#102 Part 1: Initial setup for type checking

### LINKED ISSUE #388: Required python version

mpsijm:
Took a while, but finally, here's the first batch of changes for #102 :smile: 

This MR adds `mypy` to the pre-commit hooks, which will catch some pesky type errors that will pop up every now and then. `pre-commit` will try its best to only type-check the changed files, although it may need to reference other files in order to do so. All-in-all, I'm (currently) not bothered by the small extra delay when creating a commit, but let me know if you think otherwise :slightly_smiling_face: 

Note that not everything is typed yet (for example, I'd still like to enable the `--strict` flag at some point), but it's time to at least merge this branch so that (1) others can benefit from type checking and (2) we prevent bug reports like the recent #388, #390, and #391 (this PR also fixes #388). Also, there are some TODOs left in the code, marked with `TODO #102:`. 

Sorry for the huge diff, this is going to be horrible to review :sweat_smile: However, all type errors needed to be fixed before we can enable `mypy` in the pre-commit hook and the CI. Most changes should be fairly local, though. Also, reviewing this PR one commit at a time should be doable. Some commit messages also have some extra explanation of why certain things happened, so if you're unsure about a certain part of the diff, `git blame` is your friend :stuck_out_tongue: 

Good luck and have fun with the new typings! :smile:


bin/program.py
```diff
@@ -99,7 +103,9 @@ def sanitizer():
 #
 # build() will return the (run_command, message) pair.
 class Program:
-    def __init__(self, problem: Problem, path: Path, deps=None, *, skip_double_build_warning=False):
+    def __init__(
+        self, problem: "Problem", path: Path, deps=None, *, skip_double_build_warning=False
```
RagnarGrootKoerkamp: this `"Problem"` is quite ugly. Can't we extract the class definition to a near-empty file without further dependencies?

bin/problem.py
```diff
@@ -425,19 +436,17 @@ def statement_samples(p):
 
         testcases.sort()
 
-        if len(testcases) == 0:
```
RagnarGrootKoerkamp, line 430: where/why did this go?

bin/tools.py
```diff
@@ -1043,13 +1044,16 @@ def read_personal_config():
...
+        home_config = Path(app_data)
     else:
-        home_config = (
-            Path(os.getenv('XDG_CONFIG_HOME'))
-            if os.getenv('XDG_CONFIG_HOME')
-            else Path(os.getenv('HOME')) / '.config'
-        )
+        home = os.getenv('HOME')
+        if not home:
+            fatal("$HOME does not exist")
```
RagnarGrootKoerkamp, line 1054: Return empty `args` instead of `fatal`?

bin/generate.py
```diff
@@ -1206,6 +1229,23 @@ def __init__(self, problem, key, name: str, yaml: dict = None, parent=None):
                             self.path,
                         )
 
+    @overload
```
RagnarGrootKoerkamp, line 1232: Add a comment what we're looking at here.

RagnarGrootKoerkamp:
Mostly LGTM. Didn't have a super close look, and I trust that all the types you put are correct :)
So mostly a few small comments about things I didn't understand.

The remaining TODOs you still plan to fix as part of this PR, right?

bin/problem.py
mpsijm: This moved to `latex.py`: https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/392/files#diff-03fcb90513814b3f133a9ad5bff703ee468f356c0023a6b88380f827529fa54cR44-R47 
I basically merged the two `if`-statements. The function `Problem.statement_samples()` is only used in `latex.create_samples_file()`.

bin/tools.py
mpsijm: Yeah fair, done :slightly_smiling_face: I mean, people's systems are probably utterly broken (or misusing `sudo`) if the `$HOME` or `%AppData%` variables are missing, but we don't need to judge them by throwing a fatal error :stuck_out_tongue:

bin/generate.py
mpsijm: The `@overload` definitions are basically two "versions" of the static types, and writing them like this was easier than trying to (ab)use generic types. I've added the comments, let me know if this clears things up :slightly_smiling_face:

mpsijm:
> I trust that all the types you put are correct :)

Same, because `mypy` throws no errors :smile: 

> The remaining TODOs you still plan to fix as part of this PR, right?

I'd rather merge this PR as it is right now and keep #102 open. The advantage of Python is that it supports gradual typing: not everything _needs_ to be strictly typed if you don't want to. The `TODO`s are mostly for trying to make types stricter at some later point in time. This means that you now get benefits of type checking in obvious scenarios or when you explicitly define the type. If you don't define types, you don't get type-checked right now, but that's fine :slightly_smiling_face: Another benefit of merging this now, is because I only sporadically have time to work on this. If we keep this open, I will keep getting merge conflicts if other changes are pushed to `master` :stuck_out_tongue:

bin/program.py
RagnarGrootKoerkamp: Ok well; string types it is then.

RagnarGrootKoerkamp:
lgtm. feel free to merge

mpsijm:
Thanks! :smile: 

Because it's a large MR: @mzuenni do you still want to take a look, or can we merge? :slightly_smiling_face:

mzuenni:
looks good overall but I cant check this all XD I think we should just see and if we notice something broke we can still fix it :)

mpsijm:
That's fair :stuck_out_tongue: I've been using `bt` from this branch in the past few weeks, should be fine™ :smile:

PR #393, opened 2024-09-28
Improve test multipass validator.

meisterT:
This makes the validator more robust against broken team submissions - see additional submissions.

Also add better messages so that you can understand what fails and why.

All of this sets a good example (people will copy validators) and is helpful when implementing multipass problem support (e.g. https://github.com/DOMjudge/domjudge/issues/2307).

### DIFF
--- a/test/problems/multipass/output_validators/multipass_validator/multipass_validator.py
+++ b/test/problems/multipass/output_validators/multipass_validator/multipass_validator.py
@@ -2,18 +2,36 @@
 import sys
 import pathlib
 
+def wrong_answer(message: str):
+    judgemessage = pathlib.Path(sys.argv[3]) / 'judgemessage.txt'
+    judgemessage.write_text(message)
+    sys.exit(43)  # WA
+
+
 test_in = int(pathlib.Path(sys.argv[1]).read_text())
-team_ans = int(input())
+try:
+    team_ans_string = input()
+    try:
+        team_ans = int(team_ans_string)
+    except ValueError:
+        wrong_answer(f"team output '{team_ans_string}' is not an integer")
+except EOFError:
+    wrong_answer('no input from team')
+
 if test_in < 0:
-	# first pass
-	if team_ans < 0:
-		sys.exit(43)  # WA
-	nextpass = pathlib.Path(sys.argv[3]) / 'nextpass.in'
-	nextpass.write_text(str(team_ans))
-	sys.exit(42)  # AC + nextpass.in => next run 
+    # first pass
+    if team_ans < 0:
+        wrong_answer(f'1st pass: team output ({team_ans}) is negative')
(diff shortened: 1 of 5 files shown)


meisterT:
cc @mzuenni

test/problems/multipass/output_validators/multipass_validator/multipass_validator.py
```diff
@@ -2,18 +2,36 @@
...
+judgemessage = pathlib.Path(sys.argv[3]) / 'judgemessage.txt'
 test_in = int(pathlib.Path(sys.argv[1]).read_text())
-team_ans = int(input())
+try:
+    team_ans_string = input()
+    try:
+        team_ans = int(team_ans_string)
+    except ValueError:
+        judgemessage.write_text(f"team output '{team_ans_string}' is not an integer")
+        sys.exit(43)  # WA
```
RagnarGrootKoerkamp, line 13: could you wrap the `judgemessage.write_text` and `exit(43)` together in a function?

test/problems/multipass/output_validators/multipass_validator/multipass_validator.py
```diff
@@ -2,18 +2,36 @@
 import sys
 import pathlib
 
+judgemessage = pathlib.Path(sys.argv[3]) / 'judgemessage.txt'
```
RagnarGrootKoerkamp, line 5: 'we' kinda always just use `stderr` instead of `judgemessage`, but yes I suppose this is better for a proper example

RagnarGrootKoerkamp (changes requested):

test/problems/multipass/output_validators/multipass_validator/multipass_validator.py
mzuenni: I would say ideally both should have an example? the system should be able to concat `judgemessage.txt` as well as `stderr` output.

test/problems/multipass/output_validators/multipass_validator/multipass_validator.py
meisterT: Done

test/problems/multipass/output_validators/multipass_validator/multipass_validator.py
meisterT: I did not use `stderr` because I am not sure how it will be displayed in other systems like Kattis. I am fine with adding one message on stderr, if you think it makes sense though

RagnarGrootKoerkamp: For normal problems both work.
Probably it's indeed good to check that both work for multipass as well.
Maybe print to both a message like `this is printed both to stderr and judgerror` or so.

PR #394, opened 2024-09-30
only add validator to cache if validation was successful

mzuenni:


RagnarGrootKoerkamp (approved):
What exactly is the behaviour change here:
Do I understand correctly that before, a failed validation run would be added to the cache, and then on a rerun validation would pass because the failing cases are skipped?

We could actually cache failed cases and require changes to the validator to re-check them, but simply not caching them is also fine.

mzuenni:
Yeno... if the validator failed the code returned `False`. However, if you use `--no-validators` it would not really fail and would be added to the cache. If you run the generator command again without `--no-validators` it would detect the cached entry and would skip validation. So we would keep invalid testcases even though `--no-validators` is no longer provided.

PR #395, opened 2024-10-01
removed deprecated typing

mzuenni:
Seems like python 3.9 deprecated some of those typing stuff https://docs.python.org/3/library/typing.html

### DIFF
--- a/bin/fuzz.py
+++ b/bin/fuzz.py
@@ -6,7 +6,7 @@ import generate
 import time
 import threading
 from colorama import Fore, Style
-from typing import Callable
+from collections.abc import Callable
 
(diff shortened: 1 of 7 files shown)


RagnarGrootKoerkamp (approved):
Looks like `collections.abc` was added in `3.3` so this should be fine :)

https://docs.python.org/3/library/collections.abc.html#module-collections.abc

mzuenni:
We say that we require 3.10 anyway?

mpsijm (approved):
That's correct, so should be fine :slightly_smiling_face: Thanks for the clean-up! :smile:

PR #399, opened 2024-10-25
faster sanity checks

### LINKED ISSUE #398: `sanity_check` is slow
~We're not measuring time/memory usage here, so we should parallellize to speed up `bt zip`.~

Old but wrong title: `bt constraints` doesn't do parallel while it should.

mzuenni:
fix #398 ?


RagnarGrootKoerkamp (approved):
Awesome! `bt constraints` went from ~200s to ~10s.

PR #400, opened 2024-10-27
bt download_submissions

RagnarGrootKoerkamp:
@mpsijm

### DIFF
--- a/bin/contest.py
+++ b/bin/contest.py
@@ -80,9 +80,7 @@ def get_contest_id():
 def get_contests():
     url = f'{get_api()}/contests'
     verbose(f'query {url}')
-    r = call_api('GET', '/contests')
-    r.raise_for_status()
-    contests = json.loads(r.text)
+    contests = call_api_get_json('/contests')
     assert isinstance(contests, list)
     return contests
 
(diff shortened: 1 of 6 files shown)


mpsijm (approved):

RagnarGrootKoerkamp:
Thanks for fixing and merging :heart_eyes:

RagnarGrootKoerkamp:
While we're at it, we should probably also download the scoreboard, just so we have it.

Also, cool how small the diff is to make the parallel download :)

mpsijm:
Adding the scoreboard to the download sounds good! I already shut down, but feel free to add it 🙂

mpsijm:
See https://github.com/mpsijm/BAPCtools/commit/cc15a1200706d8af2f4811d20501848b503d2574, but it doesn't completely work yet. Besides the scoreboard's JSON data (i.e. the stuff we download in `bin/misc/download_scoreboard.sh`), I wanted to also download the static scoreboard zip and clarifications html, but that doesn't work yet, so I've opened https://github.com/DOMjudge/domjudge/issues/2768 for that :stuck_out_tongue:

PR #402, opened 2024-11-08
Reorder

mzuenni:
add `bt generate --reorder` to reorder testcases by difficulty.

Some notes here:
-  I put this as a sub command of `bt generate` fore many reasons
   - This only works with `generator.yaml` and auto numbered cases
   - it heavily depends on the `GeneratorConfig` object
- the reordering needs to run all non AC submissions which is new for generate...
- the algorithm for reordering takes `|s|*|t|^2` time

### DIFF
--- a/bin/generate.py
+++ b/bin/generate.py
@@ -9,6 +9,7 @@ import shutil
 import secrets
 
 from collections.abc import Callable
+from colorama import Fore, Style
 from pathlib import Path, PurePosixPath, PurePath
 from typing import Literal, overload
 
@@ -70,6 +71,10 @@ def is_testcase(yaml):
 
 def is_directory(yaml):
     return isinstance(yaml, dict) and not is_testcase(yaml)
+
+
+def has_count(yaml):
+    return isinstance(yaml, dict) and 'count' in yaml and isinstance(yaml['count'], int)
 
 
 # Returns the given path relative to the problem root.
@@ -654,7 +659,7 @@ class TestcaseRule(Rule):
             elif target.is_file():
                 # Target exists but source wasn't generated -> remove it
                 generator_config.remove(target)
-                bar.log(f'REMOVED: {target.name}')
+                bar.log(f'REreordered: {target.name}')
             else:
                 # both source and target do not exist
                 pass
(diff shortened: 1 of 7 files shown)


bin/run.py
```diff
@@ -328,9 +328,9 @@ def run(self, in_path, out_path, crop=True, args=[], cwd=None, default_timeout=F
     # Run this submission on all testcases for the current problem.
```
RagnarGrootKoerkamp, line 328: update the comment to match that `testcases` is passed in now, or maybe rename to just `run_testcases`, since it's only called from `run_some`.

RagnarGrootKoerkamp (commented):
Pretty nice!

Some remaining comments:
1. update docs
2. I think it doesn't work currently for `generate --reorder data/secret/1-small` subdirs?

bin/generate.py
```diff
@@ -2002,11 +2003,155 @@ def get_or_add(yaml, key, t=ruamel.yaml.comments.CommentedMap):
...
+        submissions = [s for s in ts_pair[1] if s.expected_verdicts != [Verdict.ACCEPTED]]
+
+        if not testcases:
+            error(f'No testcases found.')
+            return False
+        if not submissions:
+            error(f'No rejected submissions found.')
+            return False
+
+        ok, verdict_table = Problem.run_some(testcases, submissions)
```
RagnarGrootKoerkamp, line 2057: This `ok` can be `False` when running WA submissions on only a subset (subdir) of cases. Probably we should just ignore it.

bin/generate.py
```diff
@@ -658,7 +658,7 @@ def link(t, problem, generator_config, bar, dst):
             elif target.is_file():
                 # Target exists but source wasn't generated -> remove it
                 generator_config.remove(target)
-                bar.log(f'REreordered: {target.name}')
+                bar.log(f'REMOVE: {target.name}')
```
RagnarGrootKoerkamp, line 661: `REMOVED` is nicer i think

RagnarGrootKoerkamp (approved):

RagnarGrootKoerkamp:
lgtm; feel free to merge

PR #404, opened 2024-11-10
[skel] Add bt forgejo_actions and github_actions

RagnarGrootKoerkamp:
Add `bt forgejo_actions` that can be run from each contest in a git repo.
The contest can either be the root, or `/<contest>`.

The docs contain some additional configuration for the runners. Everything else
should just work.


RagnarGrootKoerkamp:
Currently we always pull the latest version of BAPCtools before running, but this requires `/opt/hostedtoolcache` to be mounted as a shared cache to be efficient (avoid constantly re-pulling).
We could also just use the version in the `ragnargrootkoerkamp/bapctools` docker container, which may be not the latest, but for those not working on bapctools that doesn't matter anyway probably.

RagnarGrootKoerkamp:
Added a `--latest-bt` flag so the user can explicitly ask for a recent image to be cached. Otherwise the BAPCtools from the docker is used, which is a bit simpler and doesn't require a cache volume to be mounted.

Only remaining TODOs for later:
- Forgejo notifactions for when CI fails.
- Triggering on pull requests as well.
- Optionally, clean up the dependency on the checkout action by doing a manual clone, like here:
  https://github.com/RagnarGrootKoerkamp/research/blob/master/.github/workflows/push.yml#L17

mpsijm:
I've run the setup for 2025. Rebased the branch, resolved the conflicts, and made some changes to make it work better. I think this is ready to merge if you agree with the changes :smile:

Regarding the TODOs:
- Yeeaahh... https://codeberg.org/forgejo/forgejo/issues/3719 is still open :see_no_evil: That will probably not be part of the action configuration YAML files, though, so it doesn't affect this PR.
- Removed the branch restriction entirely. In the past few years, we only push to `master`, and sometimes, somebody creates a branch to test something out if they are afraid that it'll break the CI on `master`. But then they _do_ want to know whether the CI would actually pass :stuck_out_tongue: 
- I'm fine with using `actions/checkout@v4` :shrug: At least this uses the [ForgeJo variant](https://code.forgejo.org/actions/checkout), right? (rather than the [GitHub variant](https://github.com/actions/checkout))

mpsijm (approved):

RagnarGrootKoerkamp:
1. jup I'm subscribed to the issue
2. fair enough
3. ok, fine with me

4. oh lol I didn't remember adding those before. I just re-added them locally to the master branch because CI was broken. Must have build a new version at some point without the `texlive-lang-{european,german}` packages.

PR #408, opened 2024-11-12
More stats

mzuenni:
more stats (mostly useful as fun stats for solution slides)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,6 +21,8 @@ repos:
         -   ruamel.yaml==0.18.6
         -   questionary==2.0.1
         -   types-colorama==0.4.15.20240311
+        -   types-Pygments==2.18.0.20240506
+        -   types-python-dateutil==2.9.0.20241003
         -   types-PyYAML==6.0.12.20240311
         -   types-requests==2.32.0.20240712
         args:
--- a/bin/config.py
+++ b/bin/config.py
@@ -96,7 +96,7 @@ grep '^  [^ ]' | sed 's/^  //' | cut -d ' ' -f 1 | sed -E 's/,//;s/^-?-?//;s/-/_
 grep -Ev '^(h|jobs|time|verbose)$' | sed "s/^/'/;s/$/',/" | tr '\n' ' ' | sed 's/^/args_list = [/;s/, $/]\n/'
 """
 # fmt: off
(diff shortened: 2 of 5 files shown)


RagnarGrootKoerkamp:
Can you improve the installation instructions for `pygount`? I don't have it and there's no pacman/AUR package.. Did you install it in your user environment with pip or so?

mzuenni:
yes installed it with pip... we could also use non python code to do the counting...

mpsijm:
Or do the counting manually in Python? Iterating over files in directories and discarding comments in 5 languages shouldn't be too much code :slightly_smiling_face:

mzuenni:
I would say counting this ourselves is not an option... and just for the fun, how many c++ lines of code are this:
```
//this is a comment\
is this a comment?
int main() {}
```

mzuenni:
any idea for more stats that we want to calculate?

mzuenni:
@RagnarGrootKoerkamp the code now only requires pygments and not pygount, is that easier to install?

mpsijm:
Fixed the rebase, it looked like all commits from master ended up in this branch without the base actually changing, so the commits were duplicated :stuck_out_tongue: 

I can confirm that it works on my machine, I already had `python-pygments` installed via Pacman.

I'm not sure if `--slides` is the best name for the flag, because it doesn't actually do anything with slides (the fact that we may use these stats for the slides, is something else). If we want to name the flag for the thing that it does (namely analyzing the submissions, test data, and Git repo), some suggestions could be `--more-analysis`, `--analyse-data`, or `--analyse-repo`. Maybe we could (ab)use `--verbose` to trigger these extra stats, but that also feels slightly odd.

mzuenni:
> Fixed the rebase, it looked like all commits from master ended up in this branch without the base actually changing, so the commits were duplicated 😛 But not anymore 😄

whoops and thanks ^^

`I'm not sure if --slides is the best name for the flag`

hm simply `--more` or `--all` instead?

mpsijm:
Hmm, `--all` also has a different meaning for other commands already 🤔 But I'm fine with `--more` 🙂

RagnarGrootKoerkamp:
I haven't tested the latest version, but feel free to merge.

.pre-commit-config.yaml
```diff
@@ -21,6 +21,8 @@ repos:
         -   ruamel.yaml==0.18.6
         -   questionary==2.0.1
         -   types-colorama==0.4.15.20240311
+        -   types-Pygments==2.18.0.20240506
+        -   types-python-dateutil==2.9.0.20241003
```
mpsijm, line 25: Why not `datetime.strptime` (https://stackoverflow.com/a/2265383) to avoid an additional dependency?

Also, this/these additional (optional) dependenc(y/ies) should be mentioned in other places, I think.

mpsijm:
Didn't check the code inside `more_stats`, but that should be fine since it's behind the `--more` flag and it works :stuck_out_tongue: For the code around it, I only have this above comment about dependencies :slightly_smiling_face:

Issue #405, opened 2024-11-10
Move generators JSON-schema from schemastore to this repo


thorehusfeldt:
Currently, there exists at https://github.com/SchemaStore/schemastore a JSON schema that validates `generators.yaml` automatically from editors that support this.

* The JSON schema: https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/problem_package_generators.json
* Positive tests: https://github.com/SchemaStore/schemastore/tree/master/src/test/problem_package_generators
* Negative tests: https://github.com/SchemaStore/schemastore/tree/master/src/negative_test/problem_package_generators

This schema should instead be hosted directly in the BAPCtools repo, possibly at `support/schemas`.

To this end,

1. Follow the steps under  https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md#how-to-move-a-json-schema-from-schemastore-to-somewhere-thats-self-hosted
2. Move the `.json` file to `support/schemas` and the content of the two test directories to `test/`, in fact maybe to `test/generator_yaml`.

Questions: 

Q1: should we finally introduce some sort of versioning, so the CUE, YAML, JSON and prose specifications of `generators.yaml` can refer to each other? The schemastore expects our schemas to have version numbers; see, e.g., the format for referring to our locally hosted version: https://github.com/SchemaStore/schemastore/pull/2421/files . Thus, some kind of decision needs to be made. I unanimously decided that the schema currently at schemastore represents version 0.9 of the generators framework. Possible names for the post-migration version could be 0.9.1 or 1.0.

Q2: schemastore validates the JSON schema against the test using a javascript validator. We have no such infrastructure. Do we want to automate this?

thorehusfeldt:
More TODO: the page https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/master/doc/generators.md contains two dead links (“here”) to schemata.

RagnarGrootKoerkamp:
Q1. I think I'm OK with making what we have version 1. I don't think we have any planned open feature requests for `generators.yaml` syntax at the moment.

Q2. If it's easy, sure we can add some test infra for this. But I'd also be OK to just run this manually whenever the schema changes, as long as we have some clear instructions on how to run it manually. (But by that point, it's probably easy to copy that into CI as well.)

mpsijm:
[...]

PR #410, opened 2024-11-16
`test` change directory structure for invalid yaml tests

thorehusfeldt:
Work in progress on unified verification of various schemas for generators


thorehusfeldt:
Addresses #405

thorehusfeldt:
`test/identity` has 

https://github.com/RagnarGrootKoerkamp/BAPCtools/blame/a486075a1adda0593ff459e72bfe57564a30ee03/test/problems/identity/generators/generators.yaml#L202

which is (I believe) disallowed. Authored by @mzuenni and committed by @RagnarGrootKoerkamp , please comment. (`data` can have 2-5 keys with fixed names, unknown keys are only allowed at the top level.)

RagnarGrootKoerkamp:
I don't think we depend on any unknown keys currently, not nested and not in the root.

Reason to allow them:
- backward/forward compatibility??

Reason to forbid them:
- Better linting, since you get more warnings for mistyped keys.

I'd say disallowing nested unknown keys is good for sure, and can also be convinced to drop root unknown keys.
So yes the `identity/generators/generators.yaml` file should be updated.

thorehusfeldt:
Note to self: What do we do with this one?
```yaml
visualizer: /visualizers/asy.py
data:
  invalid_inputs:
    visualizer: /foo
    data:
      foo: 
        in: "1 2 3"
  sample:
    data:
      foo: 
        in: "a b c"
  secret:
    data:
      'bar':
        in: "x y z"
```
This is nonsense (invalid inputs can’t be visualised, so it makes no sense to specify a visualiser for them). No schema forbids this.

I think we can ignore this. (I have it lying around as a yaml file to possibly validate against.)

thorehusfeldt:
The shiny new script at `test/generators_yaml` now gives:
```
Processing ../../doc/generators.yaml... OK
Processing ../problems/divsort/generators/generators.yaml... OK
Processing ../problems/identity/generators/generators.yaml... OK
Processing ../problems/generatorincludes/generators/generators.yaml... OK
Processing ../problems/multipass/generators/generators.yaml... Error
Output from cue vet:
data.secret: field is required but not present:
    ../../support/schemas/generators.cue:56:3
Processing ../problems/pickanumber/generators/generators.yaml... OK
Processing ../problems/interactivemultipass/generators/generators.yaml... Error
Output from cue vet:
data.secret: field is required but not present:
    ../../support/schemas/generators.cue:56:3
Processing invalid_yaml/bad-argument.generators.yaml... OK (correctly rejected)
Processing invalid_yaml/bad.generators.yaml... OK (correctly rejected)
Processing invalid_yaml/bad_casename.generators.yaml... OK (correctly rejected)
Processing invalid_yaml/bad_generators.yaml... OK (correctly rejected)
Processing invalid_yaml/missing_sample.generators.yaml... OK (correctly rejected)
Processing invalid_yaml/missing_secret.generators.yaml... OK (correctly rejected)
One or more cue vet commands failed.
```

The offending problems are the multipass problems in test. Should I just given them a pathological `secret` key?

thorehusfeldt:
OK, Step 1 done.

What’s new is that `test/generator_yaml/test_schemata.sh` verifies all valid-looking `*generators.yaml` in the entire rep against the `cue` schema in `support/schemas`. Also, it checks various `invalid_yaml` and makes sure that these files fail.

This is already useful and replaces with pride the CI line
```
- run: cue vet **/generators.yaml support/schemas/*cue -d '#Generators'
```
(which should now just run the bash script instead).

Next step is to also validate the same files against the JSON schema.

thorehusfeldt:
As for 

> I don't think we depend on any unknown keys currently, not nested and not in the root.

a useful key I’d like to advocate for is `version:`, defined as
```cue
  version: =~ "[0-9]{4}-[0-9]{2}" | *"2024-11"
```

test/generator_yaml/test_schemata.sh
```diff
@@ -0,0 +1,39 @@
...
+
+any_failed=0
+
+# Iterate over each .yaml file in the directory
+for dir in "${all_valid_yaml[@]}"; do
+    for file in $(find "$dir" -type f -name '*generators.yaml'); do
+	    echo -n "Processing $file... "
+	    output=$(cue vet "$file" $schemadir/*.cue -d "#Generators" 2>&1)
+	    exit_code=$?
+	    if [ $exit_code -eq 0 ]; then
```
RagnarGrootKoerkamp, line 12: Both here and below: you may want to check for 0, 1, and anything else, to catch cue crashes? (But that should be rare so maybe not worth the effort.)

RagnarGrootKoerkamp:
- Setting visualizer for invalid cases: yes let's just allow it even though it's useless.
- The empty `secret: { data: [] }` keys: lgtm
- Yes feel free to update the CI workflow.
- Yes, adding a version makes sense. Feel free to add that to the skel files in the repo as well.

thorehusfeldt:
I found an inconsistency. The JSON schema has (after converting back to CUE for readability):
```
input_validator_flags?: string
```
whereas the CUE scheme has
```
input_validator_flags?:  *"" | string | {[string]: string}
```
This makes a difference, for instance for `test/problems/generatorincludes/generators.yaml`, which has
```yaml
```
  secret:
    data:
      - small:
          testdata.yaml:
            output_validator_flags: space_change_sensitive
            input_validator_flags:
              connected: --small
              strongly-connected: --small
```
thus following the CUE schema. In problemformatpackagespeak, both `legacy` and `draft` `2023-07` versions have

> String or map with the keys "name" and "flags"

and

> Sequence of strings or map of strings to sequences of strings

Note that the latter (`draft`) version changed its names again and the key is now called `input_validator_args` (not `_flags`). It’s not easy to keep this in sync, in particular because of a lack version numbers, stringent schemas, or reference (versioned) implementations.

I think I’ll try to upgrade the JSON schema.

**Comment by thorehusfeldt** [comment]
Three TODOs:

[ ] in `invalid_yaml/bad_generators.yaml`, make `cue vet` go through _all_ of the `---`-separated yaml snippets (and reject them) instead of bailing out at the first by default
[  ] understand why `cue import` of a JSON schema adds a `!~ "^()^$` regex to key names (escalated to CUE slack; this may be an error)
[  ] refactor `testdata.yaml` in JSON schema
[ ] update `testdata.yaml` in JSON schema to allow multiple `input_validator_flags`.

**Comment by thorehusfeldt** [comment]
In `bad_generators.yaml`, many entries are rejected for the wrong reason. For instance:
```yaml
# Solution ans visualizer must have an absolute path:
solution: a
```
if an invalid generator, but that's clear already from the missing `data.sample`.

thorehusfeldt:
Here is where I register changes. 

In /Users/thore/Developer/BAPCtools/test/problems/identity/generators/generators.yaml:80:

```yaml
 - dir: dir/ 306 
 ```
 
 I suggest to disallow the `/`. The first token in a generator call is the name of a file, the name of a directory, or the name of a generator.

File: test/generator_yaml/invalid_yaml/empty-generators.yaml [hunk 2]
```diff
@@ -0,0 +1,2 @@
+# The smallest valid generator
```
RagnarGrootKoerkamp, line 1: comment says valid, but it's in the invalid directory

RagnarGrootKoerkamp (commented):
Just some small comments, but overall things look nice :)

test/generator_yaml/valid_yaml/unusual-keys.generators.yaml
```diff
@@ -0,0 +1,5 @@
+data:
+  sample: {data: []}
+  secret: 
+    data:
+      1: b
```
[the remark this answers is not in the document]
thorehusfeldt: Neither CUE nor JSON can see the difference between
```yaml
1: foo
```
and
```yaml
"1": "foo"
```

I’m not quite sure what is best here. I want to validate as much as possible in a schema language, but for the rest I think we need a python script. Undecided so far.

PR #411, opened 2024-11-23
add rust to Dockerfile

chistopher:
To make it work in CI, someone with permissions has to update the image in DockerHub


mpsijm (approved):
@RagnarGrootKoerkamp can you update the image on DockerHub? :slightly_smiling_face:

PR #413, opened 2024-12-14
[latex] Add command problem_slides

mpsijm:
At long last, here's the final PR for `bt problem_slides`, which has been used successfully for the live streams of the past three NWERCs now :smile:

The `bt problem_slides` command compiles a slide deck with one slide per problem, each slide containing a summary of the problem that can be used to show during the live stream of a contest. For an example, see https://2024.nwerc.eu/main/problem-slides.pdf.

Note that this adds `skel/problem/problem_statement/problem-slide.en.tex`. However, the majority of contests will not have a live stream. I guess this is fine though, because the file can be safely deleted and/or be left out in a custom `skel/problem` tree.

I'd like someone to double-check before merging, perhaps try to run the command on the NWERC repo again (both on the problem and contest levels), just in case. Only the first commit was actually used during the past NWERC, I've added the last three commits only today to wrap things up :slightly_smiling_face:


bin/tools.py
```diff
@@ -1074,11 +1111,27 @@ def run_parsed_arguments(args):
...
+                        contest,
+                        problems,
+                        tmpdir,
+                        statement_language,
+                        build_type=latex.PdfType.SOLUTION,
+                        web=True,
                     )
-                    success &= latex.build_contest_pdfs(
-                        contest, problems, tmpdir, statement_language, solutions=True, web=True
+                    success &= latex.build_contest_pdf(
```
RagnarGrootKoerkamp, line 1128: Should this be more granular? It's annoying if things fail because no problem slides are present.

Or maybe better is to never make it an error if problem slides are missing? (But that's kinda tricky when you want slides for all problems?)

Also I checked an doing `bt problem_slides` at the contest level with some `problem-slide.en.tex` missing crashes in latex (because file not found). We should probably detect when there's not a single problem-slide present and then just Log that (without error). When only some files are present, it's fine to error IMO.

At the problem level it gives a nice message `problem-slide.en.tex not found`, so there all is good.

Otherwise LGTM! Thanks for making the PR :)

RagnarGrootKoerkamp (approved):

bin/tools.py
mpsijm: That's a good point! Will reduce some errors to warnings today or tomorrow :smile:

mpsijm: I've added some extra checks :slightly_smiling_face: The behaviour at the problem level hasn't changed, this is the new behaviour at the contest level:
- When no `problem-slide.*.tex` are present:
  - `bt problem_slides` will warn for all problems, followed by a failing LaTeX compilation (because the user is explicitly trying to compile them)
  - `bt zip` will log "No problem has problem-slide.*.tex, skipping problem slides"
- When at least one `problem-slide.*.tex` is present:
  - Both `bt problem_slides` and `bt zip` will warn for the problems that do not have it, but will compile the rest

RagnarGrootKoerkamp: That sounds good. Thanks! Merging now.

RagnarGrootKoerkamp (approved):

PR #417, opened 2025-01-31
test python3.10

mzuenni:


mpsijm:
The test log still says 3.12? :thinking: 

```
platform linux -- Python 3.12.7, pytest-8.3.3, pluggy-1.5.0
```

mzuenni:
hm can we just add the same thing for `test` ?

mzuenni:
hm that doesn't seem to change anything... https://github.com/RagnarGrootKoerkamp/BAPCtools/actions/runs/13071402045/job/36473679763

mpsijm:
Oh, that makes sense actually, because `pre-commit` runs using `ubuntu-latest` (where the `setup-python` action can be used), but `test` runs on the BAPCtools image, which is based on Arch Linux, so that simply always runs the latest version of Python (which was 3.12 at the time when it was last built) :joy: So then we could either force-install 3.10 from the AUR (which I guess is not easy to do unless we download and extract some `.deb` manually from somewhere), or we just leave this be :stuck_out_tongue: At least the type-checker in the pre-commit job should check that we don't use any language features of 3.11 or newer :slightly_smiling_face:

mzuenni:
> At least the type-checker in the pre-commit job should check that we don't use any language features of 3.11 or newer 🙂

hopefully it does that... i didn't actually verify this 🙃

Issue #352, opened 2024-03-05
Separate universally and problem-specifically invalid inputs


thorehusfeldt:
Some input files are invalid _universally_ (for all problems, certainly all problem in a contest), others for _this problem_. I suggest to separate these issues.

For instance, a default `contest.yaml` could contain 

```yaml
invalid_inputs:
  latin-1: Naïve coöperation
  unicode: ¯\\_(ツ)_/¯
  empty: ""
  random: YVRtr&*teTsRjs8ZC2%kN*T63V@jJq!d
  newline: "\n"
  not-printable_1: "\x7f"
  not_printable_2: "\xe2\x82\xac"
  bismillah: ﷽
```
and always generate the corresponding testcases. (It could do so explicitly, so that `data/invalid_inputs/latin-1.in` is indeed created by `bt generate`.)

An Icelandic contest would _not_ forbid `latin-1` because it wants input files like `jökulsá`.

Invalid pseudotestcases that are _particular_ to a problem instead belong to `generators.yaml`. An informative and easy-to-repurpose skeleton could look like this:
```yaml
invalid_inputs:
  data:
    prose: { in: 2 two }
    leading-zero: { in: 02 2 }
    floating-point: { in: 2.0 2 }
    too-many-tokens: { in: 2 2 2 }
    too-few-tokens: { in: "2" }
    extra-line: { in: "2 2\n\n" }
    leading-whitespace: { in: " 2 2" }
    trailing-whitespace: { in: "2 2 " }
invalid_outputs:
  data:
    orwell:
      in: 2 2 # valid input
      ans: 4  # well-formed default answer
      out: 5  # wrong output
invalid_answers:
  data:
    trailing_whitespace:
      in: 2 2
      ans: "4 "
```     
Putting the above list into the skeleton would incentivise problem authors to provide those invalid inputs.

RagnarGrootKoerkamp:
I think stress testing the input and output validators sense, and this is a reasonable way to do it.

One thought: in NWERC we always use some library (header file), and it feels like all the invalid inputs here are rather tests of the library being used. Once the library is well tested, whitespace and encoding errors should be handled by it and are not really errors specific to the problem.

On the other hand when every problem has it's own non-library-based validator, having extra tests makes a lot of sense.

[...]

PR #420, opened 2025-02-07
Add validation of generically generated invalid files

mzuenni:
Dynamically generates (likely) invalid testcases to challenge validators
related: #352

### DIFF
--- a/bin/config.py
+++ b/bin/config.py
@@ -5,20 +5,20 @@ import argparse
 import os
 import re
 from pathlib import Path
-from typing import Final, Literal, Optional
+from typing import Final, Literal, Mapping, Optional, Sequence
 
 # return values
-RTV_AC: Final = 42
-RTV_WA: Final = 43
+RTV_AC: Final[int] = 42
+RTV_WA: Final[int] = 43
 
-SUBMISSION_DIRS: Final = [
+SUBMISSION_DIRS: Final[Sequence[str]] = [
     "accepted",
     "wrong_answer",
     "time_limit_exceeded",
     "run_time_error",
 ]
 
-KNOWN_LICENSES: Final = [
+KNOWN_LICENSES: Final[Sequence[str]] = [
     "cc by-sa",
     "cc by",
     "cc0",
@@ -29,25 +29,26 @@ KNOWN_LICENSES: Final = [
 ]
 
 # When --table is set, this threshold determines the number of identical profiles needed to get flagged.
-TABLE_THRESHOLD: Final = 4
+TABLE_THRESHOLD: Final[int] = 4
 
-FILE_NAME_REGEX: Final = "[a-zA-Z0-9][a-zA-Z0-9_.-]*[a-zA-Z0-9]"
-COMPILED_FILE_NAME_REGEX: Final = re.compile(FILE_NAME_REGEX)
+FILE_NAME_REGEX: Final[str] = "[a-zA-Z0-9][a-zA-Z0-9_.-]*[a-zA-Z0-9]"
+COMPILED_FILE_NAME_REGEX: Final[re.Pattern[str]] = re.compile(FILE_NAME_REGEX)
 
-KNOWN_TESTCASE_EXTENSIONS: Final = [
+KNOWN_TESTCASE_EXTENSIONS: Final[Sequence[str]] = [
     ".in",
     ".ans",
     ".out",
 ]
 
-KNOWN_VISUALIZER_EXTENSIONS: Final = [
+KNOWN_VISUALIZER_EXTENSIONS: Final[Sequence[str]] = [
     ".png",
     ".jpg",
     ".svg",
     ".pdf",
 ]
 
-KNOWN_TEXT_DATA_EXTENSIONS: Final = KNOWN_TESTCASE_EXTENSIONS + [
+KNOWN_TEXT_DATA_EXTENSIONS: Final[Sequence[str]] = [
+    *KNOWN_TESTCASE_EXTENSIONS,
     ".interaction",
     ".hint",
     ".desc",
@@ -56,9 +57,12 @@ KNOWN_TEXT_DATA_EXTENSIONS: Final = KNOWN_TESTCASE_EXTENSIONS + [
     #'.args',
 ]
 
-KNOWN_DATA_EXTENSIONS: Final = KNOWN_TEXT_DATA_EXTENSIONS + KNOWN_VISUALIZER_EXTENSIONS
+KNOWN_DATA_EXTENSIONS: Final[Sequence[str]] = [
+    *KNOWN_TEXT_DATA_EXTENSIONS,
+    *KNOWN_VISUALIZER_EXTENSIONS,
+]
 
-INVALID_CASE_DIRECTORIES: Final = [
+INVALID_CASE_DIRECTORIES: Final[Sequence[str]] = [
     "invalid_inputs",
     "invalid_answers",
     "invalid_outputs",
@@ -66,13 +70,13 @@ INVALID_CASE_DIRECTORIES: Final = [
 ]
 
 
-SEED_DEPENDENCY_RETRIES: Final = 10
+SEED_DEPENDENCY_RETRIES: Final[int] = 10
 
 # The root directory of the BAPCtools repository.
-TOOLS_ROOT: Final = Path(__file__).resolve().parent.parent
+TOOLS_ROOT: Final[Path] = Path(__file__).resolve().parent.parent
 
 # The directory from which BAPCtools is invoked.
-current_working_directory: Final = Path.cwd().resolve()
+current_working_directory: Final[Path] = Path.cwd().resolve()
 
 # Add third_party/ to the $PATH for checktestdata.
 os.environ["PATH"] += os.pathsep + str(TOOLS_ROOT / "third_party")
@@ -81,7 +85,7 @@ os.environ["PATH"] += os.pathsep + str(TOOLS_ROOT / "third_party")
 
 args = argparse.Namespace()
 
-DEFAULT_ARGS: Final = {
+DEFAULT_ARGS: Final[Mapping] = {
     "jobs": (os.cpu_count() or 1) // 2,
     "time": 600,  # Used for `bt fuzz`
     "verbose": 0,
@@ -93,10 +97,10 @@ DEFAULT_ARGS: Final = {
 """
 for cmd in $(bapctools --help | grep '^  {' | sed 's/  {//;s/}//;s/,/ /g') ; do bapctools $cmd --help ; done |& \
 grep '^  [^ ]' | sed 's/^  //' | cut -d ' ' -f 1 | sed -E 's/,//;s/^-?-?//;s/-/_/g' | sort -u | \
-grep -Ev '^(h|jobs|time|verbose)$' | sed "s/^/'/;s/$/',/" | tr '\n' ' ' | sed 's/^/ARGS_LIST: Final = [/;s/, $/]\n/'
+grep -Ev '^(h|jobs|time|verbose)$' | sed "s/^/'/;s/$/',/" | tr '\n' ' ' | sed 's/^/ARGS_LIST: Final[Sequence[str]] = [/;s/, $/]\n/'
 """
 # fmt: off
-ARGS_LIST: Final = ['1', 'add', 'all', 'answer', 'api', 'author', 'check_deterministic', 'clean', 'colors', 'contest', 'contest_id', 'contestname', 'cp', 'default_solution', 'depth', 'directory', 'error', 'force', 'force_build', 'generic_invalid', 'input', 'interaction', 'interactive', 'invalid', 'kattis', 'language', 'memory', 'more', 'move_to', 'no_bar', 'no_generate', 'no_solution', 'no_solutions', 'no_testcase_sanity_checks', 'no_time_limit', 'no_validators', 'no_visualizer', 'open', 'order', 'order_from_ccs', 'overview', 'password', 'post_freeze', 'problem', 'problemname', 'remove', 'reorder', 'samples', 'sanitizer', 'skel', 'skip', 'sort', 'submissions', 'table', 'testcases', 'time_limit', 'timeout', 'token', 'tree', 'type', 'username', 'watch', 'web', 'write']
+ARGS_LIST: Final[Sequence[str]] = ['1', 'add', 'all', 'answer', 'api', 'author', 'check_deterministic', 'clean', 'colors', 'contest', 'contest_id', 'contestname', 'cp', 'default_solution', 'depth', 'directory', 'error', 'force', 'force_build', 'generic_invalid', 'input', 'interaction', 'interactive', 'invalid', 'kattis', 'language', 'memory', 'more', 'move_to', 'no_bar', 'no_generate', 'no_solution', 'no_solutions', 'no_testcase_sanity_checks', 'no_time_limit', 'no_validators', 'no_visualizer', 'open', 'order', 'order_from_ccs', 'overview', 'password', 'post_freeze', 'problem', 'problemname', 'remove', 'reorder', 'samples', 'sanitizer', 'skel', 'skip', 'sort', 'submissions', 'table', 'testcases', 'time_limit', 'timeout', 'token', 'tree', 'type', 'username', 'watch', 'web', 'write']
 # fmt: on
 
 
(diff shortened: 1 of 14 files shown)


mzuenni:
@mpsijm @RagnarGrootKoerkamp more input here is welcome :)

bin/problem.py
```diff
@@ -916,38 +918,104 @@ def validate_data(problem, mode: validate.Mode, constraints: dict | bool | None
                 log(f"Not running answer_validators for{msg} problems.")
             return True
 
+        action = (
+            "Invalidation"
+            if mode == validate.Mode.INVALID
+            else (
+                f"Collecting {mode} constraints" if constraints else f"{mode} validation"
+            ).capitalize()
+        )
```
mpsijm, line 927: Why `.capitalize()` the `action` when both branches return something that already starts with a capital? :thinking: 

Oh wait, the `mode` is not capitalized. Maybe this can be simplified to only `mode.capitalize()` then :slightly_smiling_face:

skel/problem/generators/generators.yaml
```diff
@@ -66,19 +66,9 @@ data:
...
-          in: YVRtr&*teTsRjs8ZC2%kN*T63V@jJq!d
-      not_printable_1:
-          in: "\x7f"
-      not_printable_2:
-          in: "\xe2\x82\xac"
-      unicode:
-          in: "¯\\_(ツ)_/¯"
+    data: []
+      #dont_start_with_whitespace:
+      #    in: "<valid testcase>"
```
mpsijm, line 71: Shouldn't there be an invalid test case here? Also, why is it called `dont_start_with_whitespace`?

bin/problem.py
```diff
@@ -916,38 +918,104 @@ def validate_data(problem, mode: validate.Mode, constraints: dict | bool | None
...
+        # find at least one (valid?) testcase to modify
+        test_paths = list(glob(p.path, "data/sample/**/*.in"))
+        test_paths += list(glob(p.path, "data/secret/**/*.in"))
+        test_paths = [path for path in test_paths if path.with_suffix(".ans").exists()]
+        test_paths.sort()
+        test_path = test_paths[0] if test_paths else None
+        if test_path:
+            test_case = test_path.relative_to(p.path / "data").with_suffix("")
+            log(f"Generating invalid testcases based on: {test_case}")
+        debug(f"writing generated invalid testcases to: {base_path}")
```
mpsijm, line 943: This debug log is visible, even without `-v`, I guess that's intended ('cause the log is "debug" and not "verbose")?

bin/validator_tests.py
```diff
@@ -0,0 +1,66 @@
+from collections.abc import Callable
+from typing import Optional, TypeVar
+
+_GENERATOR_NAMES: set[str] = set()
+GENERATORS: list[tuple[str, str | Callable[[str], Optional[str]], bool]] = []
```
mpsijm, line 5: Maybe add documentation comments about what going on here? From what I'm gathering, these generators are created once `validator_tests` is imported, and these generators are then used to generate invalid test cases in `problem.py`.

One note about creating these generators in top-level code: this does mean that this is _always_ executed, also in the many instances where a command different than `bt validate` is called, and also when performing tab-completion in your shell. Since these generators are only used inside `Problem.validate_invalid_extra_data`, perhaps we should wrap this in a function anyway :stuck_out_tongue: I think Python will still us nest the decorator magic inside that function :slightly_smiling_face:

bin/problem.py
```diff
@@ -916,38 +918,104 @@ def validate_data(problem, mode: validate.Mode, constraints: dict | bool | None
...
+                full_path.parent.mkdir(parents=True, exist_ok=True)
+
+                for ext in copy:
+                    assert test_path is not None
+                    shutil.copy(test_path.with_suffix(ext), full_path.with_suffix(ext))
+                full_path.with_suffix(write).write_text(content)
+
+                testcases.append(testcase.Testcase(p, full_path, short_path=short_path))
+
+        return p._validate_data(validate.Mode.INVALID, None, "Generic Invalidation", testcases)
```
mpsijm, line 985: Currently, `doc/validation.md` only talks about Input/Answer/Output (in)validation, I think it would be good to add a section about this new "Generic Invalidation" as well :slightly_smiling_face:

mpsijm (changes requested):
Looking good at first glance! :smile: Tried it on some problems but didn't try to break it yet. Here are some first comments based on the code :slightly_smiling_face:

skel/problem/generators/generators.yaml
mzuenni: this is the same for every other subdir in the example yaml.

mpsijm: Sure, but the `invalid_answer` is the only one that actually starts with whitespace :stuck_out_tongue: I've cleaned this section up a bit :slightly_smiling_face:

bin/validator_tests.py
mpsijm: > I think its just clearer this way and easier to extend.

It'll be just as easy to extend if it's not a global var, right? With the advantage that the creation of `GENERATORS` only happens when it's needed, and not every time any other `bt` command is used :slightly_smiling_face:

mzuenni: well every other constant is also filled when bt is run even though they don't get used?

bin/problem.py
mzuenni: I just moved this, so no idea what exactly is going on XD

bin/config.py
```diff
@@ -5,20 +5,20 @@
...
+from typing import Final, Literal, Mapping, Optional, Sequence
 
 # return values
-RTV_AC: Final = 42
-RTV_WA: Final = 43
+RTV_AC: Final[int] = 42
+RTV_WA: Final[int] = 43
 
-SUBMISSION_DIRS: Final = [
+SUBMISSION_DIRS: Final[Sequence[str]] = [
```
mpsijm, line 14: I assume that `Sequence` is used here instead of `list`, because a `list` is mutable and a `Sequence` is not? Nice :smile:

Note however that `typing.Sequence` is a deprecated alias to `collections.abc.Sequence`: https://docs.python.org/3/library/typing.html#typing.Sequence

(same goes for `collections.abc.Mapping`)

bin/generate.py
```diff
@@ -1414,7 +1414,7 @@ def parse_generators(generators_yaml):
         return generators
 
     # Only used at the root directory level.
-    ROOT_KEYS: Final = [
+    ROOT_KEYS: Final[Sequence] = [
```
mpsijm, line 1417: This could be `Final[Sequence[tuple[str, dict[Path, list[Path]], Callable[...something...]]]]`, but I also understand if you don't want to go through the trouble of writing out the entire type :stuck_out_tongue:

bin/validator_tests.py
```diff
@@ -1,20 +1,25 @@
...
+# can be used on its own or as decorator for a function
 def register(name: Optional[str] = None, only_whitespace_change: bool = False) -> Callable[[T], T]:
     def decorator(func: T) -> T:
         nonlocal name
         if not isinstance(func, str) and not name:
             assert hasattr(func, "__name__")
             name = func.__name__
         assert name
+        assert isinstance(_GENERATOR_NAMES, set)
+        assert isinstance(GENERATORS, list)
```
mpsijm, line 22: If I can be honest, I think type-juggling immutable types back to mutable types is slightly uglier than writing one big function that returns the list of generators, instead of storing them in global variables :innocent:

bin/generate.py
mzuenni: yeah... i have no idea what the exact type here even is XD

mpsijm (approved):
@mzuenni The PR is still marked as draft, is there anything else you'd like to do here? It looks merge-able to me :smile:

mzuenni:
it is merge-able, the question is if the current behavior is the one we want ^^

mpsijm:
I've tried it on some old problems, and it's looking good to me! On problems that only have a single floating-point number as output (e.g. NWERC 2024 Kruidnoten), I receive the following output:

```
LOG: Generating invalid testcases based on: sample/01
Generic Invalidation: invalid_answers/leading_zero          was not rejected by answer validation
Generic Invalidation: invalid_outputs/leading_zero          was not rejected by invalid files validation
```

But I'd say outputting `042.0` is just as valid as `42.0`. I tried changing an accepted submission to print leading zeroes, and `bt run` accepts that. Our tests for the default output validator agree with that: https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/master/test/default_output_validator/default_output_validator.yaml#L825-L826. So I guess the leading-zero check should be disabled for problems with floating-point output? :thinking: 

For problems with integer output (e.g. NWERC 2024 Connect Five), leading zeroes are rejected, as expected, because we do an exact match there.

The exploration of the floating-point numbers made me think: what if the first sample is `impossible`? Should we generate invalid test cases based on all samples, so that we are generating invalid cases for all possible variants of the output? :thinking:

mzuenni:
> But I'd say outputting `042.0` is just as valid as `42.0`
I would say as output that is ok, as jury answer it is not.

Right now they are generated on the first testcase that we find... could also be random, or a few more

mpsijm:
> > But I'd say outputting `042.0` is just as valid as `42.0`
> 
> I would say as output that is ok, as jury answer it is not.

Sure, but we're not considering jury output in the "generic invalidation" step :stuck_out_tongue: The generic invalidation generates potential outputs (and I think we should indeed treat them as _outputs_, not (jury) _answers_) that we think should always be rejected by an output validator. However, I think the generically generated test cases are currently tested on both the answer validator and the output validator?

> Right now they are generated on the first testcase that we find... could also be random, or a few more

I think random is too non-deterministic, which is annoying when we run this in CI. Generating generic invalid cases for all test cases in `generators.yaml` is probably too much. I think doing this for all samples should be enough, because (1) `data/sample/` is guaranteed to exist and (2) a nice problem setter will always show all cases of output in the samples? :slightly_smiling_face:

mzuenni:
> Sure, but we're not considering jury output in the "generic invalidation" step 😛 The generic invalidation generates potential outputs (and I think we should indeed treat them as outputs, not (jury) answers)

Right now we consider inputs, outputs and jury answers, i.e., we create `invalid_inputs`, `invalid_outputs`, and `invalid_answers`, and i would say a leading zero is not ok as input or as (jury) answer.

>think random is too non-deterministic, which is annoying when we run this in CI. Generating generic invalid cases for all test cases in generators.yaml is probably too much. I think doing this for all samples should be enough, because (1) data/sample/ is guaranteed to exist and (2) a nice problem setter will always show all cases of output in the samples? 🙂

mostly agree, my only concern is that there are too many samples -> too many invalid tests ^^

bin/validator_tests.py
```diff
@@ -1,24 +1,45 @@
+import validate
+
 from collections.abc import Callable, Sequence
 from typing import Final, Optional, TypeVar
+from validate import AnswerValidator, InputValidator, OutputValidator
```
mpsijm, line 5: Can the `import validate` and `from validate import ...` be merged? :innocent:

bin/validator_tests.py
```diff
@@ -34,11 +55,11 @@ def decorator(func: T) -> T:
     register("bismillah")("﷽")
 
     # simple generators
-    register("leading_zero")(lambda x: f"0{x}")
-    register("leading_space", True)(lambda x: f" {x}")
+    register("leading_zero", IN_ANS_VALIDATORS)(lambda x: f"0{x}")
```
mpsijm, line 58: > Right now we consider inputs, outputs and jury answers, i.e., we create invalid_inputs, invalid_outputs, and invalid_answers, and i would say a leading zero is not ok as input or as (jury) answer.

I'm still getting this warning on Kruidnoten, while none of the jury solutions print leading zeroes (`bt validate` does not even run any jury submissions, I think):
```
$ bt validate
PROBLEM kruidnoten
Invalidation: Done
LOG: Generated invalid testcases based on: sample/01
LOG: Generated invalid testcases based on: sample/02
Generic Invalidation: invalid_answers/0/leading_zero          was not rejected by answer validation
```
The answer validator accepts any valid string representation of a floating-point number, which includes strings with leading zeroes. Does that mean that the validation header should do stricter floating-point parsing? :thinking:

Maybe we should want to maintain two separate definitions of floating-point numbers: as long as the string representation is valid, it should be accepted as a valid answer (even though it may "look ugly" to the human eye to have leading zeroes), because any tooling that reads the floating-point number from the answer file will also simply ignore the leading zeroes :slightly_smiling_face: 

But then we'd have another tricky challenge... how are we going to distinguish problems with floating-point output from those without? Because we _do_ want to reject leading zeroes for problems with integer output :thinking: Simply checking `validator_flags` from `problem.yaml` doesn't cut it, because that's removed in `2023-07-draft` in favour of `output_validator_args` in `testdata.yaml`, but there can be `testdata.yaml` files spread all over the `data/` tree with different `output_validator_args` :sweat_smile:

bin/problem.py
```diff
@@ -981,56 +981,69 @@ def validate_data(problem, mode: validate.Mode, constraints: dict | bool | None
...
-        validators: list[tuple[type[validate.AnyValidator], str, str, str, list[str], bool]] = [
-            (validate.InputValidator, "invalid_inputs", ".in", ".in", [], True),
-            (validate.AnswerValidator, "invalid_answers", ".ans", ".ans", [".in"], True),
-            (validate.OutputValidator, "invalid_outputs", ".ans", ".out", [".in", ".ans"], False),
+        # pick at most first 3 samples (assuming they are valid and have .ans)
+        samples: list[Path | Literal[None]] = list(glob(p.path, "data/sample/**/*.in"))
+        samples.sort()
+        samples = [path for path, _ in zip(samples, range(3))]
+        if not samples:
+            samples = [None]
```
mpsijm, line 989: This can be simplified:
- The `sorted()` function automatically turns any iterable into a `list`
- To select the first three items, `zip` with `range(3)` can be replaced with `[:3]`
- I don't see a need to put `None` in the list, when any `sample is None` is ignored down in the loop below, so the list `testcases` will stay empty anyway :stuck_out_tongue: 

```suggestion
        samples: list[Path] = sorted(glob(p.path, "data/sample/**/*.in"))[:3]
```

bin/problem.py
```diff
@@ -981,56 +981,69 @@ def validate_data(problem, mode: validate.Mode, constraints: dict | bool | None
...
-
-                testcases.append(testcase.Testcase(p, full_path, short_path=short_path))
+                    if isinstance(data, str):
+                        if i > 0 and not copy:
+                            continue
+                        content = data
+                    elif sample is None:
+                        continue
+                    elif not sample.with_suffix(read).exists():
+                        continue
```
mpsijm, line 1020: These two checks are based on the `sample`, not on `validator_tests.GENERATORS`, can't they be moved to an outer loop?

(note that `if sample is None` can probably be removed based on my previous comment, as well as the `assert sample is not None` on line 1034)

mpsijm (changes requested):
> mostly agree, my only concern is that there are too many samples -> too many invalid tests ^^

Fair enough! I see you capped it to three samples, that sounds like a sane default for now :slightly_smiling_face:

bin/problem.py
mzuenni: oh i thought `[:x]` could throw an error ^^
the None is important, and also the reason why you next suggestion does not work ^^

bin/validator_tests.py
mzuenni: what exactly do you mean?

bin/validator_tests.py
mzuenni: >The answer validator of Kruidnoten accepts any valid string representation of a floating-point number, which includes strings with leading zeroes. Does that mean that the validation header should do stricter floating-point parsing (but only for jury-answers, of course)? 🤔

IMO yes. The Answer validator only needs to accept one representation for each float (My header has a `strict_real()` and a `real()` method for this). For example the float `0` could be written as `.0`, `-0.0`, `+0.0`, `0e10` and a lot more, but the answer validator should reject everything looking weird.

>Maybe we should not want to maintain two separate definitions of floating-point numbers: as long as the string representation is valid, it should be accepted as a valid answer (even though it may "look ugly" to the human eye to have leading zeroes), because any tooling that reads the floating-point number from the answer file will also simply ignore the leading zeroes 🙂

My problem here is not tooling related. `.ans` files are those shown in the pdf and I don't want the pdf to contain `042.0`.

mpsijm: Fair point, I'm convinced that we _do_ want to reject leading zeroes for `.ans` files :smile: Of course, the input/answer validators _only_ run on the `*.in` and `*.ans` _files_ (respectively), and never on any submission output, and we indeed want the input/answer files to "look nice" (not only in the PDF, but also as guarantee to the solvers). That leaves the shape of "strict" floating-point string representations for a separate discussions, since this should be dealt with in the implementation of the input/answer validator and is not related to how this PR generates generic invalid test cases :slightly_smiling_face:

bin/problem.py
mpsijm: Can you explain what the `None` does here? Perhaps also good to write that in comments because it feels a bit magical (at least to me, but perhaps also to future readers of the code) :stuck_out_tongue: Also, is it correct that `None` is now always added to the list, also when it has elements already?

mzuenni:
@mpsijm i added some comments/documentation

mpsijm (approved):
Thanks! I'm happy with the behaviour of the generic invalidation :slightly_smiling_face: If you're also satisfied, we can unmark as draft and hit "Squash and Merge" :smile:

PR #421, opened 2025-02-09
Implement 'type' field in problem.yaml from 2023-07-draft spec

mpsijm:
As promised in https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/419#discussion_r1947193778, here's an implementation of the [`type`](https://icpc.io/problem-package-format/spec/2023-07-draft.html#type) field in `problems.yaml` for `2023-07-spec` :smile: 

There's still two `TODO`s I want to tackle before merging, but at the very least I can test this setup when working on the Freshmen Programming Contests now :slightly_smiling_face:

### DIFF
--- a/bin/problem.py
+++ b/bin/problem.py
@@ -159,7 +159,7 @@ class ProblemSettings:
             or (
                 "custom" in mode
                 if self.is_legacy()
-                # TODO: output_validator should be singular, but DOMjudge does not support this yet, so this should be fixed during export.
+                # TODO #424: output_validator should be singular, but DOMjudge does not support this yet, so this should be fixed during export.
                 else (problem.path / "output_validators").exists()
             )
         )
@@ -715,6 +715,7 @@ class Problem:
             return problem._validators_cache[key]
 
         assert hasattr(cls, "source_dirs")
+        # TODO #424: We should not support multiple output validators inside output_validator/.
         paths = [p for source_dir in cls.source_dirs for p in glob(problem.path / source_dir, "*")]
 
         # Handle default output validation
--- a/bin/problem.py
+++ b/bin/problem.py
@@ -309,15 +309,15 @@ class Problem:
         # Aliasing fields makes life easier for us 😛
         self.limits: ProblemLimits = self.settings.limits
         self.interactive: bool = self.settings.interactive
-        self.multipass: bool = self.settings.multi_pass
+        self.multi_pass: bool = self.settings.multi_pass
         self.custom_output: bool = self.settings.custom_output
 
         # Handle dependencies...
         has_validation_passes = self.limits.validation_passes is not None
-        if self.multipass and not has_validation_passes:
+        if self.multi_pass and not has_validation_passes:
             self.limits.validation_passes = 2
-        if not self.multipass and has_validation_passes:
-            warn("limit: validation_passes is only used for multipass problems. SKIPPED.")
+        if not self.multi_pass and has_validation_passes:
+            warn("limit: validation_passes is only used for multi_pass problems. SKIPPED.")
 
     def _parse_testdata_yaml(p, path, bar):
         assert path.is_relative_to(p.path / "data")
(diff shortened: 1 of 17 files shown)


bin/problem.py
```diff
@@ -675,10 +698,12 @@ def _validators(
...
 
         # Handle default output validation
-        if cls == validate.OutputValidator and problem.settings.validation == "default":
-            if paths:
+        if cls == validate.OutputValidator:
+            if problem.settings.is_legacy() and not problem.custom_output and paths:
                 error("Validation is default but custom output validator exists (ignoring it)")
-            paths = [config.tools_root / "support" / "default_output_validator.cpp"]
+                paths = []
+            if not paths:
```
RagnarGrootKoerkamp, line 705: it's checked elsewhere that `custom_output` requires at least one to be present and non-empty `path` here, right?

mpsijm: Yep, that should happen in `_read_settings`. Reminder to self to add some `assert`ion for that here 🙂 

(Also note that the new spec does not have an explicit type for custom output, this is determined by the existance of the output validator)

bin/problem.py
```diff
@@ -120,7 +142,28 @@ def __init__(self, yamldata: dict[str, Any], legacy_time_limit: Optional[float]
...
+        self.interactive: bool = "interactive" in mode
+        self.multi_pass: bool = "multi-pass" in mode
+        self.custom_output: bool = (
+            self.interactive
+            or self.multi_pass
+            or (
+                "custom" in mode
+                if self.is_legacy()
+                # TODO: output_validator should be singular, but DOMjudge does not support this yet, so this should be fixed during export.
+                else (problem.path / "output_validators").exists()
```
mpsijm, line 163: I don't think we want to support both plural and singular (`output_validators/` and `output_validator/`) right now? DOMjudge definitely doesn't support the latter (https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Service/ImportProblemService.php#L932), and I doubt that Kattis does (at least Problemtools doesn't yet: https://github.com/Kattis/problemtools/blob/develop/problemtools/verifyproblem.py#L1465-L1466) :stuck_out_tongue:

bin/skel.py
```diff
@@ -153,20 +153,29 @@ def new_problem():
...
-            log("Using default float tolerance of 1e-6")
-        if validation == "custom":
-            if _ask_variable_bool("interactive", False):
-                validation += " interactive"
-            if _ask_variable_bool("multi-pass", False):
-                validation += " multi-pass"
+        problem_type = _ask_variable_choice("type", ["default", "float", "custom"])
+    # TODO Maybe remove command-line option "default" and replace it with "pass-fail"?
+    if problem_type == "default":
+        problem_type = "pass-fail"
```
mpsijm, line 163: I don't expect that people have `bt new_problem --validation default` in some automation script anywhere, so I think it's safe to remove? Although I don't mind leaving in the `default` problem type and simply replace that with the new `pass-fail`, similar to how we have the `float` problem type to add the `validator_flags`, and the `custom` problem type to make sure the `output_validators/` folder from the skeleton is included :slightly_smiling_face:

mpsijm (commented):
I think I'm done implementing `type` from the new spec :smile: There's two things I'd like to discuss, see the comments.

For the review: the largest change is that `ProblemSettings.validation` is no longer exposed, since this only works for legacy problems. Instead, the parsing of `validation` (for legacy) or `type` (for `2023-07-draft`) now only happens in the constructor of `ProblemSettings`, resulting in three booleans in both `ProblemSettings` and `Problem`: `interactive`, `multi-pass`, and `custom_output`. The boolean `custom_output` is true for `interactive`, `multi-pass`, and problems with `output_validators/`. In other words, `custom_output` is only false for problems that use the default output validator.

Any other thoughts @RagnarGrootKoerkamp @mzuenni? :smile:

bin/problem.py
mpsijm: Following up in #424 so we can merge this PR for now :slightly_smiling_face:

bin/skel.py
mpsijm: Removed it! :sparkles:

PR #422, opened 2025-02-11
Mark constants in config as Final

mzuenni:
Just annotated some stuff as final and changed some names in config to reflect that they should be final


mpsijm (approved):
Very nice, that's some progress towards #102 again :smile: Thanks! :heart:

PR #423, opened 2025-02-13
#102 More types in util.py

mzuenni:
i am not 100% sure if all types here are correct... but so far nothing breaks

### DIFF
--- a/bin/program.py
+++ b/bin/program.py
@@ -145,7 +145,7 @@ class Program:
                 problem.path.resolve() / self.subdir
             )
             self.short_path = relpath
-            self.name = str(relpath)
+            self.name: str = str(relpath)
             self.tmpdir = problem.tmpdir / self.subdir / relpath
         except ValueError:
             self.short_path = Path(path.name)
--- a/bin/util.py
+++ b/bin/util.py
@@ -712,7 +712,7 @@ write_yaml_lock = threading.Lock()
 
 
 # Writing a yaml file (or return as string) only works when ruamel.yaml is loaded. Check if `has_ryaml` is True before using.
-def write_yaml(data: Any, path: Optional[Path] = None, allow_yamllib: bool = False) -> str | None:
+def write_yaml(data: Any, path: Optional[Path] = None, allow_yamllib: bool = False) -> Optional[str]:
     if not has_ryaml:
         if not allow_yamllib:
             error(
--- a/bin/util.py
+++ b/bin/util.py
@@ -23,11 +23,11 @@ from typing import (
     NoReturn,
     Optional,
     overload,
+    Protocol,
     TextIO,
     TypeAlias,
     TypeVar,
     TYPE_CHECKING,
-    Union,
 )
 from uuid import UUID
 
@@ -50,8 +50,6 @@ except Exception:
 
 if TYPE_CHECKING:  # Prevent circular import: https://stackoverflow.com/a/39757388
     from problem import Problem
-    from program import Program
-    from run import Run
     from verdicts import Verdict
 
 
@@ -140,7 +138,11 @@ class MessageType(Enum):
         }[self]
 
 
-ITEM_TYPE: TypeAlias = Union[str, Path, "Program", "Run"]
+class Named(Protocol):
+    name: str
+
+
+ITEM_TYPE: TypeAlias = str | Path | Named
 
 
 def message(
--- a/bin/util.py
+++ b/bin/util.py
@@ -14,8 +14,21 @@ import threading
 import time
 from enum import Enum
 from collections.abc import Sequence
+from collections.abc import Callable
 from pathlib import Path
-from typing import Any, NoReturn, Optional, TypeVar
+from typing import (
+    Any,
+    Iterable,
+    Literal,
+    NoReturn,
+    Optional,
+    overload,
+    TextIO,
+    TypeAlias,
+    TypeVar,
+    TYPE_CHECKING,
+    Union,
+)
 from uuid import UUID
 
 import yaml as yamllib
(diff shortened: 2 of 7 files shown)


bin/program.py
```diff
@@ -145,7 +145,7 @@ def __init__(
                 problem.path.resolve() / self.subdir
             )
             self.short_path = relpath
-            self.name = str(relpath)
+            self.name: str = str(relpath)
```
mpsijm, line 148: I don't think an explicit type annotation is needed here because `str` deterministically returns `str`, so that can always be inferred, right?

bin/util.py
```diff
@@ -209,65 +237,81 @@ def __init__(
...
 
-    def total_width(self):
+    def total_width(self) -> int:
         cols = ProgressBar.columns
         if is_windows():
             cols -= 1
         return cols
 
-    def bar_width(self):
+    def bar_width(self) -> int | None:
```
mpsijm, line 286:
```suggestion
    def bar_width(self) -> Optional[int]:
```

bin/util.py
```diff
@@ -598,21 +686,33 @@ def read_yaml_settings(path):
...
+        return value  # type: ignore
 
 
 # Only allow one thread to write at the same time. Else, e.g., generating test cases in parallel goes wrong.
 write_yaml_lock = threading.Lock()
 
 
 # Writing a yaml file (or return as string) only works when ruamel.yaml is loaded. Check if `has_ryaml` is True before using.
-def write_yaml(data, path=None, allow_yamllib=False):
+def write_yaml(data: Any, path: Optional[Path] = None, allow_yamllib: bool = False) -> str | None:
```
mpsijm, line 715:
```suggestion
def write_yaml(data: Any, path: Optional[Path] = None, allow_yamllib: bool = False) -> Optional[str]:
```

bin/util.py
```diff
@@ -623,15 +723,12 @@ def write_yaml(data, path=None, allow_yamllib=False):
             return yamllib.dump(data)
         with open(path, "w") as stream:
             yamllib.dump(data, stream)
-        return
+        return None
```
mpsijm, line 726: I think I prefer a blank `return` over `return None` :thinking: Ruff also has a rule for this, not sure why it's not triggered here: https://docs.astral.sh/ruff/rules/unnecessary-return-none/

bin/util.py
```diff
@@ -972,13 +1079,13 @@ def setlimits():
 
 
 # Subclass Popen to get rusage information.
-class ResourcePopen(subprocess.Popen):
+class ResourcePopen(subprocess.Popen[Any]):
```
mpsijm, line 1082: Looks like the generic type parameter `Popen` is either `str` or `bytes`, do we know which of the two? I guess not, and in that case, we could also write `str | bytes` instead of `Any` :slightly_smiling_face:

bin/util.py
```diff
@@ -1150,30 +1257,30 @@ def inc_label(label):
     return "A" + label
 
 
-def combine_hashes(values):
-    values.sort()
+def combine_hashes(values: Sequence[str]) -> str:
+    # values.sort()
```
mpsijm, line 1261: This commented-out line can be removed now:
```suggestion
```

mpsijm (changes requested):
Thanks for the progress towards #102! :heart: Looking good when scrolling through the code :smile: If `mypy` doesn't complain, it's probably fine :innocent: If we want to use any function from `util` in a different way later, we can always rewrite the types :yum: 

The only thing I'm doubting about, is the type alias `ITEM_TYPE`. It looks quite wide, but at the same time, we can't really make it stricter, with how it's currently used :stuck_out_tongue: So I think it's fine :slightly_smiling_face:

bin/util.py
mzuenni: https://github.com/python/mypy/issues/13203

bin/program.py
mzuenni: we probably should annotate all variables, i.e. make clear that `str` is intended and not `Optional[str]` but i did the bare minimum outside of util

mzuenni:
yea `ITEM_TYPE` is weird, its actual `str | path | hasattr(name)` but i don't know how to write that...

mzuenni:
maybe:
```
class Item(Protocol):
    name: str
```
and change `ITEM_TYPE` to `str | Path | Item` ?

mpsijm:
Oohh, `Protocol`s look fancy :smile: Long live duck typing in Python! :duck: 

Maybe we can name it `LogItem` or `ProgressBarItem` (anything less generic than `Item`) :stuck_out_tongue: Or, if we want to stay close to it _is_, rather than what it's _used for_, we could name it something like `SupportsName` (inspiration from https://stackoverflow.com/a/57977926) :slightly_smiling_face:

bin/program.py
mpsijm: > we probably should annotate all variables

We don't _have_ to do this :slightly_smiling_face: We do want to enable the `--strict` flag at some point (see https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/master/.pre-commit-config.yaml#L33), but that only requires function signatures to be fully type-annotated. For fields and variables, MyPy can infer the type from the assigned value, which does not change anymore after its initial assignment. So in this case, `self.name = str(relpath)` will lock the type of `self.name` to `str`, and trying to assign `None` to it later, will throw an error :smile:

mzuenni: Yeah my point here is that I want to pin those attribute types. Even though we don't yet reassign those variables we might want to do that in the future and specifying that we really want a `str` here and not an `Optional[str]` makes this more precise then the automatically interfered `str`.

mpsijm: Fair enough! In that case, we may want to find some automated tooling that enforces type annotations for class fields, but let's leave that for another time (I'll add it to #102) :slightly_smiling_face:

mpsijm (approved):

PR #425, opened 2025-02-15
[problem] Full support for problem.yaml with 2023-07-draft spec

mpsijm:
Since we were already well underway supporting the new features of https://icpc.io/problem-package-format/spec/2023-07-draft.html#problem-metadata, this PR aims to finishes it. I've also set up a test-bench for validating whether `problem.yaml` files are correctly accepted or rejected (along with any expected warnings/errors/fatals). Still to-do in the future (not for this PR):

- [ ] Read `validator_flags` only from `testdata.yaml` for the new spec
	- Keep supporting it for legacy problems
	- Also revert back to `validator_flags` for the DOMjudge export... :smiling_face_with_tear: 
- [ ] Add more tests

In theory, both these points can also be picked up in later PRs to keep the reviews manageable, so feedback is welcome already :smile: Specifically, please check whether all fields for both `legacy` and `2023-07-draft` now have a one-to-one mapping between the specification and the `ProblemSettings` initializer. I've tried to keep the order of parsing as close as possible to the order in the spec :slightly_smiling_face:

### DIFF
--- a/bin/problem.py
+++ b/bin/problem.py
@@ -89,6 +89,59 @@ class ProblemCredits:
                 warn(f"found unknown problem.yaml key: {key} in credits")
 
 
+class ProblemSource:
+    def __init__(self, name: str, url: Optional[str] = None):
+        self.name = name
+        self.url = url
+
+    def __repr__(self) -> str:
+        return self.name + (f" ({self.url})" if self.url else "")
+
+
+class ProblemSources(list[ProblemSource]):
+    def __init__(
+        self,
+        yaml_data: dict[str, Any],
+        problem_settings: "ProblemSettings",
+    ):
+        # If problem.yaml uses the legacy version, do not support the new type of the `source` key.
+        # If problem.yaml uses 2023-07-draft, prefer `source`, but also support `source_url` and warn for it.
+        legacy_source_url = parse_optional_setting(yaml_data, "source_url", str)
+        if problem_settings.is_legacy():
+            source_name = parse_setting(yaml_data, "source", "")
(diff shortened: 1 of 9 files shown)


bin/problem.py
```diff
@@ -123,34 +250,41 @@ def __init__(
...
+            mode = parse_legacy_validation(parse_setting(yaml_data, "validation", "default"))
         else:
-            if "validation" in yamldata:
+            if "validation" in yaml_data:
                 warn(
-                    "problem.yaml: 'validation' is removed in 2023-07-draft, please use 'type' instead"
+                    "problem.yaml: 'validation' is removed in 2023-07-draft, please use 'type' instead. SKIPPED."
+                )
+                yaml_data.pop("validation")
+            mode = set(parse_setting(yaml_data, "type", "pass-fail").split(" "))
```
mzuenni, line 280: type an also be of type list right?

mzuenni (commented):
(I did not check everything I detail, added some comments)

mpsijm:
Thanks for your first comments, all good points! I've fixed them :smile: 

Are you still planning to check in more detail, or can we simply merge? :slightly_smiling_face:

mzuenni:
I can take a closer look, but likely only in a few days

mpsijm:
All right! Next week Wednesday, we're kicking off with the Freshmen Programming Contests, so I'd like to merge a few days before that, so I can prepare the repo with the latest version of the spec 🙂

mzuenni:
I read through it again, looks fine to me.

PR #426, opened 2025-02-19
More consistent floating-point parsing in validators

mpsijm:
Follow-up of https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/420#discussion_r1961484595. The outcome of this discussion is that input and answer files should look "clean". We already enforce case- and whitespace-sensitivity for `*.in` and `*.ans` files in `validation.h`, and this PR adds strict floating-point parsing to this.

Note that output validation does not use the strict floating-points, because from submissions, we want to accept any valid floating-point number (as defined in [the spec](https://icpc.io/problem-package-format/spec/2023-07-draft.html#general-requirements)), regardless of how ugly it looks :stuck_out_tongue: 
- In `validation.h`, we accept any floating-point number that is accepted by `stold` for `OutputValidator` by passing `strict_float=false` to the `Validator` super-constructor. I don't think we need to add a flag to enable stricter floating-point parsing for custom output validators.
- In `default_output_validator.cpp`, the `util::is_float` function now checks against the RegEx based on the grammar from the spec. 
  - Note that this validator is only used when running BAPCtools and not during a contest, because there are different "default output validators" in [DOMjudge](https://github.com/DOMjudge/domjudge/blob/main/sql/files/defaultdata/compare/compare.cc) and [Kattis](https://github.com/Kattis/problemtools/blob/develop/support/default_validator/default_validator.cc).

I've tested this on Kruidnoten from NWERC 2024 (which has floating-point answers), and the warnings in `bt validate --generic-invalid` are now gone :smile:

### DIFF
--- a/headers/validation.h
+++ b/headers/validation.h
@@ -1,15 +1,17 @@
 #pragma once
 // A header library to safely parse team input.
-// It does not support floating points or big integers.
+// It does not support big integers.
 // Author: Ragnar Groot Koerkamp
 
 // The easiest way to use this is to symlink it from a validator directory,
 // so that it will be picked up when creating a contest zip.
 
-// The default checking behaviour is lenient for both white space and case.
-// When validating .in and .ans files, the case_sensitive and
-// space_change_sensitive flags should be passed. When validating team output,
-// the flags in problem.yaml should be used.
+// The default checking behaviour is strict for case, whitespace, and floats.
+// This strict checking mode is used for *.in and *.ans files.
+// When validating submission outputs, the checking is more lenient,
+// but the case_sensitive and space_change_sensitive flags can be passed
+// via the output_validator_args in testdata.yaml to enable strict checking behaviour
+// for submission outputs regarding case and whitespace, respectively.
 
 #include <algorithm>
 #include <array>
@@ -25,6 +27,7 @@
 #include <map>
 #include <optional>
 #include <random>
+#include <regex>
 #include <set>
 #include <stdexcept>
 #include <string>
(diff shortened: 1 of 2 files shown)


mzuenni:
I actually prefer to not use regex in c++ as they are kind of broken, but this should be fine

mpsijm:
I got the idea for using a RegEx from https://github.com/mzuenni/icpc-header/blob/main/src/validate.h#L142-L143 actually, since you mentioned that your header also had a `realStrict` function :stuck_out_tongue: 

I prefer the RegEx over the manual parsing, because it's more declarative rather than imperative, and it's easy to verify that the grammar in the spec matches the RegEx.

Do you know what exactly is broken in C++ `std::regex`? I've found online that it's slower that it should be, but I think it's still fast enough™ for our purposes? If it does add some slowness, we'll only have it during development, not during a contest :slightly_smiling_face: The alternative is including e.g. https://github.com/hanickadot/compile-time-regular-expressions, but I don't think adding a header file of 15k is a very good idea either :joy:

mzuenni:
> I got the idea for using a RegEx from https://github.com/mzuenni/icpc-header/blob/main/src/validate.h#L142-L143 actually, since you mentioned that your header also had a realStrict function 😛

🙈

> Do you know what exactly is broken in C++ std::regex?
- There are no performance guarantees at all, but that shouldn't be too much of a problem in our case
- They are slow kind of "by design" and this will never be change

Yea including a better regex parser is not an option.
In theory we could write a limited regex parser which would also allow us to generate strings that match the regex...

mpsijm:
Writing our own limited regex parser would add a lot of code complexity though :stuck_out_tongue: I think it's fine to simply use `std::regex` then, to keep the code simple, even though it's indeed not as performant as we'd want it to be.

mzuenni:
ready to merge?

PR #428, opened 2025-02-21
Force the use of output_validator_args in testdata.yaml, instead of validator_flags in problem.yaml

mpsijm:
~~I still want to double-check and double-test everything, but opening the draft PR already so others can also do that if they feel like it :slightly_smiling_face:~~

EDIT: Ready for review now :smile:


bin/interactive.py
```diff
@@ -53,9 +54,13 @@ def get_validator_command():
                 run.testcase.ans_path.resolve(),
                 run.feedbackdir.resolve(),
             ]
-            + run.problem.settings.validator_flags
+            + run.testcase.testdata_yaml_validator_args(
+                output_validator,
+                bar or PrintBar("Run innteractive test case"),
```
mzuenni, line 59: typo: interactive :)

mpsijm:
Since I'd like to use this for the Freshmen Programming Contests, for which we'll start work today, I'll merge already. I've tested this on some floating-point output problems and the new problems for the FPCs, and it's all looking good! Feedback on the code is still welcome though, even after merging :smile:

PR #429, opened 2025-02-23
Use singular names for (in)valid input/answer/output directories

mzuenni:


RagnarGrootKoerkamp:
Oh, dont we instead want to make the spec plural?

mzuenni:
its `sample` and `secret` not `samples` and `secrets`, so singular is more consistent?

mpsijm:
- Do we know what has been used outside BAPCtools? Kattis' Problemtools doesn't seem to support `(in)valid_*` at all, are there other tools that follow the spec?
- Should we keep backwards-compatibility with the plural form? I don't want to, because that's gonna be a mess... 😅 I think there's only one or two years where we haae used this, so the damage shouldn't be too bad? 🙈 Maybe we could even auto-migrate when parsing `generators.yaml` to keep the mess contained?

mzuenni:
we can add a warning if we find the plural form of the dirs in the problem directory

bin/problem.py
```diff
@@ -403,6 +403,10 @@ def __init__(self, path: Path, tmpdir: Path, label: Optional[str] = None):
 
         self.statement_languages = self._determine_statement_languages()
 
+        for d in ["invalid_inputs", "invalid_answers", "invalid_outputs", "valid_outputs"]:
+            if (self.path / "data" / d).is_dir():
+                warn(f"Found: data/{d}, should be: data/{d[:-1]} (singular form).")
```
mpsijm, line 408: Would this warning (also) fit in the processing of `generators.yaml`?

mpsijm:
https://github.com/Kattis/problem-package-format/issues/355#issuecomment-2677075793 mentioned that "input" and "output" are technically uncountable, so can't have a plural-`s`. Then I realized, in BAPCtools, we extend this with `invalid_answer(s)`, but the noun "answer" _is_ actually countable. :joy: I'm still fine with making everything consistently singular, but note that the correct grammatical interpretation of this then becomes:
- `invalid_input` contains test cases that have invalid input
- `invalid_answer` contains test cases that have _an_ invalid answer
- `invalid_output` contains test cases that have invalid output

Which is slightly asymmetric, but probably fine :joy: Alternatively, when expanding the last part to "an invalid input/answer/output file" (https://github.com/Kattis/problem-package-format/issues/355#issuecomment-2677088529), it _is_ symmetric. Close enough :stuck_out_tongue:

mpsijm (approved):
Time to migrate all problems from last year, haha :joy:

mzuenni:
ready to merge?

PR #430, opened 2025-02-24
Implement constants using environment variables

mzuenni:


mzuenni:
I implemented this to toy around with. So far I like the behavior.

The only issue I see is that the judge has to set the environment variables we can't easily substitute them. Should we write some kind of wrapper script for the export which sets the environment variables?

latex/contest-problem.tex
```diff
@@ -4,6 +4,7 @@
 	\renewcommand{\problemauthor}{{%problemauthor%}}
 	\renewcommand{\timelimit}{{%timelimit%}}
 	\input{{%builddir%}/samples.tex}
+	\input{{%builddir%}/constants.tex}
```
RagnarGrootKoerkamp, line 7: Can/should we scope the constants, so that they don't acidentally leak into the next problem?

mzuenni: `\begingroup` and `\endgroup` should already scope this

PR #431, opened 2025-02-26
Implement constants

mzuenni:
Replaces #368 and #430, see also https://github.com/Kattis/problem-package-format/issues/344.

TODO:
- [x] Inline constants in ZIP export
- [x] add test problem
- [x] ~~substitute in `generators.yaml`?~~


mzuenni:
I rewrote the export to create a "copy" (links if possible) in the tmpdir that then gets zipped.
@mpsijm can you fix the conflict with args and take a look at the export changes?

mzuenni:
>Is it intentional that you don't rewrite the \constant{<key>} commands? It's probably fine, because DOMjudge doesn't read the TeX anyway, and Kattis will (hopefully) be able to handle constants. Or if not, we can always rewrite later 😛

yea its intentionally not replaced :)

mzuenni:
@mpsijm do you know how to add a test to call `bt pdf`, `bt solutions` and `bt validate` on the constants problem?

RagnarGrootKoerkamp:
For generator command substitution we could consider using the classic {n} syntax and then also support expressions Witt it as {n//2} and such. That's also consistent with {seed}

mzuenni:
> For generator command substitution we could consider using the classic {n} syntax and then also support expressions Witt it as {n//2} and such. That's also consistent with {seed}

using just `{}` would result in conflicts for count and seed which is kind of weird. And i dont think we want to support more expressions here that sounds kind of arbitrary?

RagnarGrootKoerkamp:
But wouldn't it be nice if we can just do `command.format(count=count,seed=seed,**substitutions)` or so?
I can totally see it being nice to do a bunch of `{n/10}` `{n-1}`, `{sqrt(n)}` kind of stuff.

(Optionally we could make it a top-level generators.yaml flag `eval_expressions: true` but not a fan of that probably)

mzuenni:
> But wouldn't it be nice if we can just do command.format(count=count,seed=seed,**substitutions) or so?

seed has to be evaluated differently anyway... and there are conflicts. I dont see the greater benefit.

> I can totally see it being nice to do a bunch of {n/10} {n-1}, {sqrt(n)} kind of stuff.

yea i see the point, but I am not sure if we want to have that evaluated while parsing. You can still write `{{n}}-1` and evaluate that inside the generator?

mpsijm:
I agree with @mzuenni on this one – evaluating arbitrary expressions inside the `{{constant}}` braces is quite the can of worms.

> You can still write `{{n}}-1` and evaluate that inside the generator? 

This, exactly: we already sometimes write generators that use `eval()` to parse its arguments to make life easier. But then that's only when you have/want/need to, instead of always available :stuck_out_tongue:

mzuenni:
so, as soon as this gets in the draft we can merge?

PR #432, opened 2025-03-03
add bt upgrade

mzuenni:


bin/upgrade.py
```diff
@@ -0,0 +1,248 @@
...
+                bar.error(
+                    f"can't rename 'data.{old_name}', 'data.{new_name}' already exists in generators.yaml",
+                    resume=True,
+                )
+                continue
+            bar.log(f"renaming 'data.{old_name}' to 'data.{new_name}' in generators.yaml")
+            data[new_name] = data[old_name]
+            data.pop(old_name)
+
+    # TODO fix testdata.yaml here too?
```
mpsijm, line 53: Yes, we should move `validator_flags` in `problem.yaml` to `output_validator_args` in `testdata.yaml` in the root of `generators.yaml` :slightly_smiling_face: 

Also, if the `generators.yaml` already contains `testdata.yaml` at any level (not only in root), we should probably migrate `*_validator_flags` to `*_validator_args`?

bin/upgrade.py
```diff
@@ -0,0 +1,248 @@
...
+    write_yaml(data, generators_yaml)
+
+
+def upgrade_statement(p: problem.Problem, bar: ProgressBar) -> None:
+    if (p.path / "problem_statement").is_dir():
+        if (p.path / "statement").exists():
+            bar.error("can't rename 'problem_statement/', 'statement/' already exists", resume=True)
+            return
+        bar.log("renaming 'problem_statement/' to 'statement/' in generators.yaml")
+        (p.path / "problem_statement").rename(p.path / "statement")
```
mpsijm, line 63: Note that we cannot merge this PR until we can actually compile LaTeX from `statement/`.

bin/upgrade.py
```diff
@@ -0,0 +1,248 @@
...
+                    limits.pop("time_multiplier")
+
+                if "time_safety_margin" in limits:
+                    limits["time_multipliers"]["time_limit_to_tle"] = limits["time_safety_margin"]
+                    limits.pop("time_safety_margin")
+
+    if "validator_flags" in data:
+        ...
+        # TODO fix this...
+        # either write a top level testdata.yaml or add it to generators.yaml ?
```
mpsijm, line 162: The latter, as per the comment above. We can't write it to `data/testdata.yaml` directly, because that'll get removed again if it's not in `generators.yaml` :stuck_out_tongue:

bin/upgrade.py
```diff
@@ -0,0 +1,248 @@
...
+
+def _upgrade(p: problem.Problem, bar: ProgressBar) -> None:
+    bar.start(p)
+
+    upgrade_data(p, bar)
+    upgrade_generators_yaml(p, bar)
+    upgrade_statement(p, bar)
+    upgrade_problem_yaml(p, bar)
+    upgrade_testdata_yaml(p, bar)
+    # TODO: output_validators -> output_validator
```
mpsijm, line 235: :+1:, pending rework of how we handle the asymmetric validators structure (i.e. multiple input/answer validators vs. singular output validator, and maybe the same for visualizers)

mzuenni: we already allow only one output validator?

bin/problem.py
```diff
@@ -223,34 +160,22 @@ def __init__(
...
-        legacy_time_limit = _get_legacy_time_limit()
-        self.time_limit: float = time_limit or legacy_time_limit or 1.0
-        self.time_limit_is_default: bool = time_limit is None and legacy_time_limit is None
+        # warn for deprecated timelimit files
+        if (problem.path / ".timelimit").is_file():
+            warn("A .timelimit file is DEPRECATED. Use limits.time_limit instead.")
+        if (problem.path / "domjudge-problem.ini").is_file():
+            warn(
+                "domjudge-problem.ini is DEPRECATED. Use limits.time_limit if you want to set a timelimit."
+            )
```
mpsijm, line 169: If we're going for a hard cut, I think we can remove this as well :stuck_out_tongue:

bin/upgrade.py
```diff
@@ -22,15 +25,37 @@ def upgrade_data(p: problem.Problem, bar: ProgressBar) -> None:
             old_path.rename(new_path)
 
 
+def upgrade_testdata_yaml(p: problem.Problem, bar: ProgressBar) -> None:
+    rename = [
+        ("output_validator_flags", "output_validator_args"),
+        ("inut_validator_flags", "inut_validator_args"),
+    ]
+
+    for f in (p.path / "data").rglob("testdata.yaml"):
```
mpsijm, line 39: I don't think `bt upgrade` needs to touch anything in `data/`, since it will be overwritten by `bt generate` anyway?

bin/upgrade.py
mpsijm: Yes, but I mean that currently, the structure is:
```
input_validators/
  validator_1/
    ...files for validator 1...
  validator_2/
    ...files for validator 2...
output_validators/
  only_validator/
    ...files for only validator...
```
But for the output validator, we should change to:
```
output_validator/
  ...files for outout validator...
 ```

File: bin/util.py [hunk 7]
```diff
@@ -793,6 +793,15 @@ def parse_optional_list_setting(yaml_data: dict[str, Any], key: str, t: type[T])
     return []
 
 
+def parse_deprecated_setting(
+    yaml_data: dict[str, Any], key: str, new: Optional[str] = None
+) -> None:
+    if key in yaml_data:
+        use = f", use '{new}' instead" if new else ""
+        warn(f"key '{key}' is deprecated{use}. SKIPPED.")
+        yaml_data.pop(key)
```
mpsijm, line 802: If we're going for a hard cut, I don't think we need this either? :innocent:

bin/problem.py
mzuenni: i think a warning os still nice for now

bin/upgrade.py
mzuenni: not every Problem has a `generator.yaml`

bin/upgrade.py
mzuenni: yes, but I would expect that this does not require a big change in the code?

mpsijm: Sure, but we do still need to do this at some point :stuck_out_tongue: I'll resolve this thread for now though, because that rewrite should happen outside this PR :slightly_smiling_face:

test/problems/divsort/generators/generators.yaml
```diff
@@ -2,7 +2,7 @@ solution: /submissions/accepted/th.py
 data:
   sample:
     data:
-      - "":
+      - '':
```
mpsijm, line 5: Can we set `preserve_quotes = True` in the YAML reader/writer to avoid silly diffs like this?

(also, I think we should force double-quotes for YAML strings that BAPCtools creates, but that's probably out-of-scope here :stuck_out_tongue: )

test/problems/identity/generators/generators.yaml
```diff
@@ -89,36 +89,36 @@ data:
...
-          - "": stdout.py 124
-          - "": stdout.py 125
-          - "": stdout.py 126
-          - "": stdout.py 127
-          - "": stdout.py 128
-          - "": stdout.py 129
-          - "": stdout.py 130
-          - "": stdout.py 131
-          - "": stdout.py 132 # 10.in
+          - null: stdout.py 123 # 01.in
```
mpsijm, line 92: Funky, does a literal `null` actually come out as `None` in Python? :joy: Perhaps better to keep using `~` in this example, that's more explicit. I don't think we need any special handling in `bt upgrade` for this though

test/problems/identity/generators/generators.yaml
```diff
@@ -169,16 +169,8 @@ data:
...
-        {
-          data:
-            [
-              { "": stdout.py 600 },
-              { "1": stdout.py 601 },
-              { "": stdout.py 602 },
-            ],
-        }
+          - '': stdout.py 503
+      json: {data: ['': stdout.py 600, '1': stdout.py 601, '': stdout.py 602]}
```
mpsijm, line 173: Can you revert this change? I think the extra newlines and indentation are helpful to understanding the syntax :slightly_smiling_face:

test/problems/interactivemultipass/problem.yaml
```diff
@@ -1,8 +1,12 @@
...
 rights_owner:
-validation: custom interactive multi-pass
+problem_format_version: 2023-07-draft
+type:
+  - pass-fail
+  - interactive
+  - multi-pass
+credits:
+  authors:
+    - Michael Zündorf
```
mpsijm, line 12: Do we want `bt upgrade` to use the short-hand syntax of `credits: <name>` when possible?

test/problems/interactivemultipass/problem.yaml
```diff
@@ -1,8 +1,12 @@
...
-author: Michael Zündorf
-source:
-source_url:
 uuid: 42c9ed2e-f579-46ac-ae2c-191069f3df70
 license: unknown
 rights_owner:
-validation: custom interactive multi-pass
+problem_format_version: 2023-07-draft
+type:
+  - pass-fail
```
mpsijm, line 7: Isn't `pass-fail` implied when not given?

test/problems/multipass/problem.yaml
```diff
@@ -1,8 +1,11 @@
...
-source:
-source_url:
 uuid: 71076b69-e9c2-4227-ba54-ec3d4e277c78
 license: unknown
 rights_owner:
-validation: custom multi-pass
+problem_format_version: 2023-07-draft
+type:
+  - pass-fail
+  - multi-pass
```
mpsijm, line 8: Same here, this could be `type: multi-pass`, right?

test/test_problems.py
```diff
@@ -215,7 +215,7 @@ def test_new_contest_problem(self, monkeypatch):
 
 class TestReadProblemConfig:
     def test_read_problem_config(self):
-        p = problem.Problem(RUN_DIR / "test/problems/test_problem_config", Path("/tmp/xyz"))
+        p = problem.Problem(RUN_DIR / "test/problems/testProblemConfig", Path("/tmp/xyz"))
```
mpsijm, line 218: I'm guessing your removed the underscores to follow the spec, but capital letters are also not allowed:

> The directory name must consist solely of lowercase letters a–z and digits 0–9.

test/problems/hellounix/problem.yaml
```diff
@@ -1,27 +1,13 @@
...
-# Time limit is 2*slowest accepted submission:
-#     time_multiplier: 2
-# Warning for submissions within 1 second of limit:
-#     time_safety_margin: 1
 limits:
   memory: 512
+  time_limit: 3.0
+problem_format_version: 2023-07-draft
+type:
+  - pass-fail
```
mpsijm, line 10: Similar shorthand comment for all problems that could have `type: pass-fail` rather than
```yaml
type:
  - pass-fail
```

test/problems/interactivemultipass/problem.yaml
mzuenni: yes, but i prefer `bt upgrade` to be explicit about this.

test/test_problems.py
mzuenni: yea... didn't check the warning properly XD

test/problems/identity/generators/generators.yaml
mzuenni: no idea, but wil fix :D

test/problems/multipass/problem.yaml
mzuenni: yea I will change this to only add `pass-fail` if nothing else is in type

mzuenni:
maybe we want to have a dedicated draft branch where we can build stuff until the draft is published?
i.e. we could merge this, the constants pr and my (still local WIP) latex branch

RagnarGrootKoerkamp:
Works for me. Alternatively we could create a new branch (or tag) `legacy` at current master and mention this in the readme?

Or do you think the breaking changes from people pulling will be to much?

mzuenni:
the second issue should be fixed now, the first one looks annoying... but we are not only keeping some spurious comments, we are also removing others right? ^^'

mzuenni:
@mpsijm i tried to address both issues, its a little bit hacky... but upgrading is only a "best effort" service IMO so this is fine?

mpsijm:
> we are also removing others right? ^^'

Yes, that was the case initially :see_no_evil: But hopefully not anymore :smile: 

I've improved the preserving/moving of comments and replacing of keys a bit further. Please test if this doesn't break other things, but if you don't find anything broken, I think we can merge this into `draft` :smile:

PR #412, opened 2024-12-12
Submissions and expectations

thorehusfeldt:

PR #433, opened 2025-03-05
Draft

### LINKED ISSUE #295: `bt upgrade` port old problems to new spec

### LINKED ISSUE #424: Support singular output_validator/ for 2023-07-draft spec
Follow-up of https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/421#discussion_r1953332548.

Currently (in the `legacy` and `legacy-icpc` specs), there can be multiple `output_validators/`. But according to the new spec (https://icpc.io/problem-package-format/spec/2023-07-draft.html#problem-package-structure-overview), there can be only one output validator, that lives in `output_validator/`. This makes sense, because if there are multiple output validators, this forces the problem packager to decide in which order these should be executed (using `build` and `run` scripts). We'd like to support this in BAPCtools as well, however:

- This is currently not yet supported by either DOMjudge (https://github.com/DOMjudge/domjudge/blob/main/webapp/src/Service/ImportProblemService.php#L932) or Kattis' (https://github.com/Kattis/problemtools/blob/develop/problemtools/verifyproblem.py#L1465-L1466), so we'd have to revert this again in `bt export`.
- This change would break the symmetry with the input/answer validators, which _do_ allow multiple validators.
  - On the other hand, we may also want to implement the singular `{input,output}_visualizer/` and `static_validator/` at some point, so it makes sense to set up our architecture to support these as well. But as long as we don't, it may not be worth it.
  - For the output visualizer, also see https://github.com/DOMjudge/domjudge/pull/2744.

See also the comments in `problem.py` and `validate.py` marked with `# TODO #424`, and https://github.com/DOMjudge/domjudge/issues/2887.

mzuenni:
- [x] constants (#431)
- [x] bt upgrade (#432, closes #295)
- [x] new latex dirs (#434)
- [x] output_valdiator (#435, closes #424)
- [x] Add .download samples and ans_is_output flag (#436)
- [x] Update parsing of problem.yaml based on Kattis/problem-package-format#372 (#437)
- [x] Input/output visualizer (#448)
- [x] Legacy export (#441)
- [x] [validate] Remove {input,answer}_format_validators (#443)
- [x] Remove support for `data/bad/` (#445)
- [x] Extend parsing of persons in `credits` to a map of strings: https://github.com/Kattis/problem-package-format/pull/415 (fixed in https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/433/commits/abb49f89f1aeca579631c2aca0a6d5baf28b20bb)
- [x] Rename `testdata.yaml` to `test_group.yaml` and split off `<testcase>.yaml` from it (#456)
  - Side-note: make sure that `args` and `full_feedback` are accepted when parsing, but we don't need to implement them per se (we should warn if we indeed don't)
- [x] Move `*.hint` and `*.desc` to `<testcase>.yaml` keys `hint:` and `description:` (#456)
- [ ] If we feel like it: properly check combinations of `license` and `rights_owner` (see https://github.com/Kattis/problem-package-format/pull/391 and https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/437#issuecomment-2727670053)

Not part of the spec, but now that we're changing things anyway:
- [x] Rename `testsession` in `contest.yaml` to `test_session` (fixed in https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/433/commits/c5adbcfa572c5969f2cc6a19f384477a81fde493)

New features to BAPCtools that can also be added separately from upgrading the current implementation from `legacy` to `2023-07-draft`:
- [ ] static_validator
- [ ] expectations @thorehusfeldt (see also #307 and #412)


mzuenni:
>  Move *.hint and *.desc to testcase.yaml keys hint: and description:

do we want to move those keys or add a new key `yaml` here that gets written verbatim?

mpsijm:
Good question, I think the latter. Since `<testcase>.yaml` contains more keys (see [the spec](https://icpc.io/problem-package-format/spec/2023-07-draft.html#test-case-configuration)), I think it makes sense to have a key `yaml` that contains this extra configuration, since this configuration should be written to a `.yaml` file with the same base name as the test case. So, we could have something like this, where we get `1.in`, `1.ans`, and `1.yaml`:
```yaml
data:
  sample:
    "1":
      in: Hello
      ans: World!
      yaml:
        args: -n 42  # This makes no sense in the Hello World example, I just wanted to spam all the keys 😛
        input_validator_args: case_insensitive
        output_validator_args: case_insensitive
        full_feedback: True
        hint: Greet the world!
        description: A test case in which the world is greeted.
```

mpsijm:
Closing this PR, since everything is in the `main` branch now :smile:

PR #434, opened 2025-03-05
Split up `problem_statement/` into `statement/`, `solution/`, and `problem_slide/`

mzuenni:


bin/export.py
```diff
@@ -242,6 +242,10 @@ def add_file(path, source):
                     f.unlink()
                     f.write_text(text)
 
+    # TODO: Remove this if we know others import the statement folder
+    if (export_dir / "statement").exists():
+        (export_dir / "statement").rename(export_dir / "problem_statement")
```
mpsijm, line 247: We should probably also pull the contents of `solution/` and `problem_slide/` into `problem_statement/` during the export?

bin/latex.py
```diff
@@ -370,10 +374,9 @@ def build_problem_pdf(
     Arguments:
     -- language: str, the two-letter language code appearing the file name, such as problem.en.tex
     """
-    main_file = build_type.value
-    main_file += "-web.tex" if web else ".tex"
+    main_file = build_type.path(ext="-web.tex" if web else ".tex").name
 
-    bar = PrintBar(f"{main_file[:-3]}{language}.pdf")
+    bar = PrintBar(f"{main_file[:4]}.{language}.pdf")
```
mpsijm, line 379: This now renders as `prob.en.pdf` or `solu.en.pdf`, that's probably not intended? :stuck_out_tongue:

bin/problem.py
```diff
@@ -333,9 +333,7 @@ def _determine_statement_languages(self):
         If problem.yaml's name key is a string, convert into dict; assume `en` as default language.
         """
         yamllangs = set(self.settings.name)
-        texlangs = set(
-            path.suffixes[0][1:] for path in glob(self.path, "problem_statement/problem.*.tex")
-        )
+        texlangs = set(path.suffixes[0][1:] for path in glob(self.path, "statement/problem.*.tex"))
```
mpsijm, line 336: It's probably best to move this glob string to the enum as well, for consistency and single-source-of-truth?

bin/problem.py
```diff
@@ -348,7 +346,7 @@ def _determine_statement_languages(self):
         for lang in texlangs & yamllangs:
             unnormalised_yamlname = self.settings.name[lang]
             yamlname = " ".join(unnormalised_yamlname.split())
-            with open(self.path / "problem_statement" / f"problem.{lang}.tex") as texfile:
+            with open(self.path / "statement" / f"problem.{lang}.tex") as texfile:
```
mpsijm, line 349: This could be `PdfType.PROBLEM.path(lang)` as well, I think

bin/skel.py
```diff
@@ -337,7 +339,9 @@ def problem_source_dir(problem: Problem):
     contest_yml = (config.TOOLS_ROOT / "skel/gitlab_ci/contest.yaml").read_text()
     contest_path = Path(".").resolve().relative_to(git_root_path)
     changes = "".join(
-        "      - " + str(problem_source_dir(problem)) + "/problem_statement/**/*\n"
+        f"      - {problem_source_dir(problem)}/{latex.PdfType.PROBLEM.path().parent}/**/*\n"
+        f"      - {problem_source_dir(problem)}/{latex.PdfType.PROBLEM_SLIDE.path().parent}/**/*\n"
+        f"      - {problem_source_dir(problem)}/{latex.PdfType.SOLUTION.path().parent}/**/*\n"
         for problem in problems
     )
```
mpsijm, line 346: I think something like this should work :smile:
```suggestion
    changes = "".join(
        f"      - {problem_source_dir(problem)}/{pdf_type.path().parent}/**/*\n"
        for problem in problems
        for pdf_type in latex.PdfType
    )
```

bin/stats.py
```diff
@@ -48,8 +49,8 @@ def problem_stats(problems):
         # Roughly in order of importance
         ("  time", lambda p: p.limits.time_limit, 0),
         ("yaml", "problem.yaml"),
-        ("tex", "problem_statement/problem*.tex", 1),
-        ("sol", "problem_statement/solution*.tex", 1),
+        ("tex", str(latex.PdfType.PROBLEM.path(ext="*.tex")), 1),
```
mpsijm, line 52: This now no longer matches `problem-slide.*.tex`, but that's probably fine?

I also don't think we need to add the problem slide as separate item to `bt stats`, because not many people use this, I think.

bin/tools.py
```diff
@@ -1217,7 +1217,7 @@ def run_parsed_arguments(args):
                     )
                 # Only build the problem slides if at least one problem has the TeX for it
                 if any(
-                    glob(problem.path / "problem_statement", "problem-slide.*.tex")
+                    glob(problem.path / "problem_slide", "problem-slide.*.tex")
```
mpsijm, line 1220: Another glob that could be defined in the enum

latex/solution.tex
```diff
@@ -1,13 +1,13 @@
 \documentclass[rgb,dvipsnames,aspectratio=169,9pt,t]{beamer}
 \input{solutions-base.tex}
 \begin{document}
-\begingroup\graphicspath{{{%problemdir%}/problem_statement/}}
+\begingroup\graphicspath{{{%problemdir%}/solution/}{{%problemdir%}/statement/}}
```
mpsijm, line 4: It is so nice that this just works :heart_eyes: Tested this on `nwerc2024/practice/blackboard`, both with the `state*.jpg`s in `statement/` and in `solution/`, and both compile just fine :relaxed:

bin/upgrade.py
```diff
@@ -160,9 +160,9 @@ def upgrade_statement(problem_path: Path, bar: ProgressBar) -> None:
...
             bar.error("can't rename 'problem_statement/', 'statement/' already exists", resume=True)
-            return
-        bar.log("renaming 'problem_statement/' to 'statement/'")
-        (problem_path / "problem_statement").rename(problem_path / "statement")
+        else:
+            bar.log("renaming 'problem_statement/' to 'statement/'")
+            (problem_path / "problem_statement").rename(problem_path / "statement")
 
     origin = problem_path / "statement"
     move = [
```
mpsijm, line 168: Talking about images, we _could_ detect which images are only used in `solution.en.tex` and move those to `solution/` as well, but that's much more effort than needed, and things will just work when all the images are in `statement/` :joy:

mpsijm (changes requested):

bin/problem.py
mzuenni: problem can't import latex because of circular imports ): any idea?

bin/tools.py
mzuenni: circulat import again

bin/upgrade.py
mzuenni: yeah... best effort service is enough here :p

bin/export.py
mzuenni: tried to implement this :)

latex/solution.tex
mzuenni: :D

bin/problem.py
mpsijm: Ah :sweat_smile: I ran into this before. I have two ideas:
- In `latex.py`, we could move `import problem` into an `if TYPE_CHECKING:` block
- We could move the `PdfType` enum to `util.py`

bin/export.py
mpsijm: For some reason, the contents of `solution/` and `problem_slide/` now end up in `statement/` rather than `problem_statement/`? :joy:

bin/stats.py
```diff
@@ -48,8 +49,8 @@ def problem_stats(problems):
...
+        ("sol", str(latex.PdfType.SOLUTION.path(ext="*.tex")), 1),
```
mpsijm, line 53: For the other globs, you wrote `str(latex.PdfType.PROBLEM.path("*"))`. Do we want that here as well?

bin/tools.py
```diff
@@ -1216,10 +1216,8 @@ def run_parsed_arguments(args):
                         web=True,
                     )
                 # Only build the problem slides if at least one problem has the TeX for it
-                if any(
-                    glob(problem.path / "problem_statement", "problem-slide.*.tex")
-                    for problem in problems
-                ):
+                slideglob = latex.PdfType.PROBLEM_SLIDE.path("*")
+                if any(problem.path.glob(slideglob) for problem in problems):
```
mpsijm, line 1220: Here, the glob is still a `Path` and not converted to `str()`. Does that mean we also don't need it in other places?

bin/stats.py
mzuenni: hm i did this to match the old glob (`*.tex` vs `.*.tex`) but  yes `.*.tex` should be fine here

bin/tools.py
mzuenni: i think it depends on the glob function we are using...

bin/stats.py
mpsijm: In that case, maybe it _was_ intentional that the glob also matched `problem-slide.*.tex` :joy: But as agreed upon, we don't have to do that anymore here :slightly_smiling_face:

mzuenni (commented):
Regarding your last commit:
This is not true. `add_args` had an early return on error. Can you revert this change?

mpsijm:
Woops, good point! I removed the commit, thanks for spotting that :slightly_smiling_face:

mzuenni:
ready to merge?

mpsijm (approved):
I think so! And if we find anything later, we're still on the `draft` branch anyway 😄

PR #435, opened 2025-03-05
Output validator

mzuenni:


mzuenni:
@mpsijm  Can this be merged?

mpsijm:
I'd like to review and try it first, but didn't have the time yet 😅 Definitely this weekend, maybe earlier 🙂

bin/problem.py
```diff
@@ -818,15 +816,16 @@ def _validators(
...
         assert hasattr(cls, "source_dirs")
-        # TODO #424: We should not support multiple output validators inside output_validator/.
-        paths = [p for source_dir in cls.source_dirs for p in glob(problem.path / source_dir, "*")]
-
-        # Handle default output validation
         if cls == validate.OutputValidator:
-            if not paths:
-                if problem.custom_output:
-                    fatal("Problem validation type requires output_validators/")
+            assert len(cls.source_dirs) == 1
```
mpsijm, line 820: Is this assertion needed? It's a class-constant, and should not change. And we may want to make it a single string, rather than a list :slightly_smiling_face:

bin/validate.py
```diff
@@ -333,12 +333,11 @@ class OutputValidator(Validator):
...
 
     def __init__(self, problem, path, **kwargs):
-        super().__init__(problem, path, "output_validators", **kwargs)
+        super().__init__(problem, path, "output_validator", **kwargs)
 
     validator_type = "output"
 
-    # TODO #424: We should not support multiple output validators inside output_validator/.
-    source_dirs = ["output_validator", "output_validators"]
+    source_dirs = ["output_validator"]
```
mpsijm, line 340: We may be able to write this as `source_dir = "output_validator"`? I have good hopes that we can convince the type checker that the singular variant used if and only if `cls == OutputValidator` :stuck_out_tongue:

bin/upgrade.py
```diff
@@ -188,6 +188,23 @@ def upgrade_statement(problem_path: Path, bar: ProgressBar) -> None:
...
+    if (problem_path / "output_validators").is_dir():
+        if (problem_path / "output_validator").exists():
+            bar.error(
+                "can't rename 'output_validators/', 'output_validator/' already exists", resume=True
+            )
+            return
+        content = [*(problem_path / "output_validators").iterdir()]
+        if len(content) == 1 and content[0].is_dir():
+            bar.log(f"renaming 'output_validators/{content[0].name}' to 'output_validator/'")
+            content[0].rename(problem_path / "output_validator")
```
mpsijm, line 201: Note that if we rename the folder by moving it one level up, any symlink to `../../../../skel/validate.h` will fail to work. We may want to recreate this symlink by removing one `../`, using some magic with [`Path.readlink`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.readlink)?

bin/export.py
```diff
@@ -242,6 +242,13 @@ def add_file(path, source):
                     f.unlink()
                     f.write_text(text)
 
+    # TODO: Remove this if we know others use the output_validator dir
+    if (export_dir / "output_validator").exists():
+        (export_dir / "output_validators").mkdir(parents=True)
+        (export_dir / "output_validator").rename(
+            export_dir / "output_validators" / "output_validator"
+        )
```
mpsijm, line 250: The same problem with symlinks shows up during the export again, but the other way around. The advantage here is that we can try to (recursively) copy the contents of the `output_validator` directory, and hopefully, that will copy the actual file instead of the symlink?

bin/problem.py
```diff
@@ -878,13 +877,8 @@ def prepare_run(problem):
         if not testcases:
             return False
 
-        if problem.interactive or problem.multi_pass:
-            validators = problem.validators(validate.OutputValidator)
-            if not validators:
-                return False
```
mpsijm, line 884: If this check is removed, where do we now assert that `output_validator/` exists when the problem is interactive or multi-pass?

mpsijm (changes requested):
It looks kind of weird to still have "lists of output validators" around, but it is simply easier to have `Problem.validators` always return a list, so that's probably fine :stuck_out_tongue: 

Some comments below:

bin/upgrade.py
mzuenni: then we need to go through all files... i think its ok if the user has to fix this?

bin/export.py
mzuenni: oh... did symlinking ever work here since I changed the export code? i could imagine that the symlinks in the tmp dir are already broken?

bin/problem.py
mzuenni: We always need exactly one output validator regardless of the type and that check is still there. (the default validator takes this place if needed)

bin/export.py
mzuenni: I would hape that it here works and also that the rename is ok since its an absolute symlink from `tmp`->`output_validators` and from there on the relative link should be valid, but i am not sure

mzuenni: yes this seems to work here.

bin/upgrade.py
mzuenni: I will see if copy is more useful then rename here ^^

mzuenni: @mpsijm can you try the last commit?

mpsijm: Yes, that works, thanks! :smile:

mpsijm (approved):

PR #436, opened 2025-03-11
Add .download samples and ans_is_output flag

mzuenni:
We still assume that `.ans` files are valid outputs, whereas the new draft spec does not. In BAPCtools, we want to keep as default behaviour to assume that `.ans` files are valid outputs, i.e. `ans_is_output` is `True` by default for "batch" problems. For interactive and/or multi-pass problems, `ans_is_output` is `False` by default and should not be changed to `True`.


mpsijm (commented):
Code review! :smile: Haven't tested it yet, what's the easiest way to do so? Which problems did you use to test `.{statement,download}` files and `ans_is_output` with?

bin/stats.py
```diff
@@ -173,8 +173,6 @@ def count(path):
         def value(x):
             if x[0] == "  time" or x[0] == "subs":
                 return x[1](problem)
-            if x[0] == "A" and (problem.interactive or problem.multi_pass):
-                return None  # Do not show an entry for the answer validator if it is not required
```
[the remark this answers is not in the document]
mzuenni: it is only required by interactive problems. But yea for interactive it should still be hidden.

bin/generate.py
```diff
@@ -964,17 +970,22 @@ def needed(ext):
 
             used_solution = False
             changed_ans = False
-            if problem.interactive or problem.multi_pass:
-                # Generate empty ans file for interactive/multi-pass problems
+            if not problem.settings.ans_is_output:
+                # Generate empty ans file
```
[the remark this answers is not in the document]
mzuenni: no it must not be empty. This if only ensures that we don't run the submission to generate the `.ans` file. Note that the empty `.ans` file is also only created if the generator did not generate a `.ans` file.

bin/problem.py
```diff
@@ -629,67 +643,112 @@ def testcases(
...
+    def statement_samples(p) -> list[Path | tuple[Path, Path]]:
+        in_extensions = [
+            ".in.statement",
+            ".in",
+        ]
+        ans_extensions = [
+            ".ans.statement",
+            ".out",
+            ".ans",
+        ]
```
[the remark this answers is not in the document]
mzuenni: no they are ordered by priority.

bin/export.py
```diff
@@ -158,21 +148,30 @@ def add_file(path, source):
...
-                        if f2.is_file():
-                            out = f2.relative_to(problem.path)
-                            add_file(out, f2)
+                    add_testcase(f)
+
+    samples = problem.download_samples()
+    if len(samples) == 0:
+        util.error("No samples found.")
+    for f, _ in samples:
+        add_testcase(f)
```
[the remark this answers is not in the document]
mpsijm: But they don't need to match, right? Because the contest system would need all three types of samples?

bin/problem.py
```diff
@@ -800,22 +859,22 @@ def validators(
...
                 problem._validators_warn_cache.add(key)
-                match cls, len(validators):
-                    case validate.InputValidator, 0:
-                        warn("No input validators found.")
-                    case validate.AnswerValidator, 0:
-                        warn("No answer validators found")
+                if cls == validate.InputValidator and not validators:
+                    warn("No input validators found.")
+                if cls == validate.AnswerValidator and not validators and not problem.interactive:
+                    # for interactive problems the .ans file should be empty
```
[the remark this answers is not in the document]
mzuenni: for multipass problems it might be useful to have extra data provided to the output checker. 

Thereason why I discourage this for interactive problems is this:
The `.in` file is already only provided to the output validator and not the team, why should there be two files which hold information that only the output validator can read?

bin/problem.py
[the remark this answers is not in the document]
mpsijm: Can we add this as a comment to explain the ordering? :slightly_smiling_face:

bin/problem.py
```diff
@@ -283,6 +283,12 @@ def __init__(
         # BAPCtools extensions:
         self.verified: Optional[str] = parse_optional_setting(yaml_data, "verified", str)
         self.comment: Optional[str] = parse_optional_setting(yaml_data, "comment", str)
+        self.ans_is_output: bool = parse_setting(
+            yaml_data, "ans_is_output", not self.interactive and not self.multi_pass
+        )
```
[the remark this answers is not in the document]
mzuenni: the intenttion here was to say something like ".ans contains output"

bin/export.py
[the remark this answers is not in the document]
mzuenni: no. This new code just properly handles that a testcase without `.in` and `.ans` but with `.in.statement` and `.ans.statement` is still a valid testcase and gets added to the zip

bin/generate.py
```diff
@@ -704,52 +704,55 @@ def validate_in(t, problem: Problem, testcase: Testcase, meta_yaml: dict, bar: P
             )
         return True
 
-    def validate_ans(t, problem: Problem, testcase: Testcase, meta_yaml: dict, bar: ProgressBar):
+    # we assume .ans is a valid output and validate it as such
+    def validate_remaining(
```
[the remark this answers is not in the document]
mzuenni: no opinion on this ^^

test/problems/multipass/answer_validators/validate.ctd
```diff
@@ -0,0 +1 @@
+EOF
```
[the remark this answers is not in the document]
mpsijm: Yep, that was my misunderstanding, thanks :smile: Can we also explain this in the documentation? It currently says "For interactive or multi-pass problems, an empty `.ans` will be generated." For interactive problems, this is always true. But for multi-pass problems, this is only true when the generator does not generate a `.ans` file, right?

bin/problem.py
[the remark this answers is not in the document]
mpsijm: Fair enough! The long form would be, "`.ans` contains output that is accepted by the output validator", right? (we should probably also document this :stuck_out_tongue:)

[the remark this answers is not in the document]
mzuenni: i think i added it to some doc ^^

bin/generate.py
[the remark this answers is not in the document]
mpsijm: Then I think I prefer `validate_ans_and_out` :smile:

bin/problem.py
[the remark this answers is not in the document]
mzuenni: i prefer shorter keys but i dont have a strong opinion

bin/export.py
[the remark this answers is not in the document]
mpsijm: But it currently doesn't work the way I expect it to :upside_down_face: 

When I create a problem that has `1.in`, `1.in.statement`, `1.ans`, `1.ans.statement`, `2.in`, and `2.ans`, then when I run `bt zip` and inspect the zip, I find `2.*` in `data/` and `1.*.statement` in `attachments/` (renamed to `1.{in,ans}`), but the basic `1.{in,ans}` are missing.

I'm fine with copying `1.{in,ans}.statement` to `attachments/1.{in,ans}` (even though that's officially not the job of BAPCtools, but of the contest system), but I think all files matching `1.*` should stay in `data/` in the ZIP, because that's what the spec prescribes. We aren't the ones selecting which "version" of the test case should be in `data/` – all versions should be there, so that a contest system can (1) run the test cases, (2) compile the problem statement, and (3) provide the downloads.

Or am I misinterpreting the spec? :sweat_smile: How I read it, [this section](https://icpc.io/problem-package-format/spec/2023-07-draft.html#samples-shown-in-the-problem-statement) and [this section](https://icpc.io/problem-package-format/spec/2023-07-draft.html#samples-available-for-download) talk about how `.statement` and `.download` files can be used to customize the behaviour of whoever/whatever is using the problem package.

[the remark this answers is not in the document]
mzuenni: are you running the zip command or the sample zip command?

the zip should have all files and no renames
the sample zip should only have `.in` and `.ans` (which is a rename of `.download` or `.sample` if possible)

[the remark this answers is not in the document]
mzuenni: ok, found the issue... `with_suffix` does not behave as expected...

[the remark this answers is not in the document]
mzuenni: @mpsijm can you try again? :)

bin/problem.py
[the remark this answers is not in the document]
mpsijm: Then I guess it's fine like this. This key shouldn't be used very often anyway :stuck_out_tongue:

mzuenni:
anything else to do here?

bin/problem.py
```diff
@@ -629,67 +646,117 @@ def testcases(
...
-        for interaction_path in interaction_paths:
-            testcases.append(interaction_path)
+            # fall back is pair of files
+            testcases.append((name.with_suffix(in_found[0]), name.with_suffix(ans_found[0])))
 
-        testcases.sort()
+        if has_raw and not p.settings.ans_is_output:
+            warn(
+                "It is advised to overwrite .in and .ans for samples if .ans does not represent a valid output"
+            )
```
mpsijm, line 723: What should a problem author do when they see this warning? Does "overwrite" mean that they should add `.*.statement` and/or `.*.download` files? If so, the warning should probably say that, to help people with less experience :slightly_smiling_face: 

Also, I receive this warning when I `bt upgrade` NWERC 2024 Hash Collision and run `bt zip`. But the `.ans` file is empty there, so I shouldn't have to overwrite anything, right? :thinking: 

The samples of this problem have valid `.in` files according to the input validator, but that's not what this warning is about. Although `has_raw` _is_ set to `True` when `.in` exists for a sample. Is that intended?

bin/export.py
```diff
@@ -162,7 +162,9 @@ def add_file(path: Path, source: Path):
                 add_file(out, f)
 
     def add_testcase(in_file: Path):
-        basename = util.drop_suffix(in_file, [".in", ".in.statement", ".interaction"])
+        basename = util.drop_suffix(
+            in_file, [".in", ".in.statement", ".in.download", ".interaction"]
```
mzuenni, line 166: `".in.download"` can never end up here?

bin/problem.py
mzuenni: maybe the warning is nicer now?

bin/export.py
mpsijm: It does:

image

`Problem.download_samples()` prioritizes `.in.download` files, which then are passed to `add_testcase`. We should probably add a test for this as well :stuck_out_tongue:

mzuenni: ah... then `.interaction` can not appear XD

mpsijm: Yes they can, but they're added from a different part of the code :joy: 

image

The fact that sample `.interaction` files are added separately, also confused me the first time I read the code. `Problem.download_samples()` only returns `.in`/`.ans` pairs (no `.interaction` files). But then, in `add_testcase`, the `.in` file is stripped from its suffix and all existing files related to this test case (i.e. that share the same base name) are added, including `.interaction` files.

Funnily enough, I have this old `guess` problem that we use for BAPC test sessions, which was written without `generators.yaml`. `data/sample/` only contains `.interaction` files. This throws a bunch of warnings, and it probably should, but the `.interaction` files still end up in the result of `bt zip` regardless :joy:

Here's what I used to test (zipped manually using `zip -r guess guess`): [guess.zip](https://github.com/user-attachments/files/19362618/guess.zip)

mzuenni: can you undo my last commit? XD

RagnarGrootKoerkamp: > @mpsijm pushed 0 commits

Should have been -1...

mpsijm: I almost started to agree that having only a `.interaction` file would be enough to represent a sample case, but after playing around with this a bit more, I actually disagree :stuck_out_tongue: An `.interaction` file can not end up in the samples zip, so some pair of `.in*`/`.ans*` should always be required, even for samples. (Of course, `.ans` must be empty for interactive problems, which is then ignored in `bt samplezip`.) So, in order to fix the `guess` problem, I had to move `attachment/1.in` to `data/sample/1.in.download` and `touch data/sample/1.ans.download`.

Because a sample test case with only a `.interaction` file is invalid, I've removed that files matching `data/samples/**/*.interaction` were added separately from the `download_samples`. The `testcases` list then only contained `("data/secret/**/*.in", True)`, so I've inlined it. Of course, the `.interaction` files will still end up in the result of `bt zip`, because this extension is in `KNOWN_DATA_EXTENSIONS`.

bin/generate.py
```diff
@@ -436,6 +436,7 @@ def __init__(self, problem, generator_config, key, name: str, yaml, parent, coun
 
         # Whether this testcase is a sample.
         self.sample = len(parent.path.parts) > 0 and parent.path.parts[0] == "sample"
+        self.required_in = [".in", ".in.statement"] if self.sample else [".in"]
```
mpsijm, line 439: Note that `bt generate` currently does not allow generating test cases with only `.{in,ans}.download` and `.interaction`: it requires at least one of `.in` or `.statement` for samples (see also `generate.py:923`). We have some valid minimal combinations of files that we should allow for samples, which is going to be slightly messy to encode, so please check whether you agree before I change it:
- For "normal" problems, either `.in` or `.in.statement` is required for samples
- For interactive problems, either `.in`, or `.in.statement`, or (`.in.download` and `.interaction`) is required for samples
  - The `.in.statement` file could be used for problems with randomized input, which are currently implemented using an interactive validator, but only interact once. This suffices for both `bt samplezip` and `bt pdf`.
  - If there is neither a `.in` nor a `.in.statement` file, then `.in.download` is required for `bt samplezip` and `.interaction` is required for `bt pdf`.
- For `multi-pass` and `interactive multi-pass` problems, I think the same requirements hold as for interactive problems?

mzuenni: yea this is fine for mo

mpsijm: I couldn't think of a simpler way to encode these constraints... :sweat_smile: Samples have quite a bit of logic around them now :joy: Please check if the last commit makes sense :innocent:

mzuenni: i think this can be simplified... give me a moment

mpsijm: Making it one unified disjunction of conjunctions looks slightly better indeed, thanks! :smile:

bin/generate.py
```diff
@@ -1017,6 +1052,17 @@ def needed(ext):
             assert ansfile.is_file(), f"Failed to generate ans file: {ansfile}"
             return True
 
+        def generate_empty_interactive_sample_ans():
```
mzuenni, line 1041: why do we need this? the empty `.ans` file is already generated and is used as fallback anyway?

mzuenni: 1. but (`x.in.statement`, `x.ans`) is still a valid in/ans combination. they don't have to have the exact same extension.
2. that is intended

mpsijm: 1. Sure, we can also let this function generate `.ans` instead :slightly_smiling_face: 
2. But we cannot manually write an empty `.ans` file in `generators.yaml`, because empty hardcoded file contents are forbidden. And even if this would be allowed, I don't want to manually add empty files for any interactive problem where I only have statement- or download-samples :stuck_out_tongue:

mpsijm: Good point, I didn't realize that this function would generate all applicable variants of `.ans*` files :smile: I've rewritten it such that it tries for all variants (in order of priority):
- Whether an answer file already exists, and if so, stop
- Whether a corresponding input file exists, and if so, write an empty answer file and stop

So, for example:
- If `.in`, `.in.statement`, and `.in.download` exist, only an empty `.ans` is generated (if it does not already exist).
- If only `.in.download` exists and `.ans.download` does not, then an empty `.ans.download` is generated.

bin/generate.py
```diff
@@ -1043,10 +1043,13 @@ def generate_empty_interactive_sample_ans():
                 return True
             if not problem.interactive and not problem.multi_pass:
                 return True
-            for ext in ["statement", "download"]:
-                ans_ext_file = infile.with_suffix(f".ans.{ext}")
-                if infile.with_suffix(f".in.{ext}").is_file() and not ans_ext_file.is_file():
+            for ext in ["", ".statement", ".download"]:
```
mzuenni, line 1046: - the empty `.ans` files is not generated yet.
- the empty `.ans` file is explicitly not generated if `solution:` is set, do we want to skip this generation as well?

mzuenni: @mpsijm ping

mpsijm: - Good point, let's move the calling of this function to a later step :slightly_smiling_face: Do you have a multi-pass problem where `.ans` files are not empty, so I can test with that as well?
- Hmmm... that's two conflicting things :sweat_smile: Because if `solution:` is empty for the `sample` directory, but the problem is interactive and therefore requires an empty `.ans` file, I still think that this empty `.ans` file should be automatically generated?

mzuenni: I don't have any problem where .ans is not "the default thingy" XD

but why would you set solution empty? (it is only used to generate the `.interaction` if there is none of `.statement` or `.out`)

mpsijm: Yeah, the value of `solution` doesn't matter when both `.in` and `.interaction` already exist...

The samples are really weird compared to "normal" test cases, because samples work perfectly well without `.in` file. But things like `solution:` only make sense with a proper `.in` file, so if a sample test case doesn't have this, the value of `solution` is not used in any way. But we still want some automation for empty `.ans` files. :upside_down_face: 

I think I'm happy with how the code currently works, is there something you would like to see changed, or do you know a certain configuration that we should reject but currently don't, or a configuration that should work but currently doesn't?

mzuenni: i would still want to skip empty `.ans` generation if the solution is explicitly set empty... However my only reason for this is backwards compatibility.

mzuenni:
The first thing is you describe should be fixed by adding empty `.download` files to the statement samples.
However, having samples which are not in the statement is explicitly not possible because it is bad practice. You can however add `.statement` files which just say `There is more in the zip` or which just display the first few lines followed by `...`.

mpsijm:
The thing is that the testing tool emulates the randomized input. All it needs for that, is an input size. The statement samples are one fixed input file, but they can't be used as input for the testing tool.

Another way to solve this problem then, is to have _no_ samples in the sample zip, and pass the value of _n_ as command-line parameter to the testing tool (so instead of `-f 1.in`, it would be `-n 20`).
However, it's currently not possible to tell BAPCtools (or a contest system) to have no samples at all in the downloadable zip...

I think the "best practice" way to solve this, is to add the statement-samples to the zip anyway, and say "you can use the samples to directly pipe to your submission, or use the testing tool with `-n <int>` to test with a random test case of size _n_". Then the samples and the testing tool are two separate things that can be used to test a submission, but that's fine :slightly_smiling_face: 

If you agree with that, I think we can squash-and-merge this PR. I currently can't think of problem types that we cannot encode, and if we find new problems (both as in "programming problem" and as in "defects in our tool") later, we can always fix it when we encounter them :smile:

mpsijm (approved):

mzuenni:
Yes, in your scenario, samples should probably just not be the input of the testing tool.
IMO the reason for a downloadable sample is that you don't have to type it manually or copy it from the pdf.

Still one last question: why don't we allow to generate empty files?

mpsijm:
- Empty input files: the submission is expected to always read _some_ input (unless we want BAPCtools to support no-input problems, but that's not currently in scope)
- Empty answer files: the output of a submission is not allowed to be empty because this will give the `NO-OUTPUT` verdict in DOMjudge, so the answer file can also not be empty.
  - We could make an exception for this for interactive/multi-pass problems, because then the `.ans` files are only read by the output validator? But I currently don't see a reason to, since we have found a different way to resolve the situation for Evolutionary Excerpt :slightly_smiling_face:

mzuenni:
at least for `.statement` and `.download` files it makes sense to allow empty... and if `.ans` is not an output empty is also fine?

mzuenni:
Also one more point: either this is a valid in/ans than we should be able to generate it, or it is not, then the validator should reject it?

mpsijm:
> at least for `.statement` and `.download` files it makes sense to allow empty... and if `.ans` is not an output empty is also fine?

I can add `and problem.settings.ans_is_output` to the check at `generate.py:938` that prevents writing an empty hardcoded file, but that _does_ allow me to use solution 2 in https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/436#issuecomment-2745241677, and I would say that it is bad practice when the statement shows a sample answer, but the sample answer is not available for download. 

> Also one more point: either this is a valid in/ans than we should be able to generate it, or it is not, then the validator should reject it?

Sure, but that is assuming that the validators actually rejects empty in/ans files. Since 2a53e6f604fdf6db0007ade6ba572dac78d918fc, this is no longer a guarantee. Can we add back the `register("empty")("")` test case for input files, and only when `ans_is_output` also for answer files?

But regardless of the above, I think that explicitly writing empty files is bad practice in general. At the very least, we should warn for this, but I think I'd prefer disallowing it until we find a valid reason to allow them :stuck_out_tongue:

mzuenni:
> Sure, but that is assuming that the validators actually rejects empty in/ans files. Since https://github.com/RagnarGrootKoerkamp/BAPCtools/commit/2a53e6f604fdf6db0007ade6ba572dac78d918fc, this is no longer a guarantee. Can we add back the register("empty")("") test case for input files, and only when ans_is_output also for answer files?

the sanity check warns for empty files

> Can we add back the register("empty")("") test case for input files, and only when ans_is_output also for answer files?

for input that should be easy for `ans_is_output` thats needs some more changes...

> But regardless of the above, I think that explicitly writing empty files is bad practice in general. At the very least, we should warn for this, but I think I'd prefer disallowing it until we find a valid reason to allow them 😛

more or less agreed ^^ for `.download` and `.statement` its very useful and even for `.ans` files it makes since (we are even generating those empty files...) so i would prefer to allow it in general and let the validator complain.

mpsijm:
Ah, you're right, I forgot the sanity check! In that case, the empty generic invalidation test case should not be necessary :slightly_smiling_face: 

> for `.download` and `.statement`, [allowing to generate empty files is] very useful

I'm still not convinced. Do you have a concrete scenario in mind? I'm still in favour of _not_ allowing this, to avoid the "hack" that I would be able to do for Evolutionary Excerpt otherwise :stuck_out_tongue: Additionally, these files are not run through the validators or sanity checks, so we would not catch the empty files later on. 

> I would prefer to allow it in [the generator] and let the validator complain.

Adding more checks earlier never hurts :slightly_smiling_face: In the initial steps of setting up a problem, you typically run the generator without validators, because they do not exist yet in the early phases. Then warning about the "bad practice" of hardcoding empty `.{in,ans}*` files is nice to have in the generator as well, to avoid it as early as possible.

mzuenni:
>>for .download and .statement, [allowing to generate empty files is] very useful
>
>I'm still not convinced. Do you have a concrete scenario in mind?

Not really, but it is somewhat weird that we generate empty `.ans` files but it's impossible to do set this explicitly in the `generator.yaml`.

>>I would prefer to allow it in [the generator] and let the validator complain.
>
>Adding more checks earlier never hurts 🙂 In the initial steps of setting up a problem, you typically run the generator without validators, because they do not exist yet in the early phases. Then warning about the "bad practice" of hardcoding empty .{in,ans}* files is nice to have in the generator as well, to avoid it as early as possible.

The warning would still come during `bt generate` because of the sanity checks. We would just clean up code and not have logic for this at two places. (it is possible to manually generate empty `.ans` files with a generator and that is only covered by the sanity checks which is feels inconsistent...)

mpsijm:
> it is possible to manually generate empty .ans files with a generator and that [...] feels inconsistent

That's also true :upside_down_face: And you're right: I thought the sanity checks were only done in `bt validate`, but the validators are (of course) also run during `bt generate`. Let's just allow writing hardcoded empty files then, even though I'm not happy about it :joy:

mzuenni:
yes can be merged (if we don't have a test that breaks :p)

RagnarGrootKoerkamp:
Thanks guys!

Should we document all/some of the details of the decisions here?
Either by updating the spec, or by adding a doc here since there are many small decisions outside of the scope of the spec?

mzuenni:
not sure what else we would want to document?

mpsijm:
The most important changes have been updated in the docs, and the discussion of the past days was mostly about nitty-gritty implementation details of empty files :stuck_out_tongue: In the end, I don't think we need to document anything extra, the warnings/errors that BAPCtools throws in various places should speak for themselves :slightly_smiling_face:

PR #437, opened 2025-03-12
[problem] Update parsing of problem.yaml based on Kattis/problem-package-format#372

mpsijm:
See Kattis/problem-package-format#372.

Changes:
- **C1**: Keywords are now a list of strings (we used to still parse them as a single string, woops)
- **C2**: No change needed, this was just a "bug" in the human-readable text of the specification
- **C3**: Add constraints to float/int types in `limits` and warnings when any of the values are out of range
- **C5**: Do not allow lists to be empty (if a list-field is optional, it should be either `None` or a non-empty list)

The discussion for **C4** was moved to https://github.com/Kattis/problem-package-format/issues/378 and is pending consensus, and the proposals for **C6**, **Q1**, and **Q2** were dropped.

### DIFF
--- a/bin/problem.py
+++ b/bin/problem.py
@@ -1,3 +1,4 @@
+import datetime
 import re
 import sys
 import threading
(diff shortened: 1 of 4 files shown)


mzuenni (approved):

thorehusfeldt:
If you find it helpful, try to validate against https://github.com/thorehusfeldt/problem-package-schemas/blob/master/yaml/valid/problem.yaml and invalidate against
https://github.com/thorehusfeldt/problem-package-schemas/blob/master/yaml/invalid/invalid-problem.yaml

mpsijm:
Thanks Thore! I've cherry-picked some of the YAML snippets in those files to the tests in BAPCtools, and it caught an error in parsing `ProblemSource` :smile:

Note that I haven't thoroughly tested the combination of `license` and `rights_owner`. Similar to `embargo_until`, BAPCtools doesn't really do much with this information anyway, so the parser there is currently quite lenient, and as such I'll consider it out-of-scope for this PR. If others feel like improving and adding tests for this, feel free to do so :slightly_smiling_face:

PR #441, opened 2025-03-21
Legacy export

mzuenni:


mzuenni:
@mpsijm I guess this can be merged as well?

mpsijm:
Didn't immediately have the time to try it out, but I did now :stuck_out_tongue: 

Is there a reason to _not_ include the input validators? They are part of the spec, and they already were in legacy, now that I took another look. I don't know why they were never included :thinking: 

We could also include the `generators/` folder since it's part of the spec now. The spec says to include a `README.md` to explain how the generators are used, but I guess it's clear enough that it's from BAPCtools (or we can automatically prepend an extra comment in `generators.yaml` that says it's from BAPCtools) :stuck_out_tongue: 

And if we want to be complete, I don't see why we can't also include the answer validators now. They're a BAPCtools-specific extension, but the spec doesn't forbid including extra files :innocent:

mzuenni:
I don't know why... probably because this zip was only ever used as import for domjudge/kattis?

mpsijm:
DOMjudge doesn't run the input validators, but Kattis does. I guess DOMjudge is free to choose this, but it also doesn't hurt to include the validators for archival completeness. 

@RagnarGrootKoerkamp do you know why the input validators are not in the zip currently?

mzuenni:
(for kattis we do export them)

RagnarGrootKoerkamp:
Yeah, what @mzuenni said. It probably was domjudge-first and we simply never bothered/realized it.

mpsijm:
Fair enough! I think we can simply export everything then, no reason to leave anything out (tools that don't know how to use some directories, can simply ignore them) :slightly_smiling_face:

bin/export.py
```diff
@@ -122,9 +127,6 @@ def build_problem_zip(problem: Problem, output: Path):
     if problem.custom_output:
```
mpsijm, line 127: (comment should actually be at lines 124–125, but that part is just outside the diff)

https://icpc.io/problem-package-format/spec/2023-07-draft.html#problem-statements says:

> Filetype can be either `.tex` for LaTeX files, `.md` for Markdown, or .pdf for PDF.

Should we remove `problem.pdf` from the root of the zip, and instead provide `statement/problem.<lang>.pdf`?

Same goes for `solution.<lang>.pdf`.

bin/export.py
```diff
@@ -122,9 +127,6 @@ def build_problem_zip(problem: Problem, output: Path):
...
         files.append(("output_validator/**/*", True))
 
-    if config.args.kattis:
-        files.append(("input_validators/**/*", True))
-
     message("preparing zip file content", "Zip", problem.path, color_type=MessageType.LOG)
 
     # prepare files inside dir
```
mpsijm, line 132: (comment should actually be at line 136, but that's outside the diff)

The comment says:

> `# For Kattis, prepend the problem shortname to all files.`

The spec says at https://icpc.io/problem-package-format/spec/2023-07-draft.html#general-requirements:

> The package must consist of a single directory containing files as described below. [...] the package can be a ZIP-compressed archive of such a directory with identical base name and extension [...] `.zip`.

So I guess we should _always_ prepend the problem short name this now (except for legacy, simply because DOMjudge doesn't support the new format yet)?

bin/export.py
mzuenni: I think I prefer that `bt pdf` (which has to be run from the top level directory) puts the files in the top level directory, but its fine for me to change this

mzuenni: we might want to wait with this until we know how/if domjudge handles multi language support?

mpsijm: Then still, I can imagine we build a kind of export where only the English version ends up in `statement/`, just like how we currently make sure only one language ends up in the zip as `problem.pdf` :slightly_smiling_face: But that's for later – I think that right now, we can strive to be as spec-compliant as possible, so other contest-system-implementors (primarily DOMjudge, but maybe also Kattis) can better test how their systems will handle them :smile:

mpsijm: Turns out that the option to _not_ provide `.<lang>` has been dropped in https://github.com/Kattis/problem-package-format/commit/e42d7cadca0d18d1cda8fd51261a3d9c63d51f64 🙂 So I'd say we only write PDFs with language codes, for the new spec 😄 (I'd even say this if there was the option to drop the language code, because dropping the language code was optional :stuck_out_tongue:)

mzuenni: I will have a look later today

mzuenni: do we want to change the `problem.yaml` to also allow a dict instead of just a single name?

mpsijm: For non-legacy export? Yes, I'd say we only need to reduce to a single name for the legacy export 🙂

mzuenni:
@mpsijm do you have time to take a look?

mpsijm (approved):
Made two small documentation fixes, looking good to me :smile: At least, it works for me on the English-only problems from last year :stuck_out_tongue:

mzuenni:
Some tests here would be good? I am really unsure if this works as we want... mostly because I have no idea what we want since I have never used these features ^^'

@thorehusfeldt do you have multi-language problems/contest to test that nothing breaks and behaves as expected?

mpsijm:
I've added some tests verifying the current behaviour (which I think is the correct behaviour). Basically, the test compares the list of files in the resulting ZIP with some hand-crafted list of files, to see if the contents match :slightly_smiling_face: 

Also, I've renamed `--languages` to `--lang`, as @RagnarGrootKoerkamp suggested in https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/442#discussion_r2020104074.

If you (and the CI) agree with the tests that I added, this can be squashed-and-merged :smile:

PR #442, opened 2025-03-29
Absolute2

mzuenni:


mzuenni:
@RagnarGrootKoerkamp I closed the other PR since it was really outdated... this should use absolute where it is ok

doc/multiple_languages.md
```diff
@@ -108,7 +108,7 @@ creates PDFs for every problem language statement `problem.xy.tex`.
 With arguments,
 
 ```sh
-bt pdf --language en --language fr
+bt pdf --languages en fr
```
RagnarGrootKoerkamp, line 111: not sure changing `--language` to `--languages` makes that much sense when probably the usual case is to pass 1 lang only, but OK.
Could also rename to `--lang` maybe?

RagnarGrootKoerkamp (approved):
oh I guess this is not a stacked diff but a diff containing all the other PRs into draft as well.

anyway lgtm i gues

RagnarGrootKoerkamp:
What was the original reason for preferring absolute paths again exactly?

#356 mentions that `.absolute()` avoids resolving symlinks, but why was that important again?
(I guess if we do `x.resolve()/../blah`, then it's better to not acidentally go through the symlink?

doc/multiple_languages.md
mpsijm: That comment actually belongs to #441 :see_no_evil: Looks like the `absolute2` branch was based onto the `legacy-export` branch. But yeah, I'm fine with `--lang` (I'll change it there) :slightly_smiling_face:

mpsijm:
I'm not sure, actually :sweat_smile: I think it's difficult to test the implications of this, I don't have a mental list of all the ways how symlinks are used in BAPCtools. Also, we may want to check if log/warn/error messages don't get blown up with full absolute paths now, in cases where shorter relative paths would suffice.

Also, I don't think this really belongs to the `draft` PR. That said, it's also currently not a good idea to re-do this on `master`, because in `draft`, we have been changing lots of code that uses these paths. :upside_down_face:

RagnarGrootKoerkamp:
Most error reporting already makes paths relative to elsewhere.

Thinking of it, that may be the reason why this is a good change: given two absolute paths it's much more reliable to the relative version from one to the other.
Especially with `.resolve()`, it can be that one of the paths had a symlink and the inferred relative path is not what we'd want it to be.

After this change, we can probably have a problem directory that's a symlink to another place and it should behave more consistently as if the symlink wasn't there at all?

mzuenni:
I think the only reason why we want to change this is if the problem itself is symlinked somewhere else? Then `resolve` would resolve the path to the problem which is weird from a user perspective?

mpsijm:
Can we:
- For the remaining instances of `.resolve()`, add a comment why `.absolute()` is not used there (or simply change them to `.absolute()`, if possible)?
- Add a pre-commit hook that prevents the use of `.resolve()` in the future, with @mzuenni's explanation as message when the hook fails? Can be as simple as a `! grep "\.resolve()" **/*.py`. Of course, we should have some way to suppress the warning with an explanatory comment.

RagnarGrootKoerkamp:
May be a bit over the top to have a precommit hook, but that stuff is fast anyway so why not :shrug:

mpsijm:
I just think it would be a shame if this decision gets lost/forgotten in the future, when nothing is preventing the use of `.resolve()` :stuck_out_tongue: And if the check is automatic, we don't have to pay attention to it during reviews :smile:

mzuenni:
Not sure how to add such a hook... and how to handle those three cases where `resolve` is still used. ^^'
- 2 of them "read the link" in that case `resolve` should be used
- the third checks if something is in the `tmp_dir`, since we never build links to that dir `absolute` would be fine but IMO semantically checking this with resolve makes more sense

PR #444, opened 2025-03-30
Use Validator.source_dir at all places

mzuenni:


bin/upgrade.py
```diff
@@ -198,11 +202,15 @@ def move(src: str, dst: str) -> None:
                 else:
                     Path(src).rename(dst)
 
-            shutil.copytree(content[0], problem_path / "output_validator", copy_function=move)
+            shutil.copytree(
+                content[0], problem_path / validate.OutputValidator.source_dir, copy_function=move
+            )
             shutil.rmtree(problem_path / "output_validators")
         else:
             bar.log("renaming 'output_validators/' to 'output_validator/'")
```
mpsijm, line 210: There's another `output_validator/` in this log message :slightly_smiling_face:

mpsijm (commented):
I don't think this is strictly necessary, but sure, why not :stuck_out_tongue: 

Besides the comment I pointed out below, there are also some usages in globs in `export.py` when building the ZIP. But those lists will become a bit ugly when changed to point to `*Validator.source_dir` :upside_down_face:

mzuenni:
replaces them as well, i think it looks fine?

Issue #181, opened 2021-11-24
`--order-from-ccs` should write the order into `contest.yaml` for reuse with `--order`


RagnarGrootKoerkamp:

mpsijm:
`contest.yaml` does not have an `order` field currently, right? `.bapctools.yaml` does, but I'm not sure if we'd want to write to that :thinking:

RagnarGrootKoerkamp:
Hrmm yes. I think I'm still in favour of 'merging' these two, but we can also cache it in `.bapctools.yaml` anyway, although I think the order should be preserved in the repo.

mzuenni:
this is now in the Draft PR

PR #446, opened 2025-04-01
Order in contest yaml

mzuenni:
#181

### DIFF
--- a/bin/tools.py
+++ b/bin/tools.py
@@ -219,7 +219,7 @@ def get_problems():
             problems.sort(key=lambda p: (-solves[p.name], p.label))
             verbose(f"order: {', '.join(map(lambda p: str(p.label), problems))}")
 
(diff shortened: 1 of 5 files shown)


RagnarGrootKoerkamp (approved):

bin/tools.py
```diff
@@ -1370,8 +1380,8 @@ def test(args):
     original_directory = Path().cwd()
     config.n_warn = 0
     config.n_error = 0
-    contest._contest_yaml = None
-    contest._problems_yaml = None
+    _contest_yaml = None
+    _problems_yaml = None
```
RagnarGrootKoerkamp, line 1384: CI error might be here? if this is a global variable now

mpsijm: Most likely yes, because this only assigns `None` to the local variables in the `tools` module, rather than the (exported) variables in the `contest` module.

PR #447, opened 2025-04-04
Add secondary sort-keys to `--order-from-ccs`

### LINKED ISSUE #377: `--order-from-ccs` orders on number solved, disregarding tiebreaker number submitted
We got the slides as:
A: tried(20), solved(8)
B: tried(10), solved(8)

Where I would say A should have been regarded harder and the order should have been B,A instead of A,B.

mzuenni:
Closes #377

### DIFF
--- a/bin/tools.py
+++ b/bin/tools.py
@@ -211,13 +211,13 @@ def get_problems():
                         self.pending += team_stats["num_pending"]
                         self.teams_pending += 1
 
-                def key(self, fall_back: str) -> tuple[int, float, str]:
+                def key(self) -> tuple[int, int]:
                     # self.solved more AC => easier
                     # possible tie breakers:
                     # self.submissions more needed to get the same number of AC => Harder
                     # self.teams_pending more teams tried => appeared easier
                     # TODO: consider more stats?
-                    return (-self.solved, self.submissions, fall_back)
+                    return (-self.solved, self.submissions)
 
             # Get active contest.
             cid = get_contest_id()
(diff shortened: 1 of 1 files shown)


bin/tools.py
```diff
@@ -193,31 +193,49 @@ def get_pos(id):
...
+                        self.teams_submitted += 1
+                    if team_stats["num_pending"]:
+                        self.pending += team_stats["num_pending"]
+                        self.teams_pending += 1
+
+                def key(self, fall_back: str) -> tuple[int, float, str]:
+                    # self.solved more AC => easier
+                    # possible tie breakers:
+                    # self.submissions more needed to get the same number of AC => Harder
+                    # self.teams_pending more teams tried => appeared easier
```
mpsijm, line 218: I don't see `team_pending` as part of the returned `key`?

bin/tools.py
```diff
@@ -193,31 +193,49 @@ def get_pos(id):
...
-            # Convert away from defaultdict, so any non matching keys below raise an error.
-            solves = dict(solves)
-            verbose("solves: " + str(solves))
+                for team_stats in team["problems"]:
+                    problem_stats[team_stats["problem_id"]].update(team_stats)
 
             # Sort the problems
-            # Use negative solves instead of reversed, to preserver stable order.
-            problems.sort(key=lambda p: (-solves[p.name], p.label))
+            problems.sort(key=lambda p: problem_stats[p.name].key(cast(str, p.label)))
```
mpsijm, line 238: The label is a fall-back so is always returned last, so we don't need to pass it as parameter to `key`:
```suggestion
            problems.sort(key=lambda p: (problem_stats[p.name].key(), p.label))
```

bin/tools.py
mzuenni: it is not used right now ^^ I think it is unlikely that `team_pending` or `submissions`  is equal. If we want to use both we most likely want to combine them in some way?

PR #448, opened 2025-04-05
Draft visualizer

mzuenni:


mzuenni:
Some Notes:
my implementation uses a different `InputVisualizer` invocation. The visualizer is called with the `input` and `answer` files instead of only reading the `input` `stdin`. On one hand this is closer to the old behaviour BAPCtools always used on the other hand the `answer` file might be necessary for the visualization especially if its not a valid output. (The name InputVisualizer might be confusing...)

thorehusfeldt:
Thank you for doing this.

We currently have
 
 ```cue
 	visualizer?:  #command & =~"^/" & !~"\\{count" | null
 ```

Should `generators.yaml` actually support  something like this instead?:

```cue
visualizer?: ["in" | "out"]: #command | null
```

In particular the requirement that visualizer invocations start with `/` is now void, because the directory structures is fixed. So the path should instead be relative to `input_visualizer` and `output_visualizer`. 

But it’s not clear to me that this shall be specified in generators at all any more. When `output_visualizer` exists, it will be run, and with a narrowly specified invocation. The same may be true of `input_visualizer`. So what’s left to specifify? (Honest question.) Sometimes may want to switch visualization off, so there’s case to be made for  

```cue
visualizer?: ["in" | "out"]: *true | false
```
I can also see myself passing _arguments_ to a visualizer, so maybe we want
```cue
visualizer?: ["in" | "out"]: string | false
```
instead?

mzuenni:
> Should generators.yaml actually support something like this instead?:
> [...]
> So what’s left to specifify?

I don't think we need to specify anything in `generators.yaml` if you provide an `input_visualizer` you already show that you want the visualization. And if you don't want to run it temporarily, there is `bt generate --no-visualizer`.

If you don't want to run it for some testgroup/test cases you can implemtent this via `<input, output>_visualizer_args` keys (not that in my implementation visualizers are not called with validator args). We could in theory also add to the spec that `_visualizer_args` is `str | list[str] | False` where False indicates that it should not be run... But i don't know if we need that.

thorehusfeldt:
Even more basic question: what does `data/sample/01.pdf` mean? (Is it the result of running the input visualiser on `01.in`?)

mpsijm:
If we'd strictly follow the spec, the resulting images of the input/output visualizer would not automatically end up in `data/` at all; this is a choice that we make in BAPCtools. Also, according to the spec, the image in `data/` is an "Illustration of the _test case_" (whether that's only input, or both input and answer, is the choice of the problem setter). I already thought about this in https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/438#issuecomment-2765489957, and in summary, I think we should do something like this:
- If the input visualizer produces an image, we place that in `data/`.
- Else, if the output visualizer produces an image with the canonical jury submission (i.e. the one from which we generate `.ans` files), we place that in `data/`.
- If both produce an image, the result of the output visualizer only stays in `feedback_dir` (this probably already happens for the non-canonical jury submissions anyway).

So I don't think we need to disambiguate `1.{in,ans}.png`, unless you think we need multiple images per test case in some cases (currently, the spec only allows one image file per test case, which I think is fine).

mzuenni:
> Even more basic question: what does `data/sample/01.pdf` mean? (Is it the result of running the input visualizer on `01.in`?)

Short answer yes.
More detailed answer: Its the visualization of the testcase (what exactly that means is up to you/the problem). In terms of `bt generate` its the output of the `input_visualizer` but if no `input_visualizer` exists we try to use the `output_visualizer` as a fallback to generate an visualization of the testcase.

> Do we really need data/sample/01.in.pdf now?
I don't see the use for this

mzuenni:
@mpsijm will you have time to take a look?

bin/run.py
```diff
@@ -420,14 +439,16 @@ def process_run(run: Run):
...
-                if f in [judgemessage, judgeerror]:
-                    continue
                 if f.name.startswith("."):
                     continue  # skip "hidden" files
+                if f.name in ["judgemessage.txt", "judgeerror.txt"]:
+                    continue
+                if f.name.startswith("judgeimage.") or f.name.startswith("teamimage."):
+                    data += f"{f.name}: {shorten_path(self.problem, f.parent) / f.name}\n"
+                    ensure_symlink(run.problem.path / f.name, f, output=True, relative=False)
+                    continue
```
mpsijm, line 451: This only works for the output _validator_, right? What if I want to run the output _visualizer_ using `bt test`?

bin/generate.py
```diff
@@ -1049,24 +1015,96 @@ def generate_visualization():
...
+                        file = teamimage.with_suffix(ext)
+                        if file.is_file():
+                            found = file
+                    for ext in config.KNOWN_VISUALIZER_EXTENSIONS:
+                        file = judgeimage.with_suffix(ext)
+                        if file.is_file():
+                            found = file
+                    if found is not None:
+                        found.rename(in_path.with_suffix(found.suffix))
+                        bar.log(f"Using {found.name} as testcase visualization")
```
mpsijm, line 1087: Maybe we can specify that this image is from the output visualizer? Maybe it's not needed though, because if the input visualizer does not exist, there's really only one place where this can come from.

```suggestion
                        bar.log(f"Using {found.name} from output_visualizer as test case visualization")
```

bin/generate.py
mpsijm, line 1: In https://icpc.io/problem-package-format/spec/2023-07-draft.html#reporting-additional-feedback, it is mentioned that the output validator can also write image files to the feedback directory. Currently, `bt generate` only runs the answer validator, but I guess that we may also want to run the output validator on the `.ans` file, if an output validator exists? I think this could be added to `TestcaseRule.validate_ans_and_out`?

mpsijm (commented):
Thank you so much for picking this up! :heart: I see a lot has changed, and the majority is looking very good :smile: Some nit-pick comments or questions below :slightly_smiling_face: 

I didn't test super extensively, but I'm very happy that @thorehusfeldt also took the time to write some visualizers! :heart:

bin/generate.py
```diff
@@ -121,7 +121,9 @@ def __init__(self, problem: Problem, string: str, *, allow_absolute: bool, allow
             raise ParseException("{seed(:[0-9]+)} may appear at most once.")
 
         # Automatically set self.program when that program has been built.
-        self.program: Optional[program.Generator | program.Visualizer | run.Submission] = None
+        self.program: Optional[program.Generator | visualize.InputVisualizer | run.Submission] = (
```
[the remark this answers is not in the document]
mzuenni: I missed this, but the actual type is just `Generator | Submission` since a visualizer does no longer appear in the `generators.yaml`

bin/generate.py
```diff
@@ -1049,24 +1015,96 @@ def generate_visualization():
...
+            in_path = cwd / "testcase.in"
+            ans_path = cwd / "testcase.ans"
+            out_path = cwd / "testcase.out"
+            assert in_path.is_file()
+            assert ans_path.is_file()
+
+            visualizer: Optional[visualize.AnyVisualizer] = problem.visualizer(
+                visualize.InputVisualizer
+            )
+            output_visualizer = problem.visualizer(visualize.OutputVisualizer)
```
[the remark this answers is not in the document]
mzuenni: the type is needed here because under some conditions we do `visualizer = output_visualizer` which yields a type error

bin/run.py
mzuenni: since the visualizer can read the output of the validator it should probably only ever run after it... so no visualization for `bt test`

bin/generate.py
mzuenni: this already happens right?
if there is a `.out` we run the output validator and if there `ans_is_output` is true the output validator is used as answer validator and therefore also run.

skel/problem/input_visualizer/readme.md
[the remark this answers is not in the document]
mzuenni: added an example

bin/run.py
mpsijm: Those two point don't seem mutually exclusive to me, though :stuck_out_tongue: Why can we not also run the output visualizer after running the output validator here?

bin/generate.py
mpsijm: Sure, but I'm talking about the case where a `.out` file does _not_ exist, which is the case for all secret test cases.

Take for example Jib Job. Say that, as implementor of that problem, I want to visualize how the output of submissions looks like, and use this as test case image as well. This problem could have an `output_visualizer` like this:

<details>
<summary>output_visualizer</summary>

```py
#!/usr/bin/env python3
import sys

f = open(sys.argv[1]).readlines()
n = int(f[0])
team_jibs = list(map(int, sys.stdin.read().split()))
assert len(team_jibs) == n

xl, xh, yl, yh = 20000, -10000, 20000, -10000

cranes = []
for line, r in zip(f[1:], team_jibs):
    x, y, h = map(int, line.split())
    cranes.append((x, y, r))

    xl = min(xl, x - r)
    yl = min(yl, y - r)
    xh = max(xh, x + r)
    yh = max(yh, y + r)

w = xh - xl
h = yh - yl
if w > h:
    h *= 1024 / w
    w = 1024
else:
    w *= 1024 / h
    h = 1024

with open(sys.argv[3] + "/judgeimage.svg", "w") as of:
    of.write(
        f'<svg xmlns="http://www.w3.org/2000/svg" viewBox="{xl} {yl} {xh - xl} {yh - yl}" width="{w}" height="{h}" style="fill:rgba(0,0,255,0.5)">'
    )
    for x, y, r in cranes:
        of.write(f'<circle cx="{x}" cy="{y}" r="{r}" />')
    of.write("</svg>")
```

</details>

I don't need an `input_visualizer` (or "test case visualizer"), because it will do exactly the same as the output visualizer, with the difference that it would read from a fixed `.ans` file, rather than read a submission's output from stdin. 

Now, I could also decide to write similar code like this in the output _validator_ instead, rather than having a separate output _visualizer_. But, currently, the output validator is not run during `bt generate` for this problem.

(to test this, I currently just appended this to the `main` function of the output validator of Jib Job, because I didn't feel like actually writing the same visualizer in C++ :stuck_out_tongue:)
```c++
    ofstream vis("judgeimage.svg");
    vis << n << endl;
    vis.close();
```

mzuenni: in the case where `.out` does not exist (and `ans_is_out`) the output Validator should already be run (as well as any other answer validator)?

we just don't copy the image... at least thats how i understand the code right now

mzuenni: mhm this is kind of ugly for hashing... we now have multiple programs that could create the same file... and we don't know which is actually responsible ^^' This would mean that if either validator or visualizer change both need to be rerun?

bin/run.py
mpsijm: Oohh woops, I confused `bt test` with `bt run` :joy: I see now that `bt run <submission> <test_case>` works as I'd expect (as in, it copies the resulting image from `<feedback_dir>` to the problem's working directory), thanks :smile:

bin/generate.py
mpsijm: > we just don't copy the image

Check, that may be it, then!

> we now have multiple programs that could create the same file... and we don't know which is actually responsible ^^'

Oh, that's a good point. :sweat_smile: We can decide to simply not implement this in BAPCtools, but I would also be okay with rerunning both "output processors" when either of the two changes.

Another option could be: if `output_visualizer` exists, then image files written by `output_validator` are ignored, and if `output_visualizer` does not exist, treat the images of the `output_validator` as if they are generated by the visualizer? :thinking:

mzuenni: > Another option could be: if output_visualizer exists, then image files written by output_validator are ignored, and if output_visualizer does not exist, treat the images of the output_validator as if they are generated by the visualizer? 🤔

the problem is more like this:
-  have an output_visualizer => `teamimage.png` is written
- delete that output_visualizer
now we don't know how `teamimage.png` was generated so we don't know if we need to delete it or not...

mpsijm: Also a good point :thinking: 

Just to be clear: for the hashing story, we're only talking about `bt generate`, which runs the canonical jury submission on some `.in` file, and the output validator/visualizer on the jury submission, right? Because in `bt run`, the feedback directories should be cleaned before starting a new run.

Then for `bt generate`, we have multiple things to take into account. Not only the output validator/visualizer can change, but also some combination of `--no-validators` and `--no-visualizer` could be passed, in which the image files should not be removed from `data/` even though they don't exist in the feedback directory.

You'd almost want some kind of snapshot of the feedback directory after running the output validator, that can then be duplicated before running the visualizer? However, simply duplicating the entire feedback directory is a waste of RAM/space, and symlinking the contents of the feedback directories can cause nasty bugs, for example when the output visualizer tries to overwrite/append to a file that the output validator left behind...

mpsijm: Oh wait, my browser didn't refresh the thread, and I only now see:

> i think its fine if generate only generates an image from a visualizer and not from a valdiator... that we are also considering the output validator is already pushing it ^^

I tend to agree with this as well :stuck_out_tongue: We can always leave document (e.g. by opening an issue in this repo) that this is a part of the spec that we do not implement, for the reasoning discussed here :slightly_smiling_face:

mzuenni: > You'd almost want some kind of snapshot of the feedback directory after running the output validator, that can then be duplicated before running the visualizer? However, simply duplicating the entire feedback directory is a waste of RAM/space, and symlinking the contents of the feedback directories can cause nasty bugs, for example when the output visualizer tries to overwrite/append to a file that the output validator left behind...

yes a snapshot after each step would be great...

> hat this is a part of the spec that we do not implement

`bt generate` is not part of the spec at all so we can do what we want.

mpsijm: > we can do what we want.

Oh yeah good point! That stays confusing. :joy: Anyway, it is something that you'd expect `bt generate` to manage. An output validator that writes images will work just fine in DOMjudge, in the end :slightly_smiling_face: Then the fact that the images do not end up in `data/`, can still be seen as a "flaw" in `bt generate` :stuck_out_tongue:

mzuenni: I think that confusion can only arises because we run the output_validator? else it would be pretty clear that the image always comes from the input_visualizer

mpsijm: Yes, that is also true :slightly_smiling_face: Then still, I think this scenario is not too uncommon (e.g. the Jib Job example that I gave in https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/448#discussion_r2044168045). Again, it's fine if we don't immediately implement this here because it adds complexity, and it can easily be avoided by simply not generating images in the output validator. But it's slightly counter-intuitive so I still think it would be nice to document it somewhere :slightly_smiling_face:

mzuenni: I think even the current implementation of running the `output_visualizer` is broken in regards to caching...
The output visualizer can read stuff written by the `output_validator` but the curren't implementation can't guarantee that these files still exist...

bin/generate.py
```diff
@@ -709,18 +707,22 @@ def validate_ans_and_out(
...
+        output_validator_hashes = testcase.validator_hashes(validate.OutputValidator, bar)
 
         mode = validate.Mode.ANSWER
-        if testcase.root in config.INVALID_CASE_DIRECTORIES:
+        if testcase.root == ["invalid_answer"]:
             mode = validate.Mode.INVALID
-        elif testcase.root == "valid_output":
+        elif testcase.root in ["valid_output", "invalid_output"]:
+            ans_out_validator_hashes.update(output_validator_hashes)
             mode = validate.Mode.VALID_OUTPUT
```
mpsijm, line 718: I think this change broke the CI? The error I'm seeing, is:
```
Generate: invalid_output/wrong_answer               default_output_validator.cpp: rejected
default_output_validator.cpp: Got: 1, wanted: 0
-> /tmp/bapctools_5d438c/identity/links/8885fb/testcase.out
```
The entry for `invalid_output/wrong_answer` in `generators.yaml` is:
```
  invalid_output:
    data:
      "wrong_answer":
        in: "0"
        ans: "0"
        out: "1"
```
So, it should be correct that the answer is wrong. However, the test case (with root `invalid_output`) now receives `Mode.VALID_OUTPUT`, but that should be `Mode.INVALID`, right?

bin/generate.py
mzuenni: there is also the case where testcases are symlinked because of `input:` in that case some of the validators are rerun ^^

> But it makes sense that the output visualizer visualizes a possible output that represents a solution, right?

not sure about this ^^' it probably depends on the kind of problem. for guess for example i think its quite strange to visualize a binary search?

mpsijm: Yeah, I'm fine with not (fully) supporting visualizers for interactive test cases for now. Same as for the question of an output validator producing image files, we may want to document this somewhere :slightly_smiling_face:

mpsijm:
Updated the spec in https://github.com/Kattis/problem-package-format/pull/439 (and hopefully also correctly updated BAPCtools accordingly, I'll check tomorrow if the tests passed) :smile:

PR #452, opened 2025-05-08
[skel/problem_cfp] move spoilers into a separate `.md` file

paul-wild:
For purposes of a Call for Problems it's nice to have all the information about the intended solution of the problem in a separate file (i.e. not in the `README.md`). The same goes for the author's difficulty estimate.

Having this information outside of the `README.md` also means that it does not get automatically shown to you if you browse the contest problems in the GitHub/GitLab web interface.


RagnarGrootKoerkamp:
Could you move the difficulty on top, so it's stills somewhat easier to read that without spoiling the rest?

PR #455, opened 2025-07-29
Extend `bt fuzz` / update signal handling

mzuenni:


mpsijm (approved):
That indeed fixes it, thanks! :smile:

PR #456, opened 2025-08-13
Rename testdata.yaml to test_group.yaml and implement Test Case Configuration

mpsijm:
This PR renames `testdata.yaml` to `test_group.yaml` (as defined [here](https://icpc.io/problem-package-format/spec/2023-07-draft.html#test-data-configuration)) and introduces Test Case Configuration .yaml files (as defined [here](https://icpc.io/problem-package-format/spec/2023-07-draft.html#test-case-configuration)).
Additionally, `.hint` and `.desc` files are no longer used: they should now be keys (`hint:` and `description:`, respectively) in the Test Case Configuration file.

A lot has changed, and I think everything still works, but please verify :innocent: `bt upgrade` should also properly take care of both migrations, both in `data/` and in `generators.yaml`.

A special request for @thorehusfeldt if you have the time (if not, also fine :slightly_smiling_face:): can you double-check the schemas for `test_group.yaml` and `<testcase>.yaml` with the latest version of the spec? :smile:


bin/problem.py
```diff
@@ -457,105 +459,102 @@ def _read_settings(self):
...
-                continue
-            with p._testdata_lock:
-                if f not in p._testdata_yamls:
-                    raw = substitute(
-                        f.read_text(),
-                        p.settings.constants,
-                        pattern=config.CONSTANT_SUBSTITUTE_REGEX,
-                    )
-                    p._testdata_yamls[f] = flags = parse_yaml(raw, path=f, plain=True)
+            if f.suffix != ".yaml":
```
mzuenni, line 470: maybe checking if `f` is a directory is nicer?

bin/upgrade.py
```diff
@@ -116,14 +119,50 @@ def rename_testcase(old_base: Path, new_dir: Path) -> None:
     if bad_dir.is_dir() and not any(bad_dir.iterdir()):
         bad_dir.rmdir()
 
+    # Move .hint and .desc files to the Test Case Configuration .yaml file
+
+    test_case_yamls = defaultdict[Path, CommentedMap](CommentedMap)
+    for f in (problem_path / "data").rglob("*.yaml"):
+        if f.with_suffix(".in").exists():  # Prevent reading test_group.yaml, which has no *.in file
```
mzuenni, line 126: does the `generato.yaml` prohibit the unnumbered test case name `test_group`?

bin/upgrade.py
```diff
@@ -116,14 +119,50 @@ def rename_testcase(old_base: Path, new_dir: Path) -> None:
...
+        else:
+            bar.log(f"moving '{f}' to 'hint' key in '*.yaml'")
+            test_case_yaml["hint"] = f.read_text()
+            write_yaml(test_case_yaml, f.with_suffix(".yaml"))
+            f.unlink()
+
+
+def rename_testdata_to_test_group_yaml(problem_path: Path, bar: ProgressBar) -> None:
+    for f in (problem_path / "data").rglob("testdata.yaml"):
+        new_name = f.with_name("test_group.yaml")
```
mzuenni, line 152: error if `new_name` already exists. (For a proper problem in legacy format this won't happen but problems might be a weird mix of legacy/draft...)

mpsijm:
Thanks for the comments, I fixed them :smile:

bin/upgrade.py
mzuenni: One more remark, do we also want to check this for not generated testcases and if yes where?
- during `bt validate`?
- in `problem.testcases()` ? (would probably need caching to not get the same message multiple times?)

bin/generate.py
```diff
@@ -1678,7 +1678,13 @@ def parse(key: str, name_gen: Callable[[], str], yaml: dict, parent: AnyDirector
...
             if is_testcase(yaml):
                 if isinstance(parent, RootDirectory):
-                    raise ParseException("Testcase must be inside Directory", name)
+                    raise ParseException("Test case must be inside a Directory.", name)
+
+                if not parent.numbered and key == "test_group":
+                    raise ParseException(
+                        "Test case must not be named 'test_group', this clashes with the group-level 'test_group.yaml'.",
+                        parent.path / name,
+                    )
```
mzuenni, line 1688: Can you move this into the `TestcaseRule`? I think the canonical place would be after checking that the name matches the `FILE_NAME_REGEX`.

bin/upgrade.py
mpsijm: Good call :smile: I put the checks in `problem.testcases()`, because I think this should always be checked, and it feels like the place to do it: we also check for `.in`/`.ans` consistency there.

Regarding caching: Placing the checks in `Problem.validate_data` would also show the warning multiple times, since this function is called once for each `mode`. I think it's fine if the warning is shown multiple times? It should not occur that often anyway, and I don't immediately see a clean way of implementing caching without writing things to the tmpdir (we could store the warnings in some global variable, but that feels ugly) :stuck_out_tongue:

mzuenni: we already have `problem._testcases` cache we could add something like `problem._warned_test_group` (or some nicer name) and just warn for a testcase if it is not yet in that set.

I thinks its fine to warn for `sample/test_group.in` and `secret/test_group.in` individually, but its a bit weird if we warn for the exact same file during input, answer and output validation and run and...

mpsijm: Ah thanks, that's already better than a global cache :smile: I think I've made it look somewhat clean, what do you think? :slightly_smiling_face:

mpsijm:
Okay, I've finally found some time to verify that the updated `generators_yaml_schema.json` works properly in my editor :smile: Turns out it was as simple as adding this line at the top of some `generators.yaml` file (only works if your editor actually uses `yaml-language-server` under the hood):
```yaml
# yaml-language-server: $schema=file:///home/maarten/git/contest/BAPCtools/support/schemas/generators_yaml_schema.json
```

I've even found and fixed a little bug that has been in the schema for a while already :stuck_out_tongue: 

Regarding the Cue schemata: we have a CI test script (`test/yaml/generators/test_schemata.sh`) that runs `cue vet` on a bunch of `generators.yaml` files, so I assume that'll be all right :smile:

PR #457, opened 2025-08-18
[doc] Add important note to readme that the latest version only supports 2023-07-draft

mpsijm:
To be merged right before we switch the main branch, but reviews are already welcome :slightly_smiling_face:


mpsijm:
Rebased and renamed, now depends on #461

PR #458, opened 2025-08-27
Rename Test Case Visualizer back to Input Visualizer

mpsijm:
See https://github.com/Kattis/problem-package-format/pull/439#discussion_r2299228708.


mzuenni:
looks good to me

PR #464, opened 2025-09-27
validation.h: stold => std::from_chars

RagnarGrootKoerkamp:
Followup of #224 but into `main`


mzuenni:
Even though from_chars for doubles is part of c++17, compiler support still does not look great, see [compiler support](https://en.cppreference.com/w/cpp/compiler_support/17#C.2B.2B17_library_features)

i guess we still want to mark this as draft for now?

PR #467, opened 2025-09-27
[tools] Separate changing to contest directory and getting problems list

### LINKED ISSUE #466: Better separate parsing of command-line arguments and changing directory
See https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/166 for previous discussions on this topic. We still think it would be nice to clean this up, but the work should probably be re-done because that PR was way too outdated.

mzuenni:
Will eventually close #466, which was a follow-up of #166

### DIFF
--- a/bin/stats.py
+++ b/bin/stats.py
@@ -380,7 +380,7 @@ def more_stats(problems: list[Problem]) -> None:
         values = []
         for problem in problems:
             directory = (
-                Path().cwd() / "submissions" / problem.name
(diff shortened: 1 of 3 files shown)


mzuenni:
@mpsijm is this what you had in mind? ^^'

mpsijm (commented):
This is definitively cleaner than how it was before, thanks! ❤️ The rewrite of `config.args` in #473 also helps 😄

PR #473, opened 2025-10-20
add more types

mzuenni:
#102

### DIFF
--- a/bin/testcase.py
+++ b/bin/testcase.py
@@ -414,10 +414,8 @@ class Testcase:
                 issues.append("All validators accepted.")
             elif ExecStatus.REJECTED not in results:
                 issues.append(f"At least one validator must exit with {config.RTV_WA}.")
-            else:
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -3,12 +3,15 @@
...
 
-[tool.ruff.lint]
-# Ignore star-imports
-# TODO #102: This is definitely fixable, but for now, postponing the clean-up of all imports
-ignore = ["F403", "F405"]
-
+[tool.ruff.lint.isort]
+# organize imports in two blocks, non local and local
+section-order = ["standard-library", "first-party"]
+default-section = "standard-library"
```
mpsijm: @mzuenni Do you have a strong opinion against the default ordering (`["standard-library", "third-party", "first-party"]`)? I think I would prefer the default ordering 🙂

mpsijm: Fixed in 45c7f0421bde9ba5394234eced854dc8889cb78e :smile:

PR #474, opened 2025-11-07
Make bapctools pip-installable

### LINKED ISSUE #136: Consider renaming `bin` to `lib`
Historically `tools.py` was the only and main binary, but by now most files in this directory are just python libraries.
It would be more consistent with other projects to move all python source code to `lib` and make a small bash wrapper script `bin/bt` that can be called from anywhere and invokes `tools.py`.

niklasmohrin:
Closes #136

To test out the installation from the branch:
```bash
pip install 'git+https://github.com/niklasmohrin/BAPCtools.git@module'
bt --help
```

Tasks:
- [x] fix tests
- [x] group imports
- [x] set correct platform requirements for wheel (only Linux, because of checktestdata) edit: we just pretend that it is universal
- [ ] check that everything still works
- [x] add compatibility script in `bin/tools.py` (optionally, make it warn about deprecation)

Things to double-check in review:
- Does shell auto completion still work with the previous setup when just upgrading via git pull?

What this PR does not do, but what can be done next:
1. Publish to PyPI: need to decide version names, set up workflow, etc. Having the pip install from git is already a nice improvement.
2. Remove list of dependencies in pre-commit config: Want to change something anyways, so easier to to separately. Current config still works in the meantime. See https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/474#discussion_r2529975209.
3. Remove list of dependencies / docker container in GH actions: Could be nice to use pip installation there too.
4. Good pinning for reproducibility in CI: Not sure if it is worth the effort, because doing it right means using something like uv or another extra tool.
5. Update README with instructions to install with pip: I would defer this change until we have figured out whether we want to publish to PyPI.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,20 +2,20 @@
 name = "bapctools"
 version = "0.0.0"
 dependencies = [
-    "pyyaml",
-    "colorama",
-    "argcomplete; platform_system != 'Windows'",
-    "python-dateutil",
-    "ruamel.yaml",
-    "matplotlib",
-    "requests",
-    "questionary",
+    "pyyaml~=6.0.3",
+    "colorama~=0.4.6",
+    "argcomplete~=3.6.3; platform_system != 'Windows'",
+    "python-dateutil~=2.9.0",
+    "ruamel.yaml~=0.18.16",
+    "matplotlib~=3.10.7",
+    "requests~=2.32.5",
+    "questionary~=2.1.1",
 ]
 requires-python = ">=3.10"
 
 [project.optional-dependencies]
 dev = [
-    "pytest",
+    "pytest~=9.0.1",
 ]
 
 [project.scripts]


pyproject.toml
```diff
@@ -1,3 +1,26 @@
+[project]
+name = "bapctools"
+version = "0.0.0"
+dependencies = [
+    "pyyaml",
```
niklasmohrin: I saw that the pre-commit config has some fixed versions for the dependencies; should we move them here?

RagnarGrootKoerkamp: i can't answer this

bapctools/__main__.py
```diff
@@ -0,0 +1,7 @@
+import sys
+
+from bapctools.tools import main
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
mpsijm: I think I would prefer it if `tools.py` is renamed to `__main__.py`, instead of adding this new file. This file already contains an `if __name__ == "__main__"` block near the bottom 🙂 

Also, `main()` returns `None`, so the `sys.exit` wrapper is not needed. The `main` function will `sys.exit(1)` by itself in case of errors or warnings. This means you don't have to edit `tools.py` (besides the imports and the file rename of course) :smile:

The only problem I foresee with that, is that `tools.py` currently exports a `test` function that is used in `test/test_problems.py`. If we do rename `tools.py` to `__main__.py`, can that test file have `from bapctools.__main__ import test`, or does that not work as I expect?
(As far as I can see, this currently is the only place where something is exported/imported from `tools.py`.)

pyproject.toml
niklasmohrin: Hm, there should probably be a way to tell precommit to just install the local package; I'll see what they offer

bapctools/__main__.py
niklasmohrin: About completion, backwards compatibility: I agree, we don't have to break anyone's setup here. I wouldn't add a new bt script in bin (maybe I confused the names in another comment). What is new is that `pip install` will give you a script called bt in the venv path.

I would leave a `bin/tools.py` around that ensures that everything still works. I believe that it should not be too much trouble to keep completion working. I don't know how exactly argcomplete works, but I think they probably just run the file without arguments and see what was given to them during the call. For our tools.py, something like
```python
sys.path += [__file__ / ".."]

from bapctools.cli import main

main()
```
would probably do the trick. (And this would also be a good place to print a deprecation warning if you ever decide to fully commit to pip installation by default.)

Another option is the builtin module runpy. I have actually performed a similar refactor in another project where we went with runpy: https://github.com/e-valuation/EvaP/pull/2387 However, there our script was also located in the correct place so that doing something like `../../manage.py` would add the repository root to the import search path, so the sys.path trick was not needed. So my previous suggestion is probably better because we don't have to think about runpy. I'll try if argcomplete works with that.

If this really doesn't work, the next best option I can think of is a refactor so that both `bapctools/__main__.py` and `bin/tools.py` do something like
```python

# (bin/tools.py does sys.path stuff first)

import argcomplete
from bapctools.cli import main, make_parser

parser = make_parser()
argcomplete.register(parser)
main(parser.parse_args())
```

mpsijm: > I think I would just rename tools.py to cli.py

Agreed, we should stick with an idiomatic (short) `__main__.py` :smile: I prefer `cli.py` over `tools.py`, thanks :slightly_smiling_face: 

> Another option is the builtin module runpy.

Never seen that before, I don't think that's necessary here? Your proposal of making `bapctools/__main__.py` and `bin/tools.py` do basically the same, sounds good to me :smile:

mpsijm:
Haha, all right! I'll delay the import rewrites from https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/473#discussion_r2507009962 anyway, but good to hear that the rebasing won't be an issue for you :slightly_smiling_face:

pyproject.toml
mpsijm: The current `.pre-commit-config.yaml` still works, but it's cleaner if we don't have to duplicate the dependencies here :slightly_smiling_face:

pyproject.toml
```diff
@@ -14,8 +31,5 @@ default-section = "standard-library"
 order-by-type = false
 
 [tool.pyright]
```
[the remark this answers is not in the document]
mpsijm: Pyright still works for me in NeoVim and from the command-line :slightly_smiling_face: (it still throws a bunch of errors, but that's currently intended: #102)

pyproject.toml
niklasmohrin: So do you want to have everything pinned like
```toml
dependencies = [
    "argcomplete==3.4.0; platform_system != 'Windows'",
    "matplotlib==3.9.1",
    # ...
]
[project.optional-dependencies]
types = [
    "types-colorama==0.4.15.20240311",
    "types-Pygments==2.18.0.20240506",
    "types-python-dateutil==2.9.0.20241003",
    "types-PyYAML==6.0.12.20240311",
    "types-requests==2.32.0.20240712",
]
```
or have the required dependencies be more relaxed with for example `argcomplete~=3.4`? I would recommend the latter, because the pinned versions could become deleted/yanked from PyPI and replaced with later patch releases (you can also vary this for every dependency).

If you want, there can even be an optional dependency group that pins the otherwise loose dependencies again. I think that such pinning is usually managed with an external tool like uv though, not sure how common it is to have it in your pyproject.toml.

By the way, I would give the pre-commit config for mypy a touch anyways, because it currently does not detect cross-file breakage. For example, say you changed a variable in config.py like so `ICPC_FILE_LIMIT: Final[str] = "100"` (it was an int before the change). Then still `pre-commit run mypy` passes because the file importing config.py was not checked, whereas just running `mypy -p bapctools` catches the error. See https://github.com/python/mypy/issues/13916

Soooo, what I want to say: I want to send you another PR for mypy and pre-commit anyways once this is done, and I could also sort these typing things out in that PR.

What do you think we should do in this PR? I would suggest going with `~=` specifiers, or even without any specifiers at all to revisit this in a separate PR. For duplicated types in pre-commit config vs. moving them to pyproject.toml, I don't have a strong opinion, except that the dependencies probably shouldn't be pinned by default.

mpsijm: Yeah, I agree that exact pinning is bad, I'm fine with only specifying major versions (or no versions at all). But if you intend to clean this up later anyways, it's fine keeping it as it is now 🙂

niklasmohrin:
I updated the PR description with the current information and my thoughts on what to do next. I think I am happy with the current state of the PR. Let me know what you think about the PR and the proposed next steps.

mpsijm:
I'm also happy with the current state and the list of things to do after merging this PR, thanks a ton! :smile: :heart: 

Because we have [NWERC](https://2025.nwerc.eu/) coming up (30 November), we'll wait with merging this until after that contest. Then we'll also have more time to do a thorough review :slightly_smiling_face: I'll keep using your branch during problem development to see if things break (so far, so good)! In the mean time, feel free to investigate some of the follow-ups if you have the time (also fine if you don't have the time) :smile:

mpsijm:
Thanks again @niklasmohrin for your work! :smile: :heart: Rebasing turned out to be not too difficult, luckily: Git is quite good at tracking the file renames, it seems. :sunglasses:

I've opened #489 and #490 as follow-ups based on your suggestions :slightly_smiling_face:

PR #476, opened 2025-11-10
Use immutable types for cached data

mzuenni:
Related to #102. This should not really change anything. It just ensures that cached stuff is typed as immutable `Sequence` or `Mapping` and that those sequences are actually immutable.

### DIFF
--- a/bin/generate.py
+++ b/bin/generate.py
@@ -81,7 +81,8 @@ UNIQUE_TESTCASE_KEYS: Final[Sequence[str]] = (
     "generate",
     "retries",
     "count",
-) + tuple(e[1:] for e in config.KNOWN_TEXT_DATA_EXTENSIONS)
+    *(e[1:] for e in config.KNOWN_TEXT_DATA_EXTENSIONS),
+)
 
 
 def is_testcase(yaml: YAML_TYPE) -> bool:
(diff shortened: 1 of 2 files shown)


mzuenni:
> Does Sequence require tuples instead of lists?

No. But why only rely on mypy when tuples allow to simply enforce this at runtime?

bin/export.py
```diff
@@ -478,7 +478,7 @@ def update_contest_id(cid: str) -> None:
 
 
 def export_contest(cid: Optional[str]) -> str:
-    data = contest_yaml()
+    data = dict(contest_yaml())
```
mpsijm: `dict(None)` crashes, I think we should cast the optional immutable `Mapping` to `dict` _after_ we check `if not data` in the lines below?

bin/generate.py
```diff
@@ -76,12 +76,12 @@ def assert_type(
...
 
 
-UNIQUE_TESTCASE_KEYS: Final[Sequence[str]] = [
+UNIQUE_TESTCASE_KEYS: Final[Sequence[str]] = (
     "copy",
     "generate",
     "retries",
     "count",
-] + [e[1:] for e in config.KNOWN_TEXT_DATA_EXTENSIONS]
+) + tuple(e[1:] for e in config.KNOWN_TEXT_DATA_EXTENSIONS)
```
mpsijm: To remove the explicit `tuple` functions, this (and similar occurrences) could be written as:

```suggestion
    *(e[1:] for e in config.KNOWN_TEXT_DATA_EXTENSIONS),
)
```

bin/export.py
mzuenni: i would have hoped that mypy catches this...

mzuenni: wait, no it would have catched this :) `contest_yaml()` does never return `None`

bin/generate.py
mzuenni: not sure if i like this more, but will change.

PR #478, opened 2025-11-13
Stricter Yaml Parsing

mzuenni:
The final goal here is to change the signature of `read_yaml` to return `object` instead of `Any`

### DIFF
--- a/bin/contest.py
+++ b/bin/contest.py
@@ -230,7 +230,7 @@ def call_api(method: str, endpoint: str, **kwargs: Any) -> "requests.Response":
     return r
 
 
-def call_api_get_json(url: str, raise_for_status: bool = False) -> Any:
+def call_api_get_json(url: str) -> Any:
     r = call_api("GET", url)
     r.raise_for_status()
     try:
--- a/bin/generate.py
+++ b/bin/generate.py
@@ -63,13 +63,16 @@ class ParseException(Exception):
 
 
 def assert_type(
-    name: str, obj: object, types: list[type[object]] | type[object], path: Optional[Path] = None
+    name: str,
+    obj: object,
+    types: tuple[type[object], ...] | type[object],
+    path: Optional[Path] = None,
 ) -> None:
-    if not isinstance(types, list):
-        types = [types]
-    if any(isinstance(obj, t) for t in types):
+    if isinstance(obj, types):
         return
-    named_types = " or ".join(str(t) if t is None else t.__name__ for t in types)
+    if not isinstance(types, tuple):
+        types = (types,)
+    named_types = " or ".join("None" if t is None else t.__name__ for t in types)
     raise ParseException(
         f"{name} must be of type {named_types}, found {obj.__class__.__name__}: {obj}",
         path,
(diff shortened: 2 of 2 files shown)


test/yaml/problem/invalid.yaml
```diff
@@ -22,7 +22,7 @@ yaml:
   limits:
     time_multipliers:
       mumbo: jumbo
-warn: "found unknown problem.yaml key: mumbo in `limits.time_multipliers`"
+warn: "found unknown problem.yaml key: mumbo in `time_multipliers`"
```
mpsijm: I don't like that the warning no longer shows the full path. Many keys can occur in different paths in the YAML tree, e.g. keys for authors in `credits`

bin/skel.py
```diff
@@ -133,8 +133,9 @@ def new_problem() -> None:
         custom_output = True
 
     # Read settings from the contest-level yaml file.
+    contest_data = {k: str(v) for k, v in contest.contest_yaml().dict().items()}
```
mpsijm: Do we want to cast to `str`, or assert that they are `str`?

mpsijm (changes requested):
Did a quick scroll through the diff 🙂

bin/skel.py
mzuenni: We also use these keys for copy and substitute later on.

mzuenni: We always substituted all of them, so people might expect that all keys from the `contest_yaml` are accessible?

test/yaml/problem/invalid.yaml
mzuenni: We never did this for any other key? Removing the `limits.` here only made it consistent with every other place which allows reusing code (also this key is also globally unique).
So yeah in general I am fine with giving more info, but in most cases that is not necessary and with this PR I don't really want to change any behavior, just cleanup/type check the code better

bin/generate.py
```diff
@@ -1023,7 +1033,7 @@ def needed(
                         if not t.config.solution.generate_interaction(bar, cwd, t):
                             return False
                         used_solution = True
-                        meta_yaml["interactor_hash"] = interactor_hash
+                        meta_yaml.interactor_hash = cast(dict[Any, Any], interactor_hash)
```
mpsijm: Shouldn't the type of `MetaYaml.interactor_hash` be `dict[str, dict[str, str]]` instead of `dict[object, object]`? That would avoid the need for a `cast` due to the invariance of key/value types in `dict` :slightly_smiling_face:

bin/skel.py
mpsijm: Fair enough, I updated the comment to match this intent :slightly_smiling_face:

bin/generate.py
mzuenni: That is true from a logical perspective, but never verified. If we define it as `dict[str, dict[str, str]]` we would need to actually verify this type while parsing the MetaYaml (both keys and values could be something weird). But for the current code this actually never needed. Thus, I would argue that it's cleaner to assume the most general type which avoids any need for type checking during the read of the yaml file. That is also always safe, since by definition any key and any value is an object... MyPy just doesn't understand this ):

mpsijm (approved):
Looks good to me, is there a reason why this PR is still marked as "draft"? 🙂

mzuenni:
There are still some `Any` like the `request.json()` things and probably some more implicit `Any` it would be nice to get rid of all of them :D

mpsijm (changes requested):
Sure, but there's still quite a few `Any` types scattered across our code base, so perhaps it's best to keep this PR focused to YAML parsing :slightly_smiling_face:

bin/contest.py
```diff
@@ -230,7 +230,7 @@ def call_api(method: str, endpoint: str, **kwargs: Any) -> "requests.Response":
     return r
 
 
-def call_api_get_json(url: str) -> Any:
+def call_api_get_json(url: str, raise_for_status: bool = False) -> Any:
```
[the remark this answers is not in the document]
mzuenni: whoops

mzuenni:
Fine, we can already merge this then, I guess?

PR #480, opened 2025-11-16
YamlParser: show the full parent path in warnings

mpsijm:
Depends on #478.

Based on a comment in that PR: I'd like the `YamlParser` to show exactly in which (sub-)key an error occurs. So, I've let `YamlParser` not only remember its direct parent, but its entire path of parents.

For parsing lists of persons and lists of sources, I manually put the list indices in the parent path.

### DIFF
--- a/bin/contest.py
+++ b/bin/contest.py
@@ -59,7 +59,7 @@ class ContestYaml:
 
 
 class ProblemsYamlEntry:
-    def __init__(self, yaml_data: dict[object, object]) -> None:
+    def __init__(self, yaml_data: dict[object, object], index: int) -> None:
         self.ok = False
         parser = YamlParser("problems.yaml", yaml_data)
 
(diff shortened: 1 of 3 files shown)


mzuenni:
Do we also want to print the path on any other error and not just for unknwon keys?

mzuenni:
Looks good, there are some more you might want to consider:
- [problem.py](https://github.com/mpsijm/BAPCtools/blob/yaml-parser-warn-parent-path/bin/problem.py#L122) Line 122
- [problem.py](https://github.com/mpsijm/BAPCtools/blob/yaml-parser-warn-parent-path/bin/problem.py#L289) Line 289
- [problem.py](https://github.com/mpsijm/BAPCtools/blob/yaml-parser-warn-parent-path/bin/problem.py#L323) Line 323
- [contest.py](https://github.com/mpsijm/BAPCtools/blob/yaml-parser-warn-parent-path/bin/contest.py)
  - ContestYaml
  - ProblemsYamlEntry

PR #485, opened 2025-11-24
allow other tex commands

mzuenni:


bin/latex.py
```diff
@@ -242,6 +242,39 @@ def make_environment(builddir: Path) -> dict[str, str]:
...
 
+TEX_MAGIC_REGEX: Final[str] = "^%\\ ?!tex\\s+program\\s*=\\s*(.*)$"
+COMPILED_TEX_MAGIC_REGEX: Final[re.Pattern[str]] = re.compile(TEX_MAGIC_REGEX, re.IGNORECASE)
+
+
+def get_tex_command(tex_path: Path, bar: PrintBar) -> tuple[str, str]:
+    command = config.args.tex_command
+    if command is None and tex_path.is_file():
+        try:
+            with tex_path.open() as f:
```
RagnarGrootKoerkamp: Are you parsing the source to determine which latex compiler to use?
Can you add some example command and usecase in a comment?

RagnarGrootKoerkamp (approved):

bin/latex.py
RagnarGrootKoerkamp: If you add a comment with a link or so then ready to go :)

Issue #490, opened 2025-12-15
Wrap up PyPI publishing and clean up dependencies


mpsijm:
Follow-up of #474. @niklasmohrin suggested the following steps:

- [x] 1. Publish to PyPI: need to decide version names, set up workflow, etc. (#510)
    - Note that if #489 is resolved before this, the PyPI package should only install the Checktestdata binary that matches the user's system architecture.
- [x] 2. Remove list of dependencies in pre-commit config: Want to change something anyways, so easier to to separately. Current config still works in the meantime. See https://github.com/RagnarGrootKoerkamp/BAPCtools/pull/474#discussion_r2529975209. Fixed in #493 
- [ ] 3. Remove list of dependencies / docker container in GH actions: Could be nice to use pip installation there too.
- [ ] 4. Good pinning for reproducibility in CI: Not sure if it is worth the effort, because doing it right means using something like uv or another extra tool.
- [x] 5. Update README with instructions to install with pip: ~~I would defer this change until we have figured out whether we want to publish to PyPI.~~
   - [x] I've decided to already put experimental Pip-installation instructions in the README in 73843fd7fbc2b53c8dcf1916f91305f255cadb2a, these should of course be updated after publishing to PyPI. → Updated to make `pip install bapctools` the primary installation method in 0853b9c0e2f239c6fdd194d37ba88dd6497aa659

mpsijm:
I'd like to eventually automate the PyPI releases, so then having a version number like `YYYY.MM.DD.pp` would sound reasonable to me. The patch-number would be an ever-increasing integer equal to the number of commits to ensure correct version ordering. Or we could even do `rXXXX.hash`, making it the same as https://aur.archlinux.org/packages/bapctools-git (although having the date in the version may be a nice thing to have).

niklasmohrin:
Ah, actually +cccccc is for when there are new commits since the version. See https://packaging.python.org/en/latest/discussions/versioning/#local-version-identifiers

PR #493, opened 2025-12-19
#490 Move mypy config to pyproject.toml and run outside of pre-commit environment

niklasmohrin:
See #490.

There are two things that change with this PR, mostly inspired by [this article](https://jaredkhan.com/blog/mypy-pre-commit):
1. The typing dependencies are not declared in the pre-commit anymore, but instead defined as another optional dependency group in pyproject.toml. However, pre-commit doesn't allow specifying anything about the environments it creates outside of the pre-commit config file. Thus, we can either run `pip install` as part of the check (this would be slow), or we can omit any isolated setup and just run the `mypy` command with whatever environment is currently available (thus developers need to set this up by running `pip install --editable .[types]`; `.[dev]` also works). This is a bit sad, because having the isolated environment ensures reproducibility. I think that having the dependencies in pyproject.toml outweighs this. With the types automatically installed as part of `.[dev]`, I now have these type hints available in my editor and can spot any errors before even committing. Note that I also relaxed the version specifiers a bit.
2. Mypy now always runs on the entire bapctools package. Previously, mypy would only check changed files. However, this means that any typing errors in files that only import the changed files do not surface. Really the only sensible way to run mypy is by letting it check everything. This is initially quite slow, however the checking is fast once the mypy cache is initialized (~ 5s -> ~200ms). There is some discussion on this here: https://github.com/python/mypy/issues/13916

Additionally, I moved the mypy configuration into pyproject.toml so that running mypy from the console or as a language server uses the same config. In doing so, I also checked which arguments are actually still required.
- excluding test is not needed, because we are always telling mypy to check the bapctools package
- no-incremental is not needed since the bug is fixed https://github.com/python/mypy/issues/12664
- scripts-are-modules: not needed because we actually have modules now

Finally, some notes:
- The `bapctools[types]` dependency specification is not really documented, not because it is not expected, but because people think it is clear that this should work. See https://github.com/pypa/pip/issues/11296
- We still need the python version for mypy, they do not use `project.requires-python` (for fair reasons). See https://github.com/python/mypy/issues/19349

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,8 +16,8 @@ repos:
     hooks:
     -   id: mypy
         name: mypy
+        entry: mypy
         pass_filenames: false
-        entry: "mypy --package bapctools"
         language: system
         require_serial: true # don't split filenames into multiple processes
         exclude: ^(bapctools/resources|bin/misc|test)/
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,6 +52,7 @@ exclude = ["bapctools/resources"]
 
 [tool.mypy]
 python_version = "3.10"
+packages = "bapctools"
 exclude = ["^bapctools/resources/"]
 check_untyped_defs = true
 enable_error_code="truthy-bool,truthy-iterable"


niklasmohrin:
> Apologies for

of course, all good, don't worry

> I agree that having the pre-commit hook install the entire environment is wasteful

Just for reference, pre-commit would also cache their environment, just separately from the development venv; the advantage of that was that you couldn't mess up the pre-commit venv from the outside, whereas now you could, for example, install a typing package locally but forget to add it to pyprojct.toml. In that case typechecking will still fail in CI

I still think the current setup is better because all failures are caught in CI, whereas before CI and local pre-commit would miss exactly the same cases :D

> CI for reproducibility

yes, but not fully because CI is also always installing the newest available packages according to the version ranges in pyproject.toml (at least the ones not using the docker image)

Issue #491, opened 2025-12-16
Feature: in generators, make `count` accept a range


thorehusfeldt:
I would like to be able to specify that my (fictional) generator that produces line segments of inclination theta is called like this:

```yaml
- 'rot':
    generate: rotated_line_segment --theta {count}
    count: [-90, 90, 10] # in, out, step
```

This should call the generator with `theta` ranging over `range(-90, 90, 10)`, with names `rot--90`, `rot--80`, ..., `rot-80`.

I n `generators.#test_case` I suggest to change
```cue
	count?:    int & >=1 & <=100
```
to
```cue
	count?:    int & >0 & <=100 | [int, int] | [int, int, int]
```
(We should maybe maintain that the resulting range has at least 1 and at most 100 integers, but this not a schema issue.)

There is a tiny wrinkle with `count: 10` enumerating 1,...,10 and `count: [0, 10]` enumerating `0,...,9`. I don’t have a strong opinion about this.

mzuenni:
I am against adding more festures like this. This is a can of worms and not necessary at all. If we add this i already see the next request like "hey i want to specify exactly the values [1,2,4,8,16]. Why cant I specify a list instead of a range"...

thorehusfeldt:
(I know I just closed this, but @mzuenni is of course right about explicit lists being a completely valid other feature request . We should then also allow `count: [int,...]`as well as `count: string` with ruby-like abbreviations `"34..37"` for `[34, 35, 36, 37]` and slice-inspired `10:25:5` for `[10, 15, 20]`.)

PR #494, opened 2025-12-20
`testcase.count` learns list[int], =~"^-?\d+\.\.=-?\d+$"

thorehusfeldt:
(Edited to remove slice syntax.)

Draft implementation that handles the following testcase generation rules in `identity`:

```yaml
  - count1:
              generate: stdout.py 700 {count}
              count: [1] # same as count: 1
          - primes:
              generate: stdout.py {count}
              count: [907, 911, 919]
          - dotdot:
              generate: stdout.py {count}
              count: 801..=803 # same as [801, 802, 803]
```

The main change is that `parse_count` in `generators.py` returns `list[int]` rather than `int`.

The test case name becomes `-{idx}` where `idx` ranges over the list items. In particular, this generates files with `--` (when the list item is negative).

This extends the expressibility of count with (what I consider) accessible syntax with minimal extra code. The following are the same:
```yaml
count: 3
---
count: [1, 2, 3]
---
count: 1..=3
```

Implements some (rejected) ideas in #271 , #491

### DIFF
--- a/test/yaml/generators/invalid_yaml/invalid.generators.yaml
+++ b/test/yaml/generators/invalid_yaml/invalid.generators.yaml
@@ -420,24 +420,24 @@ data:
           generate: my_generator {count}
           count: 1...2
 ---
-# Count range expression: range must not be empty
-data:
-  sample: {}
-  secret:
-    data:
-      - "":
-          generate: my_generator {count}
-          count: 2..=1
----
-# Count range expression: range must not be too long
-data:
-  sample: {}
-  secret:
-    data:
-      - "":
-          generate: my_generator {count}
-          count: 0..=999999999
----
+# # Count range expression: range must not be empty
+# data:
+#   sample: {}
+#   secret:
+#     data:
+#       - "":
+#           generate: my_generator {count}
+#           count: 2..=1
+# ---
+# # Count range expression: range must not be too long
+# data:
+#   sample: {}
+#   secret:
+#     data:
+#       - "":
+#           generate: my_generator {count}
+#           count: 0..=999999999
+# ---
 # Count range expression: no float
 data:
   sample: {}


mzuenni:
> Why is 1000 even a thing?

Good question... I thought the limit was supposed to be 100?

> I like low..=high and low..high

I still don't really like the syntax of `from..=to`, but if everyone else is fine with this, it's ok for me.

(@thorehusfeldt github says "This branch cannot be rebased due to conflicts")

RagnarGrootKoerkamp:
100 sgtm 

I like both a..b and a..=b, but of course I am rust-pilled.

bapctools/generate.py
```diff
@@ -97,7 +99,9 @@ def is_directory(yaml: YAML_TYPE) -> bool:
 
 
 def has_count(yaml: YAML_TYPE) -> bool:
-    return isinstance(yaml, dict) and "count" in yaml and isinstance(yaml["count"], int)
+    return (
+        isinstance(yaml, dict) and "count" in yaml and isinstance(yaml["count"], int | list | str)
```
mzuenni: the second `isinstance` looks weird, shouldn't it be `isinstance(yaml["count"], (int, list, str))`

mzuenni (changes requested):

bapctools/generate.py
thorehusfeldt: As far as I understand, in `isinstance(object, classinfo)`, the value of `classinfo` can be a `Union` type (since python 3.10), which can be written `int | str` since I don’t know when.

mzuenni: ah, interesting, didn't know this actually works. Should we still change this to match the rest of the codebase?

mzuenni: There are even more with the old syntax, because we have custom parsing functions similar ot isinstance that (currently) take a single type or a tuple of types, these should also be consistent. I think it is easier to stay with the old syntax ^^'

thorehusfeldt:
@RagnarGrootKoerkamp , this, too, is ready to be merged. There is an unresolved conversation between @mzuenni and me about python style above, which is wider than the present PR and I’m willing to have somewhere else.

bapctools/generate.py
RagnarGrootKoerkamp: (I like the new `A | B` syntax, but don't really care either way; consistency on this is not so important IMO.)

bapctools/generate.py
```diff
@@ -50,6 +50,8 @@
 
 YAML_TYPE = Optional[str | dict[object, object]]
 
+DOTDOT = re.compile(r"^(-?\d+)\.\.=(-?\d+)$")
```
RagnarGrootKoerkamp: rename from `DOTDOT` to `COUNT_RANGE` or `INCLUSIVE_RANGE_REGEX` so; or `DOTDOTEQUALS` otherwise.

test/yaml/generators/valid_yaml/rich-generators.yaml
```diff
@@ -10,6 +10,18 @@ data:
       'count-100':
         generate: my_generator {count}
         count: 100
+      'count-list':
+        generate: my_generator {count}
+        count: [1]
```
RagnarGrootKoerkamp: add some longer lists here.

doc/generators.md
```diff
@@ -41,7 +41,8 @@ The two main object types are `directory` and `generator`. The root of `generato
 - `copy: <base_path>`: For all known extensions `<ext>` the corresponding file `<base_path>.<ext>` will be copied to the specified location should it be present. `<base_path>` must be relative to `generators/`.
 - `<ext>: <content>`: A file with extension `ext` and the `content` will be generated. `<ext>` must be a known file extension.
 - `count: <int>`. To generate multiple Generator objects. If `generate` is used and `{seed}` or `{seed:(0-9)+}` is present all Generator objects will use a different seed. The arguments of `generate` may contain `{count}` to refer to the index of this generator invocation.
```
RagnarGrootKoerkamp: list extra types

bapctools/generate.py
```diff
@@ -1745,24 +1767,28 @@ def add_known(obj: TestcaseRule | Directory) -> None:
...
-                bar.log(f"Found large count: {count}.")
-            return count
+            match count:
+                case int():
+                    count_list = list(range(1, count + 1))
+                case list(idx_list):
+                    count_list = idx_list
+                case str(s):
+                    if m := DOTDOT.match(s):
+                        count_list = list(range(int(m[1]), int(m[2]) + 1))
```
RagnarGrootKoerkamp: This should probably just error if `m[2] < m[1]`.

test/yaml/generators/invalid_yaml/invalid.generators.yaml
```diff
@@ -375,6 +375,42 @@ data:
...
+       count: 1 ..= 2
+---
+# Count range expression: no float
+data:
+  sample: {}
+  secret:
+    data:
+    - '':
+       generate: my_generator {count}
+       count: 1..=2.0
```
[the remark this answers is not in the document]
thorehusfeldt: Syntactically, `5..=1` is valid. (It produces the empty list, which is then warned for in BAPCtools and replaced by `[1]`.) 

We _can_ extend the schema to forbid it, and I certainly thought about it. But such rules take a lot of “space” in the schema for very little practical gain. So I decided against it. Say the word and I add it.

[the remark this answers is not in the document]
RagnarGrootKoerkamp: Yeah makes sense as-is then. I did add some more explicit python code for erroring on this.

PR #496, opened 2025-12-26
Submissions & Expectations

mzuenni:
I think we should change the code such that there is no longer any hardcoded logic about what folder is used for what, all of that should be encoded in the (default)submissions.yaml.

TODOS:
- [x] parse `submissions.yaml`
- [x] update expected verdict code
- [x] update run method
- [x] update `bt time_limit`
- [x] add test problem for submissions.yaml
- [x] verify judgemessage
- [x] resolve all new `TODO`

### DIFF
--- a/bapctools/program.py
+++ b/bapctools/program.py
@@ -347,7 +347,7 @@ class Program:
             )
         return candidates
 
-    def _get_entry_point(self, files: list[Path], bar: ProgressBar) -> tuple[str, str, str]:
+    def _get_entry_point(self, files: list[Path], bar: ProgressBar) -> tuple[Path, Path, str]:
         binary = self.tmpdir / "run"
         mainfile = None
         if not self.has_deps:
@@ -361,7 +361,7 @@ class Program:
         else:
             mainfile = self.tmpdir / self.source_files[0].name
         mainclass = mainfile.with_suffix("").name
-        return (str(binary), str(mainfile), str(mainclass))
+        return (binary, mainfile, str(mainclass))
 
     # Sets self.language and self.env['mainfile']
     def _get_language(self, bar: ProgressBar) -> bool:
--- a/bapctools/run.py
+++ b/bapctools/run.py
@@ -380,11 +380,12 @@ class Submission(program.Program):
             bar.warn(f"Unknown language: {self.expectations.language}{msg}")
         return candidates
 
-    def _get_entry_point(self, files: list[Path], bar: ProgressBar) -> tuple[str, str, str]:
+    def _get_entry_point(self, files: list[Path], bar: ProgressBar) -> tuple[Path, Path, str]:
         if self.expectations.entrypoint is None:
             return super()._get_entry_point(files, bar)
         entrypoint = self.expectations.entrypoint
-        return (entrypoint, entrypoint, entrypoint)
+        file = self.tmpdir / entrypoint
+        return (file, file, entrypoint)
 
     # Run submission on in_path, writing stdout to out_path.
     # args is used by SubmissionInvocation to pass on additional arguments.


RagnarGrootKoerkamp:
I'd say multiple braces expand to the Cartesian product of options.

Indeed, no nesting of braces.

Nesting [] inside {} is also not quite needed I think, but we can allow it.

thorehusfeldt:
> File names can have whitespaces.

Ours cannot. I broached the subject exactly because I want us to agree on what can be inside `{...}`. Your valuable work helps us agree on that.

I am _more than fine_ with allowing only alphanumerics (possibly with `_-.`, but not space) inside `{...}`. I am also fine with forbidding more than one `{...}`. However, my suggested implementation above supports all finds of stuff; implementation is in fact easier than specification in this case.

mzuenni:
@thorehusfeldt the spec currently states this:
>Using [xyz] to match only a subset of characters is not supported.

also, can you take a look at the schemas, this is failing now? (I tried to fix it, but I don't really know how to use cue)

thorehusfeldt:
Looking at the schema now.  In `test/yaml/validate_yaml.sh` I added

```bash
declare -A schema_map=(
["*generators.yaml"]="#Generators"
["*problem.yaml"]="#Problem"
["*test_group.yaml"]="#test_group_configuration"
["*submissions.yaml"]="#Submissions"     #<<<<<< THIS IS NEW
# add more schemas here
)
```
to look through the whole repo. The errors so far are

```
    "**/nested_wa.py": field not allowed:
        ../problems/testexpectations/submissions/submissions.yaml:63:1
    "expansion/ac.{py,cpp}": field not allowed:
        ../problems/testexpectations/submissions/submissions.yaml:18:1
    "wa/42.py"."secret/*{ 4 }{ 2 }": field not allowed:
        ../problems/testexpectations/submissions/submissions.yaml:45:3
    "wa/ac*": field not allowed:
        ../problems/testexpectations/submissions/submissions.yaml:27:1
    "wa/even.py"."secret/*{,0,1,2,3,4,5,6,7,8,9}{0,2,4,6,8}": field not allowed:
        ../problems/testexpectations/submissions/submissions.yaml:40:3
    "wa/odd.py"."secret/*{1,3,5,7,9}": field not allowed:
        ../problems/testexpectations/submissions/submissions.yaml:35:3
    tle.sample.use_for_time_limit: field not allowed:
        ../problems/testexpectations/submissions/submissions.yaml:73:5
    tle.secret.use_for_time_limit: field not allowed:
        ../problems/testexpectations/submissions/submissions.yaml:75:5
```

I’ll see how far I get with them. The first `**/nested_wa.py` _should_ be not allowed, right?

test/problems/testexpectations/submissions/submissions.yaml
```diff
@@ -0,0 +1,75 @@
...
+match_wa.py:
+  required: [WA]
+
+# * does not match /
+"*/nested_wa.py":
+  required: [WA]
+
+# check message
+wa/hello.py:
+  message: hello
```
RagnarGrootKoerkamp: hmm, `message` is not very clear. It means 'judge error message' right? Should we improve it?

bapctools/resources/support/schemas/submissions_yaml_schema.json
RagnarGrootKoerkamp: this looks out-of-date or like a placeholder?

bapctools/expectations.py
```diff
@@ -0,0 +1,253 @@
...
+from bapctools.testcase import Testcase
+from bapctools.util import error, fatal, read_yaml, warn, YamlParser
+from bapctools.verdicts import Verdict
+
+if TYPE_CHECKING:
+    from bapctools.problem import Problem
+    from bapctools.run import Submission
+
+
+class Person:
```
RagnarGrootKoerkamp: shouldn't this stay in `problem.py`?

bapctools/expectations.py
```diff
@@ -0,0 +1,253 @@
...
+        parts[i] = part
+    glob = "".join(parts)
+    # stars match anything exept /
+    glob = glob.replace(re.escape("*"), "[^/]*")
+    # match from start and only match complete directories
+    glob = f"^{glob}(/|$)"
+    return re.compile(glob)
+
+
+class TestcaseExpectation:
```
RagnarGrootKoerkamp: Add a brief comment to `{testcase,submission,}expectation` on what they are (relative to each other)

bapctools/resources/config/submissions.yaml
RagnarGrootKoerkamp: Should we add `mixed` to the spec, which is basically `permitted: <everything>`?
That might help with backwards compatibility of problems that already have such a directory.

bapctools/resources/support/schemas/submissions_yaml_schema.json
thorehusfeldt: Fixed.

test/problems/testexpectations/submissions/submissions.yaml
thorehusfeldt: It does not have to be a judge _error_ message. For instance, in `guess` it‘s something like “Picked 52 randomly”. As far as I remember I was unable to get both `judge_message` and `team_message` through the discussions, but I forget the reasoning.

bapctools/expectations.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def all_matches(self, submission: "Submission") -> SubmissionExpectation:
+        if submission.path in self._combined:
+            return self._combined[submission.path]
+
+        languages = set()
+        entrypoints = set()
+        authors = set()
+
+        combined = SubmissionExpectation(submission.name, {})
```
RagnarGrootKoerkamp: For my understanding (maybe add a comment): so here you're parsing the submission 'metadata' rather than expectations, and since we allow globs, you have to iterate over all the matching ones and combine them.

bapctools/generate.py
```diff
@@ -2276,7 +2275,11 @@ def reorder(self) -> bool:
             return False
 
         testcases = [t for t in ts_pair[0] if t.in_path in testcase_filter]
-        submissions = [s for s in ts_pair[1] if s.expected_verdicts != [Verdict.ACCEPTED]]
+
+        def not_accepted(s: run.Submission) -> bool:
+            return any({Verdict.ACCEPTED} != s.expectations.all_permitted(t) for t in testcases)
```
RagnarGrootKoerkamp: A submission is accepted is there is at least one `permitted: [AC]` for it, right?

Ah wait nvm; that's already handled inside the `all_permitted` and the `any` is over testcases: a submission is not accepted if it's permitted to non-AC on at least one testcase.

Maybe add a comment?

bapctools/program.py
```diff
@@ -329,12 +334,8 @@ def _get_language(self, bar: ProgressBar) -> bool:
                     Program.warn_cache.add(lang.id)
                     bar.debug(f"Falling back to {name}.")
 
-            if len(files) == 0:
-                self.ok = False
-                bar.error(f"No file detected for language {name} at {self.path}.")
-                return False
-
             self.language = lang
+            # TODO determine entry point
```
RagnarGrootKoerkamp: was this done already? I see logic for it in `expectations.py`.

bapctools/run.py
```diff
@@ -284,26 +291,19 @@ def __init__(
...
 
-        # The first element will match the directory the file is in, if possible.
-        self.expected_verdicts = self._get_expected_verdicts()
-
+        if self.path.absolute().is_relative_to((problem.path / "submissions").absolute()):
+            self.expectations: expectations.SubmissionExpectation = (
+                problem.expectations().all_matches(self)
+            )
+        else:
+            self.expectations = expectations.SubmissionExpectation(self.name, {})
```
RagnarGrootKoerkamp: What does this mean? Zero expectations? I think currently we assume 'accepted/`-like expectations for anything outside `submissions` (or anything outside `submissions/{accepted,wrong_answer,...}`?)

bapctools/run.py
```diff
@@ -284,26 +291,19 @@ def __init__(
...
+        self.expected_results = self._parse_expected_results()
```
RagnarGrootKoerkamp: Add a comment explaining `self.expectations` vs `self.expected_results`

bapctools/run.py
```diff
@@ -488,80 +503,122 @@ def process_run(run: Run) -> None:
...
-        # Use a bold summary line if things were printed before.
-        if bar.logged:
-            color = (
-                Style.BRIGHT + Fore.GREEN
-                if self.verdict in self.expected_verdicts
-                else Style.BRIGHT + Fore.RED
-            )
-        else:
-            color = Fore.GREEN if self.verdict in self.expected_verdicts else Fore.RED
+        # TODO what color is intended here??
```
RagnarGrootKoerkamp: yeah, i think green if expectations passed (as it's now) is good.

bapctools/problem.py
```diff
@@ -1642,45 +1612,51 @@ def determine_time_limit(problem) -> bool:
...
             durations = [get_slowest(result)[1] for result in verdict_table.results]
-            selected = durations.index(select(durations))
+            selected = durations.index(select_duration(durations))
             testcase, duration = get_slowest(verdict_table.results[selected])
             return verdict_table.submissions[selected], testcase, duration
 
-        submission, testcase, duration = run_all(lambda vs: vs == [verdicts.Verdict.ACCEPTED], max)
+        # determine lower bound for time limit
+        submission, slowest, duration = run_all(
+            lambda s, t: all(not e.lower_time_limit for e in s.expectations.all_matches(t)),
```
RagnarGrootKoerkamp: Hmm, how did/should we spec what happens when multiple expectations (implicitly) say something about whether to use a submission for time limits? As coded: as soon as one expectation states/implies it, then we use it.

Probably that's good, because it means expectations work in isolation.

Either way, maybe negate the bool returned by the lambda? `use_test_case` rather then `skip_test_case` is more intuitive? Then it's simply `any(e.lower_time_limit)`

RagnarGrootKoerkamp (commented):
Some more detailed comments now. I have not yet run this in the terminal myself, but I trust that that all looks good.

bapctools/expectations.py
```diff
@@ -0,0 +1,253 @@
...
+                return default
+            if any(v not in KNOWN_VERDICTS for v in verdicts):
+                warn(
+                    f"some values for key `{parser.parent_path}.{key}` in {parser.source} are unknown. SKIPPED."
+                )
+                return default
+            return {KNOWN_VERDICTS[v] for v in verdicts}
+
+        self.permitted: set[Verdict] = extract_verdicts("permitted")
+        self.required: set[Verdict] = extract_verdicts("required", self.permitted)
```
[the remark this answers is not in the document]
mzuenni: currently required defaults to permitted? (and permitted defaults to `[AC, WA, TLE, RTE]`)

bapctools/run.py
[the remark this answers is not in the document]
mzuenni: as long as domjudge does not support the `submissions.yaml` we must keep this?

bapctools/problem.py
```diff
@@ -1003,15 +977,8 @@ def add(s: Path) -> None:
...
 
-        # - first all submission with just one verdict (grouped by that verdict and sorted by the path)
-        # - then by subdir
-        # - then by list of verdicts
-        # - then by name
-        def submissions_key(
-            x: run.Submission,
-        ) -> tuple[int, str, Sequence[verdicts.Verdict], str, str]:
-            group = "" if len(x.expected_verdicts) == 1 else x.subdir
-            return (len(x.expected_verdicts), group, x.expected_verdicts, x.subdir, x.name)
```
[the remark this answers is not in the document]
mzuenni: yeah, this is kind of sad but i had no better idea yet ^^'

bapctools/expectations.py
RagnarGrootKoerkamp: Maybe more to `person.py`?

bapctools/problem.py
```diff
@@ -954,16 +915,29 @@ def selected_or_accepted_submissions(problem) -> Sequence[run.Submission]:
             return submissions
         else:
             return tuple(
-                s for s in submissions if s.expected_verdicts == [verdicts.Verdict.ACCEPTED]
+                s
+                for s in submissions
+                if all(
```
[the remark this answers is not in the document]
RagnarGrootKoerkamp: But you currently test `e.required == {verdicts.Verdict.ACCEPTED}`?

So that should be `e.permitted == {AC}` right?

But then actually it's sufficient if the intersection of permitted verdicts over all expectations is `{AC}`.

bapctools/run.py
```diff
@@ -488,80 +503,122 @@ def process_run(run: Run) -> None:
...
+                    # the spec explicitly says judgemessage, not cerr/cout
+                    judgemessage = run.feedbackdir / "judgemessage.txt"
+                    if (
+                        message is not None
+                        and judgemessage.is_file()
+                        and message in judgemessage.read_text(errors="replace")
+                    ):
+                        message = None
+                else:
+                    # if we do not have verdict we skipped that case
```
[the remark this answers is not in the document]
RagnarGrootKoerkamp: Ah yes, I agree that erroring here is good. If you have a TLE submission that you want to run just on 1 slow case you want the `required: [TLE]` to trigger if it does not actually TLE on it. (And if you did not want it, then it's clear what's going on.)

The opposite of only activating `required:` constraints when really running on _all_ cases is more subtle and could hide important errors.

But yeah, we _could_ slightly modify the error message, but maybe not very important if it's annoying to do so.

bapctools/problem.py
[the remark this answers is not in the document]
mzuenni: maybe checking if all `permitted` entries are `{AC}` makes more sense?

bapctools/resources/support/schemas/submissions.cue
```diff
@@ -0,0 +1,33 @@
...
+let glob_path = =~"^(\(component_re)/)*\(component_re)$" & !~"\\*\\*"
+
+#Submissions: {
+	[glob_path]: #submission
+}
+
+#submission: {
+	language?: string
+	entrypoint?: string
+	author?: #Persons
```
thorehusfeldt: The specification has `authors` (plural), which does make sense with `#Persons`.

thorehusfeldt: Fixed.

RagnarGrootKoerkamp:
Yeah, probably good enough for now. I think we can merge and iterate on this later if we want to.

Thanks a lot for making this one happen! 
So from now on we can use `rejected` and `brute_force`, at last :)
(Although we still need the domjudge magic string for exports right?)

(I recall there was also another name change; adding/dropping an underscore to time_limit_exceeded or so. We already change that right, so this is purely additive now.)

PR #501, opened 2026-01-04
Fix generator count indexing

thorehusfeldt:
(squashed, force pushed, and rewrote the commit message)

For invalid `count`, raise ParseException and do not catch it, so that a generators.yaml with invalid count effectively
aborts fatally (without generating other test cases). This is because their numbering cannot be determined anyway. Add a line number to the exception message.

Clean up parse_count, which is now only ever called once on invalid `count`. Do check for noninteger lists and lists with duplicates (both of which are considered invalid.)

In TestcaseRule, rename argument count_index to count_value, fix earlier regression so that this is indeed be the value that replaces {count} (rather than that value -1).

In test case naming, making padding more consistent when negative numbers are present; avoid padding of count_lists.
In effect, consecutive `count_list`s are now 0-padded to have equal width (including `-1`, `00`, `01`.

Add a test/generatorcount problem to check that {count} is correctly substituted.

### DIFF
--- a/bapctools/generate.py
+++ b/bapctools/generate.py
@@ -490,7 +490,7 @@ class TestcaseRule(Rule):
         self.count_value = count_value
 
         # used to decide if this was supposed to be a duplicate or not
-        self.intended_copy = self.count_value != 1
+        self.intended_copy = has_count(yaml)
 
         # used to handle duplicated testcase rules
         self.copy_of = None
@@ -580,11 +580,12 @@ class TestcaseRule(Rule):
                             )
                     self.generator = GeneratorInvocation(problem, command_string)
 
-                    # TODO: Should the seed depend on white space? For now it does, but
+                    # IMPORTANT: The seed depends on white space, but
                     # leading and trailing whitespace is stripped.
                     seed_value = self.config.random_salt
                     if self.count_value != 1:  # distinguish different count values
-                        seed_value += f":{self.count_value}"
+                        # IMPORTANT: We need to use `self.count_value - 1` for backwards compatibility.
+                        seed_value += f":{self.count_value - 1}"
                     seed_value += self.generator.command_string.strip()
                     self.seed = int(hash_string(seed_value), 16) % 2**31
                     self.in_is_generated = True


thorehusfeldt:
Huh. I think `parse_count` should just throw `ParseException` instead of triggering its own bespoke `bar`. 

Also, not to self: remember test test `list[int]`, not just `list()`.

bapctools/generate.py
```diff
@@ -591,9 +570,9 @@ def __init__(
 
                     # then replace {count} and {seed}
                     if "{count}" in command_string:
-                        if "count" in yaml:
+                        if has_count(yaml):
                             command_string = command_string.replace(
-                                "{count}", f"{self.count_index + 1}"
+                                "{count}", f"{self.count_index}"
```
RagnarGrootKoerkamp: maybe rename `count_index` to `count_value` or so? `index` isn't quite accurate anymore

mzuenni:
> I think there's nothing to figure out. It just isn't checked for. (I think.)

from what I see in the code any directory should be created? The only issue might be that your example is: `data: None` instead of `data: []`.

bapctools/generate.py
```diff
@@ -604,8 +583,8 @@ def __init__(
                     # TODO: Should the seed depend on white space? For now it does, but
                     # leading and trailing whitespace is stripped.
                     seed_value = self.config.random_salt
-                    if self.count_index > 0:
-                        seed_value += f":{self.count_index}"
+                    if self.count_value > 1:  # distinguish different count values
```
RagnarGrootKoerkamp: This should probably be !=1?

RagnarGrootKoerkamp: Ah right, yeah this line of thought makes sense. But still it's more consistent to _always_ have a new seed value rather than only for distinct positive values.

thorehusfeldt: Good points. Thank you for being awake.

So in the old world, `count_index` went (for `count: 10`) through 0,1,2,...,9. For the final 9 of those values, it appended `:1`, ..., `:9` to the seed.

In the current PR, `count_value` is at this stage `1,2,...,10`.  We need to

```python
if count_value != 1:
    seed_value += f":{self.count_index-1}"
```
Did I get this right?

thorehusfeldt: Only tangentially related: When retrying in line 217, we have

> `(seed=(seed + retry) % 2**31)`

Does that work for seeds of the from `1234:2`?

mzuenni: > Did I get this right?

Yes, looks good to me.

> Does that work for seeds of the from 1234:2?

the `seed` is always an int generated from the `seed_value` (a string) via `int(hash_string(seed_value), 16) % 2**31`

PR #514, opened 2026-04-13
chore(docker): add texlive-langextra for indonesian language support

lungsin:
Related to #512, `texlive-langextra` package is needed for the indonesian language for `bt pdf` command. This PR adds it to the Dockerfile.

I saw that #511 also did some modification to the Dockerfile, but it seems orthogonal to this PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,7 @@ RUN pacman -Syu --noconfirm \
 	texlive-binextra \
 	texlive-langeuropean \
 	texlive-langgerman \
+	texlive-langextra \
 	texlive-latexextra \
 	texlive-pictures \
 	texlive-science \


RagnarGrootKoerkamp:
Merged but the docker image needs rebuilding and pushing. Unfortunately I'm getting timeouts on the Arch mirrors and no time to investigate right now. (@mpsijm quick ideas?)

lungsin:
No rush! There are workarounds for installing additional dependencies in the docker image, so it's not blocking for us.

But as a quick dirty solution, you can try to up a container using the existing image, install the pacman deps, then use `docker commit <container id> <image tag>` to snapshot the updated container.

Ref:
- https://docs.docker.com/reference/cli/docker/container/commit/

mpsijm:
@RagnarGrootKoerkamp No clue, building the image from scratch works on my machine :shrug: Just pushed the latest version (including #511) to [Docker Hub](https://hub.docker.com/r/ragnargrootkoerkamp/bapctools) :smile: 

And thanks @jonathanirvings and @lungsin for the contributions! :heart:

RagnarGrootKoerkamp:
thanks :)

PR #515, opened 2026-04-16
Hardcode statement during zip: id to root, and en to attachments

lungsin:


lungsin:
My bad, wrong PR target

Issue #172, opened 2021-11-19
CI test that validation.h is c++17 compatible


RagnarGrootKoerkamp:
And/or for compilation of validators to use c++17 instead of c++20 command line flag (could be ugly search&replace; we really need to enforce this)

PR #516, opened 2026-05-01
[validation.h] Fix CountIterator interface

niklasmohrin:
Before this change, the following line in an input validator

```cpp
v.read_integer("x", 0, 42, Unique);
```

would yield a compilation error

```plain
<snip>/include/c++/14.3.0/bits/stl_algobase.h:514:64: error: no type named ‘iterator_category’ in ‘struct std::iterator_traits<Validator::gen_number<long long int, UniqueTag>(const std::string&, long long int, long long int, UniqueTag)::CountIterator>’
  514 |       typedef typename iterator_traits<_II>::iterator_category _Category;
```

The definition of CountIterator has multiple problems:
1. It does not declare the type aliases required for any of the iterator
   concepts.
2. When incrementing, it should return another iterator, not a value.
   This probably worked coincidentally because the number can be
   converted back to an iterator immediately
3. Dereference and comparison were non-const only.

### DIFF
--- a/bapctools/resources/headers/validation.h
+++ b/bapctools/resources/headers/validation.h
@@ -759,12 +759,16 @@ class Validator {
 					} while(!seen_here.insert(v).second);
 
 					struct CountIterator {
+						using value_type = T;
+						using difference_type = std::ptrdiff_t;
+
 						T v;
-						T& operator*() { return v; }
-						T& operator++() { return ++v; }
-						T operator++(int) { return v++; }
-						bool operator!=(CountIterator r) { return v != r.v; }
+						value_type operator*() const { return v; }
+						CountIterator& operator++() { ++v; return *this; }
+						CountIterator operator++(int) { return { v++ }; }
+						bool operator==(CountIterator r) const { return v == r.v; }
 					};
+					static_assert(std::forward_iterator<CountIterator>);
 
 					if(seen_here.size() > (high - low) / 2) {
 						use_remaining = true;
--- a/bapctools/resources/headers/validation.h
+++ b/bapctools/resources/headers/validation.h
@@ -25,6 +25,7 @@
 #include <functional>
 #include <iomanip>
 #include <iostream>
+#include <iterator>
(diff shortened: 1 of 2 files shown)


bapctools/resources/headers/validation.h
```diff
@@ -759,12 +759,16 @@ class Validator {
...
-						T& operator*() { return v; }
-						T& operator++() { return ++v; }
-						T operator++(int) { return v++; }
-						bool operator!=(CountIterator r) { return v != r.v; }
+						value_type operator*() const { return v; }
+						CountIterator& operator++() { ++v; return *this; }
+						CountIterator operator++(int) { return { v++ }; }
+						bool operator==(CountIterator r) const { return v == r.v; }
 					};
+					static_assert(std::forward_iterator<CountIterator>);
```
niklasmohrin: Note that this line is C++ 20; I saw that there are some backports of C++20 features above. What is the current state, should we remove this assert if it is not available?

I think that generally, we should have the assert if the concept is available

mzuenni: validators should be c++17 compliant

niklasmohrin: Should we update this then? https://github.com/RagnarGrootKoerkamp/BAPCtools/blob/f774626c9a6235b4aa38cd25d0dfac333954de64/bapctools/resources/headers/compile_flags.txt#L1

mzuenni: i think that file is not used for anything? (we don't currently enforce c++17 for the header https://github.com/RagnarGrootKoerkamp/BAPCtools/issues/172)

Side note: work on this header has mostly stalled... we might at some point drop it entirely / ship https://github.com/mzuenni/icpc-header instead.

I will have a look at the PR later

niklasmohrin: It is used by clangd in editors when opening validation.h; with the change my editor now suggests that `forward_iterator` is not a thing

PR #517, opened 2026-05-03
Refactor

mzuenni:


bapctools/generate.py
```diff
@@ -2639,7 +2639,8 @@ def generate(problem: Problem) -> bool:
 
 def testcases(problem: Problem) -> set[Path]:
     try:
-        gen_config = GeneratorConfig(problem)
+        with config.suppress_warnings(level=2):
```
mpsijm: Why suppress warnings here?

bapctools/config.py
```diff
@@ -286,11 +292,23 @@ def copy(self) -> "ARGS":
...
+    try:
+        yield
+    finally:
         args.__dict__.clear()
-        args.__dict__.update(self.args.__dict__)
+        args.__dict__.update(old_args.__dict__)
+
+
+@contextmanager
+def suppress_warnings(level: int = 1) -> Generator[None]:
```
mpsijm, line 316: We should make it clearer what the different levels mean. Possibly even better, the `level` could be an enumeration that allows selecting values like `WARN`(`ING`) or `ERROR`

bapctools/problem.py
```diff
@@ -490,8 +505,15 @@ def __init__(self, path: Path, tmpdir: Path, label: Optional[str] = None):
         self.path = path
         self.tmpdir: Path = tmpdir / self.name
         self.tmpdir.mkdir(parents=True, exist_ok=True)
+
+        bar = PrintBar(self.name)
+        if not self.path.is_dir():
+            bar.fatal("problem directory not found")
+        if not Problem._SHORTNAME_REGEX.fullmatch(self.name):
+            bar.warn(f"name does not match {Problem._SHORTNAME_REGEX.pattern}")
```
mpsijm, line 513: Since you changed the warning to say "name" instead of "shortname", we should print the violating `problem.name` in this warning, to make it clearer to the user which name is violating the regex (in particular, it's not the name in `problem.yaml`).

bapctools/fuzz.py
```diff
@@ -280,26 +279,22 @@ def runner(task: GeneratorTask | SubmissionTask) -> None:
...
-        self.finish_task()
-
-        # wait for the queue to run empty (after config.args.time)
-        self.queue.join()
-        # At this point, no new tasks may be started anymore.
-        self.queue.done()
-
-        signal.signal(signal.SIGINT, old_handler)
+        except KeyboardInterrupt:
+            fatal("Running interrupted", force=True)
```
mpsijm, line 297: Isn't this forceful exit already done by `main()` in `cli.py`?

bapctools/config.py
```diff
@@ -284,3 +290,25 @@ def copy(self) -> "ARGS":
 
 
 args = ARGS("config.py")
+
+
+@contextmanager
+def restore_args() -> Generator[None]:
```
mpsijm: In its current form, the name `restore_args` could be slightly confusing. It does not immediately restore the args when _entering_ the `with` block, but rather, it restores the args when _exiting_ the `with` block.

What do you think of the idea that this context manager can be used to _configure_ the temporary arguments, rather than to clean up changes when exiting the context? This matches with the intuition that we're opening some context with temporarily changed arguments, which will then be cleaned up when exiting the `with` block. So, this would look something like:
```py
        with config.temporary_args({"jobs": (os.cpu_count() or 1) // 2}):
            solve_stats.generate_solve_stats(config.args.post_freeze)
```
The only thing is that this makes it hard to type-check these temporary arguments. It could also be a function that manipulates a temporary version of args, but that's a bit more verbose:
```py
        def use_half_cores(args: config.ARGS):
            args.jobs = (os.cpu_count() or 1) // 2

        with config.temporary_args(use_half_cores):
            solve_stats.generate_solve_stats(config.args.post_freeze)
```
Or perhaps we could pass some temporary `ARGS` object:
```py
        tmp_args = config.ARGS(dummy_path?)
        tmp_args.jobs = (os.cpu_count() or 1) // 2
        with config.temporary_args(tmp_args):
            solve_stats.generate_solve_stats(config.args.post_freeze)
```
But `ARGS` also has a (very untyped) `**kwargs` argument that could be (ab)used to avoid type checking the fields of `ARGS` anyway 😂

mpsijm (changes requested):
Nice changes, I especially like the context managers to automatically restore the arguments 😄

bapctools/generate.py
mzuenni: only the generate function should print these errors. Here it is sufficient to just return no testcases if there are errors.

bapctools/problem.py
mzuenni: The message now looks like this:
```
<the invalid shortname>: name does not match [a-z0-9]{1,255}
```
That should be fine right? I could also say "directory"?

bapctools/fuzz.py
mzuenni: The one in `cli.py` does not set `force=True`

bapctools/config.py
mzuenni: We could do something like:
```python3
tmp_args = config.args.copy()
tmp_args.jobs = (os.cpu_count() or 1) // 2
with config.temporary_args(tmp_args):
```
But (besides the naming) IMO the current syntax looks nicer/cleaner? It makes clear that we are inside a scope and only change args temporarily... Having this `tmp_args` as an actual variable is kind of messy...?

`ARGS` is fully checked at runtime, so that is fine.

bapctools/generate.py
mzuenni: (this function is only used for stats, and could also just return an int...)

bapctools/config.py
mzuenni: should I just rename `restore_args` to `temporary_args`? It does create the temporary `ARGS` object so that would be fitting?

bapctools/config.py
mzuenni: an enum might be a bit overkill (for now)? could also be a bool `suppress_errors`.

Initially I made this an int to represent a "log level" similar to `verbose`... but I came to the conclusion that we should not have a public flag like `suppress_warnings`.

mpsijm: Sure, the ad-hoc integer argument works for now :stuck_out_tongue: Thanks for adding the comment with the explanation at least :slightly_smiling_face:

bapctools/config.py
mpsijm: With the current syntax (which is indeed nice), `temporary_args` is already better yeah, thanks :smile:

mpsijm (approved):
The PR is still in draft mode, what's still needed before it's ready to merge? :slightly_smiling_face:

mzuenni:
> The PR is still in draft mode, what's still needed before it's ready to merge? 🙂

I am currently rewriting the `generators.yaml` parsing based on the `YamlParser` but that could also become a new PR

mpsijm:
> I am currently rewriting the `generators.yaml` parsing based on the `YamlParser` but that could also become a new PR

Nice! :smile: Separate PRs for independent rewrites is easier to review anyway (also prevents generic PR titles like "Refactor") :slightly_smiling_face:

PR #518, opened 2026-05-21
rewrote parser

mzuenni:
There are still some inconsistencies that I don't like:
- Using Exceptions VS `.n_parser_error`
- passing the `generator_config` to `TestcaseRule` but not to `DirectoryRule`
- What work is done in `parse_XXX` vs in `XXXRule.__init__`

But I don't have a good idea on how to resolve/unify this


mzuenni:
@mpsijm do you have time to take a look?

PR #521, opened 2026-06-25
less is_bsd checks

mzuenni:


mzuenni:
@thorehusfeldt @chistopher do you have a mac and can verify that `bt test` works on an interactive problem with these changes?

chistopher:
Yes, it works. The interaction is now visible for me. Thanks for the fix!

PR #522, opened 2026-07-05
Feat/macostest

thorehusfeldt:
Five independent bugs, all exposed by running the full pytest suite on macOS on my own machine. (Pytest has never been all-green on my machine. Now it is. Claude-generated summary below:)

## Path identity under /var symlink (2 fixes)

macOS symlinks /var → /private/var. Path.__eq__ compares strings without resolving symlinks, so paths constructed from tempfile.gettempdir() (which returns /var/folders/…) didn't compare equal to paths returned by os.getcwd() or iterdir() (which return /private/var/folders/…).

- cli.py: apply .resolve() when constructing the tmpdir so all derived paths are canonical from the start.
- run.py (_prepare_nextpass): the feedbackdir comparison f == self.feedbackdir silently failed, causing the feedbackdir to be deleted between passes of multi-pass problems. Fixed with .resolve() on both sides.

## remove_path on non-empty directories (util.py)

On Linux, path.unlink() on a directory raises IsADirectoryError. On macOS it raises OSError(ENOTEMPTY) instead. Collapsed the exception handling to catch OSError broadly and check is_dir() before calling shutil.rmtree.

## javac inheriting user's CLASSPATH (languages.yaml)

javac without -cp inherits the environment's CLASSPATH. On a developer machine this can include unrelated jars (e.g. from a university course), causing Operation not permitted during compilation. Added -cp {path} to override it with just the tmpdir.

## questionary bypassing sys.stdin monkeypatch (test_problems.py)

questionary reads from the raw TTY file descriptor, not sys.stdin, so monkeypatching sys.stdin had no effect. Fixed by also patching bapctools.util.has_questionary = False to force the plain input() fallback path.

## Platform-specific type: ignore comments (pyproject.toml)

sched_getaffinity/sched_setaffinity exist in Linux's typeshed but not macOS's, so # type: ignore[attr-defined] comments needed on macOS are flagged as ed per-file warn_unused_ignores = falseoverrides for the four affected modules.

## from_chars for long double on macOS (validate.h)

Apple's libc++ implements std::from_chars for float and double but not long double. The upstream icpc-header already had a DOUBLE_FALLBACK (stold) foh #ifdef __APPLE__.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,6 +25,33 @@ jobs:
       - run: bash test/yaml/validate_yaml.sh
       - run: pytest
 
+  test-macos:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-python@v6
+        with:
+          python-version: '3.10'
+      - uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: '21'
+      - run: brew install kotlin
+      - run: brew install --cask basictex
+      - run: |
+          export PATH="/Library/TeX/texbin:$PATH"
+          sudo tlmgr update --self
+          sudo tlmgr install \
+            latexmk lm \
+            collection-latexrecommended collection-latexextra \
+            collection-science \
+            babel-german hyphen-german \
+            asymptote
+      - run: pip install --editable . --group dev
+      - run: |
+          export PATH="/Library/TeX/texbin:$PATH"
+          pytest
+
   test-wsl:
     runs-on: windows-latest
     steps:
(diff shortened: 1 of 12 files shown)


RagnarGrootKoerkamp:
This summary looks very scary but the changes look fine I'd say. But will let someone else have a look as well.

Only the changes to `interaction.py` feel a bit iffy. If we `killpg` a 'process group', we want that to succeed and why should it not just work?

Maaaybe `tmpdir = tmpdir.resolve()` is also bad because there are some paths where we check their relative location before resolving. But probably `tmpdir` is not one of those and its fine.

PR #532, opened 2026-08-16
update langauges.yaml

mzuenni:
should make our code match problemtools again (https://github.com/Kattis/problemtools/pull/458)

### DIFF
--- a/bapctools/resources/support/schemas/languages.cue
+++ b/bapctools/resources/support/schemas/languages.cue
@@ -23,12 +23,19 @@ template_string: s={
 #Languages: {
 	// Language names must be lowercase and start with a letter, and can only contain letters and digits
 	[=~"^[a-z][a-z0-9]*$"]: {
-		name!:     string
-		priority!: int
-		files!:    string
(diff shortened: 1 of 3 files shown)


bapctools/resources/support/schemas/languages.cue
```diff
@@ -23,12 +23,19 @@ template_string: s={
...
-		name!:     string
-		priority!: int
-		files!:    string
-		shebang?:  string
-		compile?:  template_string
-		run!:      template_string
+		name!:          string
+		priority!:      int
+		files!:         string
+		shebang?:       string
```
thorehusfeldt, line 29: This should be

```cue
 { 
  shebang!: string
  shebang_files!: string
} | *{}
```
This construction includes both keys mandatorily, or _absence_ of both keys. (The `*` is important, otherwise CUE doesn’t know how resolve absence.) Because `#Languages` has `#`, the struct is a definition (=“closed” in CUE-speak) that does not allow other keys. LInes 33–38 are then redundant.

PR #533, opened 2026-08-25
preallocate buffer

mzuenni:


mzuenni:
this seems to perform worse than main, will close for now.

